PR #64, opened 2021-10-07
Kick unused variables, moove some forms styles and use badge instead of custom flags style

NeOMakinG:


matks:
@NeOMakinG Are you sure we have to kick unused variables? Cannot they be useful for a child theme?

PR #99, opened 2021-11-08
Add jQuery and prestashop object without core.js

NeOMakinG:


SharakPL:
What's the point of Bootstrap 5 if you're still going to use jQuery?

PR #123, opened 2021-11-23
Add .editorconfig

rodriciru:
I copy from https://github.com/PrestaShop/PrestaShop/blob/develop/.editorconfig and add scss rules


NeOMakinG (approved):

NeOMakinG:
Thanks @rodriciru 🚀

PR #124, opened 2021-11-24
Refactor of layout, my account, addresses, contact

Hlavtox:
Ping @SharakPL @rodriciru if you want to take a look. :-)


templates/components/account-menu.tpl
```diff
@@ -98,3 +98,9 @@
     </a>
   </div>
 {/block}
+
+{block name='account_link'}
+  <a class="btn btn-outline-primary hidden-on-desktop" href="{$urls.pages.my_account}">
```
rodriciru, line 103: replace hidden-on-desktop by d-md-none
this .btn has to have an .mb-3 Its too close to the title

Hlavtox: @rodriciru We didn't do any styling at all, just the basic code and blocks now. :-)

.hidden-on-desktop is our own class, basically extending d-none and others. We can discuss if we want to use it or use the default boostrap ones.

Hlavtox: @NeOMakinG Maybe using reverting to `d-none d-md-block` /  `d-md-none` would bring more flexibility...

rodriciru:
`.label .label-pill` relace by `.badge .rounded-pill`

templates/components/account-menu.tpl
rodriciru: i think that use classes wich do the same as bootstrap ones are redundant and unnecesary thats why i want to replace hidden-on-desktop class

Ok, i think .mb-3 class is layout as its write in .tpl classes

Hlavtox:
@rodriciru ✔️

SharakPL:
Am I really the only one that finds mixing BEM with kebab-case messed up? 😕

templates/components/account-menu.tpl
SharakPL: Exactly! What's the point of all boostrap utility classes if you create your own for the same thing? That's why [I suggested](https://github.com/PrestaShop/theme-refacto/pull/121#issuecomment-975453760) refactoring theme structure. Almost whole theme can be made with just bootstrap. No need to make duplications when all it takes is using proper bootstrap class.

NeOMakinG: Well, `hidden-on-desktop` and `hidden-on-mobile` extends Bootstrap utility classes, it's respecting CSS good practices as bootstrap is not an utility framework like Tailwind is, so we're trying to keep the paradigm between CSS and HTML. Using tailwind for example doesn't mean that you can't use custom class composed with utilities (https://tailwindcss.com/docs/extracting-components#extracting-component-classes-with-apply)

It's more readable, maintainable, avoid any responsive inconsistency because every breakpoints will be the same everywhere

@Hlavtox How would it bring more flexibility?

NeOMakinG: > 

We're not creating classes to do the same thing, we're composing with bootstrap utilities in order to avoid redundancy, inconsistency...

NeOMakinG: > 



> Am I really the only one that finds mixing BEM with kebab-case messed up? 😕

No you're not the only one, I think some components have been done before the BEM decisions, their refacto should be a WIP

NeOMakinG:
> Am I really the only one that finds mixing BEM with kebab-case messed up? 😕

No you're not the only one, I think some components have been done before the BEM decisions, their refacto should be a WIP

rodriciru:
Any problems whit my suggestion???

src/scss/custom/components/customer/_customer-menu.scss
```diff
@@ -24,7 +24,7 @@ $componentName: account-menu;
     }
 
     &.#{$componentName}--signout {
-      .#{$componentName}__link {
+      .link-item {
```
NeOMakinG, line 27: I think this is pretty bad honestly, we're going against our practices in order to unify with old modules that should apply to our naming, don't you think?

Overwise we would have kept every old classes :/

NeOMakinG:
> Any problems whit my suggestion???

Except the composed utility class advice from you and @SharakPL that I think I don't agree at all (I would like your advice @PierreRambaud), nothing ;)

SharakPL:
You really would want to go with new utility classes for the same functionality Bootstrap already provides? That's just duplicating css.

src/scss/custom/components/customer/_customer-menu.scss
Hlavtox: @NeOMakinG I renamed to the old one because `.#{$componentName}__link` is confusing, it's not a link.

Can we name it `.#{$componentName}__link-content`?

PierreRambaud:
> You really would want to go with new utility classes for the same functionality Bootstrap already provides? That's just duplicating css.

The css is not duplicated, it's composed.
When you use `.hidden-on-mobile` you don't do the job twice, you're composing your css class with other classes, that the aim of the extend utility classes. 
That's exactly what bootstrap is doing with [containers](https://github.com/twbs/bootstrap/blob/57d80fcd3274ba1ff0a1580966e098fda80f74e7/scss/_containers.scss#L16), instead of doing `container-fluid container-md` you're able to do only `container-md`.

That's the same thing when you write a method that is calling other methods, sometimes wrapping calls is really better for a better understanding.

I agree with the others, it's more meaningful to use `hidden-on-something`.

templates/customer/my-account.tpl
```diff
@@ -26,93 +26,80 @@
...
-          <a class="{$componentName}__link col-12 col-md-4" id="history-link" href="{$urls.pages.history}">
-            <span class="link-item">
-              <i class="material-icons">&#xE916;</i>
-              {l s='Order history and details' d='Shop.Theme.Customeraccount'}
-            </span>
-          </a>
-        {/if}
+{block name='page_content'}
+  <div class="{$componentName} row g-3">
+    <a class="{$componentName}__link col-md-6 col-lg-4" id="identity-link" href="{$urls.pages.identity}">
```
Hlavtox, line 35: Let's rethink this naming to be short and clean.
Also, I think we can move `<span class="link-item">` and wrap only the text with it. I don't like mixing text with other elements, wdyt?

SharakPL:
@PierreRambaud I get it, but it would require a class for [each screen size](https://getbootstrap.com/docs/5.1/utilities/display/#hiding-elements). I've been dealing with composed classes for mobile and desktop only an it was a nightmare so it would have to be:
```scss
. hidden-on-mobile {
  @extends .d-none, d-md-block;
}
. hidden-on-tablet {
  @extends .d-block, .d-md-none, d-lg-block;
}
...
```
And what if I don't want it to be block, but inline or flex? You'll make whole set of classes for this: `hidden-on-mobile-block`, `hidden-on-mobile-flex`, `hidden-on-tablet-otherwise-inline`, `hidden-on-laptop-below-block-above-flex` etc.? Otherwise you would have duplicate things and end up using `d-block hidden-on-tablet d-flex` anyway. Is there any better way to handle this?

Besides `hidden` doesn't mean `display: none`. It can also be
- `visibility: hidden;`
- `opacity: 0;`
- `height: 0; overflow: hidden`

and I believe we are talking about simple `display: none` here so bootstrap's `d(-xy)-none` classes are IMHO best choice here

Hlavtox:
@NeOMakinG @PierreRambaud I have to agree with @SharakPL.

`d-none d-md-block`
- Is the same length as `hidden-on-mobile`
- Smaller CSS size by some bytes
- You know which breakpoint it is and don't have to dig. What is a mobile, is it md or lg?
- You can control the switching easily and don't have to have many utility classes.
- You can control block/flex display

PierreRambaud:
@SharakPL @Hlavtox as you wish, but keep in mind, if you think about theme developer, you should remove all markup linked to the CSS and use semantic class naming instead, otherwise, each time you want to change something, you have to redo the template and the CSS, instead of only the CSS :wink:

Hlavtox:
Let everybody think for a while, nothing is final now. Let just get the basic things together! We can think about the right balance between utility classes and CSS in meantime.
- Finish coding.
- Make everything working without jQuery.
- Then have a second pass and optimize, remove, add hooks.
- Have a third pass and remove as much things possible from core to custom, so developers have a nice base.

src/scss/custom/modules/_contactinfo.scss
```diff
@@ -0,0 +1,23 @@
+.contact-rich {
+
+  .block {
+    display: flex;
+
+    &.email {
```
Hlavtox, line 6: Todo fix

rodriciru:
i made a PR regarding cart, but i think maybe it should be here....
https://github.com/PrestaShop/theme-refacto/pull/130

src/scss/custom/components/customer/_customer-menu.scss
NeOMakinG: I think so, no problem with that

templates/checkout/cart.tpl
```diff
@@ -25,69 +25,60 @@
...
+        {block name='cart_overview'}
+          {include file='checkout/_partials/cart-detailed.tpl' cart=$cart}
+        {/block}
+      </div>
 
-        <!-- Right Block: cart subtotal & cart total -->
-        <div class="cart-grid-right col-lg-4">
-          <h2 class="h4">{l s='Order summary' d='Shop.Theme.Checkout'}</h2>
+      {block name='continue_shopping'}
+        <a class="label" href="{$urls.pages.index}">
```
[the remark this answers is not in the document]
NeOMakinG: @Hlavtox what about this one?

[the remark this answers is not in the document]
NeOMakinG: Only focusing on layout, so it's not the goal of the PR? That's why you resolved?

NeOMakinG:
@rodriciru what do you think about merging this one? As it's aiming mainly layout things, we can work on your comments inside another one? And so we can get further!

NeOMakinG (approved):

rodriciru:
Yours are the presta-pros!
Im new to git or contribute to anithing so whatevere you choose i will stick to it

NeOMakinG:
> Yours are the presta-pros! Im new to git or contribute to anithing so whatevere you choose i will stick to it

I did implement your comment reviews already

rodriciru:
Yes i see.
Awesome work @Hlavtox

PR #128, opened 2021-12-01
Slider home ratio

rodriciru:
| Questions         | Answers
| Description?      | 
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/theme-refacto/issues/127
| How to test?      | Check home slider in mobile and desktop and product page slider
| Possible impacts? | None

I try the BS5 ratio aproach for home slide only


rodriciru:
i Scss lint fix _custom.scss which i dont see ani reason to disable stylelint inside, but your tell me

src/scss/custom/modules/_imageslider.scss
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.ratio-homeSlider {
+  min-height: 300px;
+
+  --bs-aspect-ratio: calc(9 / 21 * 100%); // 21x9
+  @include media-breakpoint-up(md) {
+    --bs-aspect-ratio: calc(7 / 30 * 100%); // 30x7
+  }
+}
```
NeOMakinG, line 19: Be careful, we already have some _slider partials inside `custom` and `core` scss folders, you should work inside these files or remove them and create _imageslider if you want to stick to module name

NeOMakinG:
> i Scss lint fix _custom.scss which i dont see ani reason to disable stylelint inside, but your tell me

It's fine, was a mistake

src/scss/custom/modules/_imageslider.scss
rodriciru: the file is already called _imageslider  ..... :)

rodriciru: should be inside core or custom folder?

NeOMakinG: image
I think it should be in custom, theme developer will be able to handle his slider styling if he removes the custom part

We already have a _slider partial inside custom, let's move his content into your new _imageslider file ? The _slider is an error I think so

rodriciru: Not agree. My _imageslider.scss are ONLY styles for the home slider. Not any other slider in the store ( product, quickview or whatever) 
Thats the reason to put in custom/modules

NeOMakinG: Ok, good arg, but in the future, if we find that some styles should only go to the homeSlider, then we'll need to move these!

NeOMakinG (approved):

NeOMakinG:
Thanks @rodriciru 

We may change these depending of the result of the tests of Swiper, or keep them if we stick with bootstrap

Anyway, these styles will need to be moved into every modules folders, so they are not loaded if the module is disabled, we need to find a way to to this!

PR #144, opened 2021-12-21
Add support for line breaks

Hlavtox:


SharakPL:
TBH I never liked contact details preformatted. It would be better to just generate whole contact information with single values. support for multiple e-mail addresses and phone numbers would also be great as client quite often want to put couple of them.

modules/ps_contactinfo/ps_contactinfo-rich.tpl
```diff
@@ -53,7 +53,7 @@
     <hr/>
     <div class="contact__item">
       <i class="material-icons">&#xE88E;</i>
-      <div class="contact__info">{$contact_infos.details}</div>
+      <div class="contact__info">{$contact_infos.details|nl2br nofilter}</div>
```
PierreRambaud, line 56: Instead of having this comment, you can do 
```
{if !empty($contact_infos.details)}
```

NeOMakinG: @Hlavtox wdyt?

NeOMakinG (approved):

PR #151, opened 2022-01-19
Replace JQuery UI sliders by noUiSlider

NeOMakinG:
| Questions         | Answers
| Description?      | We wants to avoid using jQuery UI


SharakPL:
No Swiper then? 😉

PR #161, opened 2022-01-22
Redesign color group UI in faceted search

mparvazi:
| Questions         | Answers
| Description?      | Change the color group UI to inline-block display and hide input checkbox.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | -
| Possible impacts? | Color selection group in faceted search.

| Current  | PR |
| ------------- | ------------- |
|   |  |

### DIFF
--- a/modules/ps_facetedsearch/views/templates/front/catalog/facets.tpl
+++ b/modules/ps_facetedsearch/views/templates/front/catalog/facets.tpl
@@ -58,12 +58,12 @@
                       {continue}
                     {/if}
 
-                    <li>
+                    <li {if isset($filter.properties.color) || isset($filter.properties.texture)}class="d-inline-block"{/if}>
                       <div class="{$componentName}-label facet-label{if $filter.active} active {/if}">
                         {if $facet.multipleSelectionAllowed}
-                          <div class="form-check">
+                          <div class="form-check{if isset($filter.properties.color) || isset($filter.properties.texture)} px-0{/if}">
                             <input 
-                              class="form-check-input" 
+                              class="form-check-input{if isset($filter.properties.color) || isset($filter.properties.texture)} visually-hidden{/if}" 
                               id="facet_input_{$_expand_id}_{$filter_key}"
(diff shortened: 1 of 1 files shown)


SharakPL:
I'm not sure this is an improvement. Colors need description, that should be visible on filters. Textures can be used for color variants  and the color name can be anything so small square alone is not enough.
Although I agree checkboxes look bad here. Green border with ✔️ in the corner for selected variant would be nice...

mparvazi:
We have 3 styles for color group in faceted search:

1. Checkbox
2. Radio button
3. Dropdon menu

This PR just focused on Checkbox style and make it similar to the style in product page.
Screenshot 2022-01-22 at 23-35-48 Hummingbird printed t-shirt

SharakPL:
On product page you at least get the label of currently selected color. IMO it's still not enough. If you had only small textures in variants you wouldn't know without the label what you actually select.
Either way I think labels are required here. I would rather go different way and allow adding icons/images to all attributes, no matter its style. Plus maybe configure it per attribute if it should load just icon/image, just label or both.

mparvazi:
We can use a check icon for selected colors in faceted search group.
And a tooltip for color name on hover.

Btw the label of colors will be displayed in active filters.

Screenshot 2022-01-23 at 00-25-54 Men

mparvazi:
Or using labels for colors:

Screenshot 2022-01-23 at 00-38-35 Men

SharakPL:
2nd one is better IMO. I'd rather use labels only than colored boxes only.
Checkmark on the box would force you to use different colors for the ✔. Otherwise it could blend with the box or at laest be very hard to notice. 

Placing active filters in completely different place than filters itself is big mistake IMO. I never liked it in Classic theme. All states should remain inside Filters block for best UX so either whole block should be moved to main column just above product grid or active filters should be moved back to Filters block in left column, just like it was in PS 1.6.

I think left column would be better to handle active filters. To avoid adding additional list to the block it would be best to simple accent activated filters and maybe move them to the top of the list. Maybe something like this:

obraz

PR #181, opened 2022-02-11
Redesign quantity input UX

### LINKED ISSUE #173: Improve input-number spin buttons for mobile
### Mobile Usability report - Google Support
**Clickable elements too close together**
Touch elements, such as buttons and navigational links, are so close to each other that a mobile user cannot easily tap a desired element with their finger without also tapping a neighboring element. 

So I think we can have two different UI for input-number:

| Desktop  | Mobile |
| ------------- | ------------- |
|   |  |

mparvazi:
| Questions         | Answers
| Description?      | Adds spin buttons to input number. Controls input value. Update cart items on input change.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #173
| How to test?      | -
| Possible impacts? | Shopping cart & Product page

https://user-images.githubusercontent.com/85633460/153760217-b514db68-782f-4bf3-8613-cb12bf131df0.mp4

### DIFF
--- a/src/js/qty-input.ts
+++ b/src/js/qty-input.ts
@@ -30,7 +30,7 @@ export default function initQuantityInput(selector = SelectorsMap.qtyInput.defau
   const incrementIcon = '&#xE145;';
   if (qtyInputList) {
     qtyInputList.forEach(function(qtyInput) {
-      const qtyInputWrapper = qtyInput.parentNode;
+      const qtyInputWrapper = qtyInput.parentElement;
       if (qtyInputWrapper && qtyInputWrapper.childElementCount === 1) {
         let decrementButton = createSpinButton(decrementIcon);
         decrementButton.addEventListener('click', () => changeQuantity(<HTMLInputElement>qtyInput, -1));
--- a/src/js/qty-input.ts
+++ b/src/js/qty-input.ts
@@ -41,11 +41,14 @@ export default function initQuantityInput(selector = SelectorsMap.qtyInput.defau
   }
 }
 
-function createSpinButton(text: string) {
+function createSpinButton(iconText: string) {
   let spinButton = document.createElement('button');
   spinButton.type = 'button';
   spinButton.classList.add('btn');
-  spinButton.innerHTML = '<i class="material-icons">' + text + '</i>';
+  let spinIcon = document.createElement('i');
+  spinIcon.classList.add('material-icons');
+  spinIcon.innerText = iconText;
+  spinButton.appendChild(spinIcon);
   return spinButton;
 }
 
--- a/src/js/qty-input.ts
+++ b/src/js/qty-input.ts
@@ -26,45 +26,37 @@ import { Toast } from 'bootstrap';
 import SelectorsMap, { listing } from './selectors-map';
 
 export default function initQuantityInput(selector = SelectorsMap.qtyInput.default) {
-  const qtyInputList = document.querySelectorAll(selector);
+  const qtyInputNodeList = document.querySelectorAll(selector);
   const decrementIcon = 'E15B';
   const incrementIcon = 'E145';
 
-  if (qtyInputList) {
-    qtyInputList.forEach(function(qtyInput) {
+  if (qtyInputNodeList) {
+    qtyInputNodeList.forEach(function(qtyInput) {
       const qtyInputWrapper = qtyInput.parentElement;
 
       if (qtyInputWrapper && qtyInputWrapper.childElementCount === 1) {
-        let decrementButton = createSpinButton(decrementIcon);
+        const decrementButton = createSpinButton(decrementIcon);
         decrementButton.addEventListener('click', () => changeQuantity(<HTMLInputElement>qtyInput, -1));
 
-        let incrementButton = createSpinButton(incrementIcon);
+        const incrementButton = createSpinButton(incrementIcon);
         incrementButton.addEventListener('click', () => changeQuantity(<HTMLInputElement>qtyInput, 1));
 
-        if (!languageIsRTL()) {
-          qtyInputWrapper.insertBefore(decrementButton, qtyInput);
-          qtyInputWrapper.appendChild(incrementButton);
-        } else {
-          qtyInputWrapper.insertBefore(incrementButton, qtyInput);
-          qtyInputWrapper.appendChild(decrementButton);
-        }
+        qtyInput.classList.add('ready');
+        qtyInputWrapper.insertBefore(decrementButton, qtyInput);
+        qtyInputWrapper.appendChild(incrementButton);
       }
     });
   }
 }
 
-function languageIsRTL() {
-  return document.querySelector('body')?.classList.contains('lang-rtl');
-}
-
 function createSpinButton(codePoint: string) {
-  let spinButton = document.createElement('button');
+  const spinButton = document.createElement('button');
   spinButton.type = 'button';
   spinButton.classList.add('btn');
 
-  let spinIcon = document.createElement('i');
-  spinIcon.classList.add('material-icons');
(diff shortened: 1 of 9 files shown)


src/scss/core/components/_qty-input.scss
```diff
@@ -1,33 +1,61 @@
...
 
   button {
     display: flex;
     align-items: center;
     justify-content: center;
-    height: 50%;
+    min-width: 2rem;
     padding: 0;
     border: 1px solid $input-border-color;
+    @include border-radius(0);
```
SharakPL, line 35: All this code (and even more) is completely unnecesary as bootstrap already handles this by default, even with input height changed:
obraz

SharakPL (commented):
It's still the same problem as with Classic. A lot of unnecessary styling which can is already handled by Bootstrap. Changes should be made mostly in _variables.scss

templates/catalog/_partials/product-add-to-cart.tpl
```diff
@@ -28,13 +28,13 @@
...
       <div class="product-actions__inputs row m-0">
-        <div class="product-actions__quantity quantity-button js-quantity-button col-2 p-0 me-2">
+        <div class="product-actions__quantity quantity-button js-quantity-button col-lg-auto p-0 mb-2 me-2">
           <input
             type="number"
             name="qty"
             id="quantity_wanted"
             inputmode="numeric"
-            pattern="[0-9]*"
+            pattern="[1-9]\d*"
```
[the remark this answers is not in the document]
mparvazi: "\d" => is for 10, 20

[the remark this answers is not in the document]
SharakPL: https://stackoverflow.com/questions/16621738/d-less-efficient-than-0-9

src/scss/core/components/_qty-input.scss
mparvazi: It's related to material-icons inside spin buttons.

templates/catalog/_partials/product-add-to-cart.tpl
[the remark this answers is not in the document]
SharakPL: Actually \d just duplicates most of [1-9]. I'd stick with `[0-9]*`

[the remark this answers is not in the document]
mparvazi: We have also "min="{$product.minimal_quantity}""
I must add controls for "min"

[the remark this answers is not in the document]
SharakPL: I noticed. Isn't `$product.minimal_quantity=1` by default?
Anyway maybe this could be more efficient: `[1-9][0-9]*`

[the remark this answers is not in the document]
mparvazi: Screenshot from 2022-02-11 19-19-39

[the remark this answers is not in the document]
mparvazi: Screenshot from 2022-02-11 19-21-16

mparvazi:
Too many commits :smile: 

Tested for:
1. Low speed connection
2. Chrome, Safari and Firefox
3. XHR handling (performance, error)
4. RTL support

NeOMakinG:
> Too many commits 😄
> 
> Tested for:
> 
> 1. Low speed connection
> 2. Chrome, Safari and Firefox
> 3. XHR handling (performance, error)
> 4. RTL support

Don't hesitate to squash in order to clear the commit history, only keep useful things packed together inside commits if you think it's relevant

`git rebase -i HEAD~31`

src/js/qty-input.ts
```diff
@@ -22,24 +22,191 @@
...
+  const decrementIcon = 'E15B';
+  const incrementIcon = 'E145';
+
+  if (qtyInputNodeList) {
+    qtyInputNodeList.forEach(function(qtyInput) {
+      const qtyInputWrapper = qtyInput.parentElement;
+
+      if (qtyInputWrapper && qtyInputWrapper.childElementCount === 1) {
+        const decrementButton = createSpinButton(decrementIcon);
+        decrementButton.addEventListener('click', () => changeQuantity(<HTMLInputElement>qtyInput, -1));
```
NeOMakinG, line 39: I think we can find a way to avoid these cast, I'll fetch your branch and take a look later

src/js/qty-input.ts
```diff
@@ -22,24 +22,191 @@
...
+        targetButton.appendChild(spinner);
+      }
+    }
+  }
+
+  xhttp.onload = function () {
+    const resp = JSON.parse(xhttp.responseText);
+
+    if (resp['hasError']) {
+      showUpdateOpertationErrors(resp);
```
NeOMakinG, line 107:
```suggestion
      showUpdateOperationErrors(resp);
```

src/scss/core/components/_qty-input.scss
```diff
@@ -1,39 +1,71 @@
...
+      }
+    }
   }
 
   button {
     display: flex;
     align-items: center;
     justify-content: center;
-    height: 50%;
+    min-width: 2rem;
```
NeOMakinG, line 33:
```suggestion
    min-width: 2rem;
```
Do we really need this? what about a padding instead? So we don't enforce any sizes, just asking, I need to deep in the subject before my review get more values

NeOMakinG:
I'll fetch your branch later in the day and check everything, this is not the definitive review 🙏

src/scss/core/components/_qty-input.scss
mparvazi: I tried to prevent flashing in size when document is not loaded, size of input is different before and after script loaded.
So spin-button size must be fixed for calculation.

Script not loaded (low speed connection or refresh):

Screenshot 2022-02-14 at 13-57-36 Cart

Script loaded:

Screenshot 2022-02-14 at 13-57-04 Cart

SharakPL: Do we even need bootstrap-touchspin in theme-refacto? Why not use BS input-group component and add functions to - and + buttons?

NeOMakinG: We don't need it, and we don't want it I think so

NeOMakinG: I think this is the goal of this PR, inputQuantity doesn't use it anymore, it was the only place where it was used, we can completely remove it

NeOMakinG: There are a lot of wrong things in the pattern of this PR, we don't need to build these buttons using JS, they can be directly built with input-group components as @SharakPL said, it will avoid this flashing content

NeOMakinG:
As I said in pm:

- Toasts are not linked to the `input-quantity` component at all, let's make it standalone, and input-quantity will sue it as a dependency. I suggest using a `template` that we can clone inside this lib, as Bootstrap doesn't permit generating toast on the fly.
- We can remove the markup generation from input-quantity, because we can directly put the input-group markup to the qty input inside smarty, the only thing we need to do is add some JS logic in order to make the `+-` buttons
- Try to separate utils function from business logic => `getRequestParameters` can be inside a `utils` folder, and this function can be unit tested in the future (maybe there are other parts that can be unit tested, it depends on the final result of everything we seen together

SharakPL:
👍👍 👍

mparvazi:
I'll work on useAlert first, then we can decide which one is better for Cart page update qty: useToast or useAlert.

There will be two options for useAlert to be shown: 
1. Inside "notfications" wrapper at the top of content
2. Inside a custom container (for example warning for qty input numbers will be displayed in corresponding product row)

But need to check it's possible or not.

PR #184, opened 2022-02-17
Add useToast

mparvazi:
| Questions         | Answers
| Description?      | Add capability for using BS5 toast in theme-refacto.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | -
| Possible impacts? | -


mparvazi:
```
useToast('Hello world!').show();

useToast('You\'re in danger!', { type: 'danger', autohide: false }).show();

useToast('You\'re good to go!', { type: 'success', autohide: true, delay: 2000 }).show();

// Event handling
const toast = useToast('Who am I?' , { type: 'warning', classlist: 'its-me' });
$('.toast.its-me').on('shown.bs.toast', () => {});

```

templates/components/toast-container
```diff
@@ -0,0 +1,31 @@
...
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ *}
+{$componentName = 'toast-container'}
+
+<div class="js-{$componentName}-wrapper">
+    <div class="{$componentName} position-fixed top-0 end-0 p-3" id="js-{$componentName}">
+        {include file='components/toast.tpl'}
+    </div>
+</div>
```
NeOMakinG, line 31: You forgot the file extension

templates/layouts/layout-both-columns.tpl
```diff
@@ -112,5 +112,7 @@
     {block name='hook_before_body_closing_tag'}
       {hook h='displayBeforeBodyClosingTag'}
     {/block}
+    
+    {include file='components/toast-container.tpl'}
```
NeOMakinG, line 116: Prefer adding it to bottom_elements to keep consistency, `page-loader` is an utility component like this one

src/js/helpers/useToast.ts
```diff
@@ -0,0 +1,131 @@
...
+    const toastClone = toastTemplate?.content.cloneNode(true) as DocumentFragment;
+    const toastElement = toastClone?.querySelector<HTMLElement>(selectorsMap.toast.toast);
+    const toastBody = toastElement?.querySelector<HTMLElement>(selectorsMap.toast.body);
+
+    if (toastElement && toastBody) {
+      toastContainer.appendChild(toastElement);
+  
+      return toastElement;
+    } else {
+      toastContainer.remove();
```
NeOMakinG, line 60:
```suggestion
      toastContainer.remove();
```

Don't you need to verify if it exist before removing it?

NeOMakinG:
The implementation looks so fine now!

src/js/helpers/useToast.ts
mparvazi: Line 49: will be verified.

NeOMakinG: My bad, but then there is another problem, what happen if the container is undefined? It's not generated right?

There are some edge cases where child theme can still remove the container, it should be also added on the fly if it doesn't exist!

It will be done only one time, as the first toast use will inject it, then next toasts won't need to inject it

src/js/constants/mocks/useToast-data.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export {
+  containerComponent,
+  btnCloseComponent,
+  testMessage,
+  testOptions,
+  defaultType,
+  validTemplate,
+  invalidTemplate,
+};
```
NeOMakinG, line 61: This file should be moved, it's not mocks, it's data used by the application, mocks only contains mocked things such as some object you need in order to test, virtual dom things...

src/js/helpers/useToast.ts
mparvazi: When fallback will be used:

1. toast-container is not exists
2. toast-container is exists but toast template has invalid structure

src/js/constants/mocks/useToast-data.ts
mparvazi: It's only will be used by useToast.test.ts

src/js/helpers/useToast.ts
NeOMakinG: Oh because you remove the container

I don't really like it, imagine that the container has some new classes added by the child theme for any reason?

I would verify if the container exist and I would generate it only if it doesn't exist

src/js/helpers/useToast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+    getBtnClose() {
+      if (testOptions.autohide !== undefined && testOptions.autohide === false) {
+
+        return `.${btnCloseComponent}`;
+      }
+    }
+  }
+
+  describe('with valid template', () => {
```
NeOMakinG, line 36:
```suggestion
  describe('with container and toast template existing in the DOM ', () => {
```

I would make it a bit more verbose

src/js/helpers/useToast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      if (expectedBtnClose) {
+        const btnCloseElement = toastElement?.querySelector<HTMLButtonElement>(expectedBtnClose);
+
+        expect(btnCloseElement?.classList.contains('d-none')).toBe(false);
+      }
+    });
+  });
+
+  describe('with invalid template', () => {
```
NeOMakinG, line 77:
```suggestion
  describe('with container and toast template not existing in the DOM', () => {
```

src/js/helpers/useToast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const toastBody = toastElement?.querySelector<HTMLElement>(`${selectorsMap.toast.toast}-body`);
+    
+    return toastBody?.innerHTML ?? '';
+  }
+
+  const isFallbackConnected = (document: Document) => {
+    const toastContainer = document.querySelector<HTMLElement>(selectorsMap.toast.container);
+
+    return toastContainer?.classList.contains(`${containerComponent}--fallback`);
+  }
```
NeOMakinG, line 121: Can it be on the top of the describe, or the file?

Do we really need document to be sent as a parameter?

src/js/helpers/useToast.test.ts
mparvazi: `without container and toast template existing in the DOM`

Wdyt?

NeOMakinG:
We are really near the end, tests are looking good! Except some little minor questions, very good job!

src/js/helpers/useToast.ts
mparvazi: Why someone need to add custom class to toast container, the position is fixed, it doesn't have any text content, or style rules,  just will be used by BS5 Toast for finding toasts.

mparvazi:
Wdyt about using BS5 Alert?

```
useAlert(' A simple light alert—check it out!');
useAlert('A simple danger alert—check it out!', { type: 'danger' });
```
`<div id="notifications"></div>` will be used as container.

src/js/helpers/useToast.ts
NeOMakinG: If they want the toast system to be on the bottom right for example, instead of top right :)

NeOMakinG: You'll end by just erasing it, because it misses the toast template in some way

I agree it's an edge case, but it can be handled so..

mparvazi: Oh man!
You right.
Let me think about that.

mparvazi: If `toast-container` exists, but `toast-template` is not exist or is not valid.,
then remove inner html of `toast-container` and replace with inner html of fallback-container.

Wdyt?

NeOMakinG: You'll have some problem if someone or a module did use `useToast` before you, because it will erase their toast instance, meaning that when they'll throw `myToast.show()`, it won't be shown anymore because the inject markup is not in the `.toast-container` anymore

NeOMakinG: I think, if `toast-container` exists, just happen the toast markup to it, if it doesn't exist, append a `toast-container` to the body and then append the `toast-template` content to it!

NeOMakinG: You should probably separate it to 2 different logic:
- The toast-container check (and append to the body if it doesn't exist)
- The toast template check and append it to the toast container

mparvazi: useToast is not static (`new Toast()`), for each request an instance will be created, before creating a toast, this modification will be done.

NeOMakinG: You'll end up with a function: `useToastContainer` for example, which will handle all the check and generation then return the `toast-container` element so you can use it

And another one: `useToast`, which will use `useToastContainer` on the top

NeOMakinG: instead of `useToastContainer`

What about a generic `useContainer` function? Taking a selector as a parameter, and a template content (for the fallback), and an `appendTo` selector? It can be used both for `useToast` and `useAlert`

mparvazi: I love generics :smile: 
I'll work on it.

src/js/helpers/useToast.ts
```diff
@@ -0,0 +1,131 @@
...
+      if (toastColor === 'white' || toastColor === 'light') {
+        btnClose.classList.add('btn-close-white');
+      }
+    }
+
+    btnClose.classList.remove('d-none');
+  }
+}
+
+export default useToast;
```
NeOMakinG, line 131: Just in order to remind me later, we'll need to expose all these components system to a global `window` attribute so anyone can use it without including core files of the theme

src/js/helpers/useToast.test.ts
NeOMakinG: Yeah agree

src/js/helpers/useToast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  containerComponent, btnCloseComponent, testMessage, testOptions, defaultType, validTemplate, invalidTemplate 
+} from '@constants/mocks/useToast-data';
+
+describe('useToast', () => {
+  const expectedOptions = {
+    getType(hasType: boolean) {
+
+      return hasType ? `.bg-${testOptions.type}` : `.bg-${defaultType}`;
+    },
+    getClassList() {
```
boubkerbribri, line 13: You can define these functions outside the describe so the test is easier to read.

```js
describe
  describe
    it
    it
  describe
    it
```

src/js/helpers/useToast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    it('should display the close button if autohide is false', () => {
+      useToast(testMessage, testOptions).show();
+      const toastElement = getToastElement(true);
+      const expectedBtnClose = expectedOptions.getBtnClose();
+
+      if (expectedBtnClose) {
+        const btnCloseElement = toastElement?.querySelector<HTMLButtonElement>(expectedBtnClose);
+
+        expect(btnCloseElement?.classList.contains('d-none')).toBe(false);
+      }
```
boubkerbribri, line 92: I don't understand much the use case here, so I don't know if you want that for the test : 
if `expectedBtnClose` is null, the test will pass (no assertion before). So if `expectedBtnClose` shouldn't be null, you can add an expect before and do your expect without a condition.

mparvazi: `const testAutoHideOptions = { type: 'danger', autohide: false };`

Screenshot from 2022-02-21 12-54-14

Wdyt?

boubkerbribri: Looks good to me.

mparvazi: Ty @boubkerbribri 

Your comment helped me to learn what's the purpose of the tests!

boubkerbribri (commented):
LGTM for tests!

boubkerbribri (approved):

src/js/constants/mocks/useToast-data.ts
```diff
@@ -0,0 +1,55 @@
...
+      <template class="js-toast-template">
+        <div class="toast" role="alert" aria-live="assertive" aria-atomic="true">
+          <div class="d-flex">
+            <div class="toast-body"></div>
+            <button type="button" class="btn-close me-2 m-auto d-none" data-bs-dismiss="toast"></button>
+          </div>
+        </div>
+      </template>
+    </div>
+    `;
```
NeOMakinG, line 50: Add this variable into something like `prestashop.templates.useToast.toast`

so it's overridable from another JS file

Same for the container bellow: `prestashop.templates.useToast.container`

Verify if `prestashop.templates` exist before, overwise you'll need to add it to `prestashop` before

NeOMakinG: Well, sorry, it's the wrong file, I was speaking about the not mocked data, the real fallback

src/js/helpers/useToast.ts
```diff
@@ -0,0 +1,172 @@
...
+      if (toastColor === 'white' || toastColor === 'light') {
+        closeButton.classList.add('btn-close-white');
+      }
+    }
+
+    closeButton.classList.remove('d-none');
+  }
+}
+
+export default useToast;
```
NeOMakinG, line 172: As it is mature enought, let's not add `useToast` to `window.prestashop.components` so every modules and child themes will be able to use it without having a JS transpiler

NeOMakinG:
2 little more things and we can merge it I think so, if you feel like it's mergeable?

Tell me if you didn't really understand or if you want me to do it, it will be a pleasure

src/js/constants/mocks/useToast-data.ts
mparvazi: Why a JS file should be able to override test data?

NeOMakinG: https://github.com/PrestaShop/theme-refacto/pull/184#discussion_r813754515

:trollface:

mparvazi: All data have been defined as const.
I think these constants must be managed by useToast only.

Developers can use their templates to style the toast, just need to have standard structure of BS Toast in tpl file.
For example they can add header for toast.
The JS files can use options for overriding styles.

mparvazi: If it's needed a JS file uses its template for toast, we can add a parameter to useToast for custom template:
`const useToast = (message: string, options?: Toaster.Option, template?: string)`

mparvazi:
> 2 little more things and we can merge it I think so, if you feel like it's mergeable?
> 
> Tell me if you didn't really understand or if you want me to do it, it will be a pleasure

I think something like this should be done by you, I don't know how to do it.

NeOMakinG:
@mparvazi do you think that it's mergeable now? I'll add what I said afterwhile

NeOMakinG (approved):

NeOMakinG:
Thanks @mparvazi 

I need to think about exposing it to a global object

NeOMakinG:
- We need the same system for alerts
- We need to expose it on a global object
- I need to dive into tests, maybe we can test some more things like unit testing little functions

Overwise, it's a really good job, appreciate it! Wp

mparvazi:
> * We need the same system for alerts
> 
>     * We need to expose it on a global object
> 
>     * I need to dive into tests, maybe we can test some more things like unit testing little functions
> 
> 
> Overwise, it's a really good job, appreciate it! Wp

When useToast exposed and tests completed, adding useAlert will be easy and fast.

PR #185, opened 2022-02-18
Improve product page

Hlavtox:


modules/ps_emailalerts/views/templates/hook/product.tpl
```diff
@@ -0,0 +1,46 @@
...
+* Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+* versions in the future. If you wish to customize PrestaShop for your
+* needs please refer to http://www.prestashop.com for more information.
+*
+* @author    PrestaShop SA <contact@prestashop.com>
+* @copyright 2007-2015 PrestaShop SA
+* @license   http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
+* International Registered Trademark & Property of PrestaShop SA
+*}
+
```
kpodemski, line 25: Take a look at the new version of this file:
https://github.com/PrestaShop/ps_emailalerts/blob/dev/views/templates/hook/product.tpl

templates/catalog/product.tpl
```diff
@@ -116,6 +117,7 @@
 
   {* SECOND PART - REASSURANCE, TABS *}
   <div class="row">
```
mparvazi, line 119: Just need to add some space before row.

```suggestion
  <div class="row mt-5">
```

mparvazi: Checked again, it's doesn't need any changes!
It works fine :100: 
Very nice :clap:

mparvazi:
@Hlavtox @NeOMakinG 

We can use vertical product-thumbnails for large screen, and horizontal for mobile.
Layout is full-width, so we have enough space for vertical thumbnails in large screen.

Wdyt?

templates/catalog/product.tpl
SharakPL: I meant: instead of `mt-5` :)

mparvazi: Try to to build and test this PR, there is no problem for spacing.
I got wrong because forget to clear cache!

Hlavtox:
@mparvazi Well, we tested it with Valentin. It looks nice with 4 images, but when I added like 15 images like I and other clients have on many products, it was a kilometer long. :-D :-(

Maybe with some image slider it could be vertical, but if we make it horizontal, we can do it simple and just show all the thumbnails. And I think it's more UX than to hide the other thumbnails, at least the people know that the photos are there. You would not believe how many customers just don't see, don't read...

mparvazi:
So the only problem is that, in xlarge screen may you can not see all part of the cover image in viewport.
```
    <div class="product__left col-lg-6 col-xl-7">
      {block name='product_cover_thumbnails'}
      {/block}
    </div>
```
I mean you can not see the bottom part of image, and when you try to scroll down, then you lost the top part of image, because the position of header is fixed.

I don't know are you planning to add zoom button for modal image view or not, because it somehow fix the problem.

NeOMakinG:
@Hlavtox we need to finish this!

NeOMakinG:
Let's go ahead on this, we will probably get back in when we will review it with the product team, if you see some improvements, don't hesitate to contribute or open an issue ❤️

templates/catalog/_partials/product-cover-thumbnails.tpl
```diff
@@ -67,14 +67,6 @@
     </div>
   </div>
 
-
```
Hlavtox: @NeOMakinG 🤣🤣🤣

Hlavtox: "Will you remove it after we finish?" "Of course dude" 😁

NeOMakinG: @Hlavtox thought the same while removing it, I was like "I'm not gonna let him do this anymore."

Hlavtox: 🤣🤣

modules/ps_emailalerts/views/templates/hook/product.tpl
```diff
@@ -0,0 +1,49 @@
+{*
+* 2007-2015 PrestaShop
```
matks, line 2: 2015 👀

PR #187, opened 2022-02-28
Expose useToast to window.prestashop

NeOMakinG:

### DIFF
--- a/src/js/expose-components.ts
+++ b/src/js/expose-components.ts
@@ -22,31 +22,31 @@
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
-/* eslint-disable */
-// @ts-ignore
-//import $ from "expose-loader?exposes=$,jQuery!jquery";
 
-import initEmitter from './prestashop';
-import 'bootstrap-input-spinner/src/bootstrap-input-spinner';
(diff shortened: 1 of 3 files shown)


src/js/expose-components.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
+
+import Toaster from '@constants/useToast-data';
```
PierreRambaud, line 26: Why you don't use webpack libraries instead?
We already use the `libraryTarget` parameter, so we should be able to create a bundle containing all our exposed components :thinking:

NeOMakinG: It will still output every chunks to window right? We will only need some components

I feel like using webpack for this is way more complicated than 15 lines of code, if you want to contribute, feel free :p

PierreRambaud (approved):

NeOMakinG:
@mparvazi merging so it's not blocking the CI PR anymore, if you have any changes that you see, don't hesitate to create an issue or provide a PR :)

PR #190, opened 2022-03-04
Improve useToast

mparvazi:


mparvazi:
The first PR in theme-refacto that has been passed all checks ;)

NeOMakinG:
> The first PR in theme-refacto that has been passed all checks ;)

I know that feeling, what a pleasure 😄

src/js/components/useToast.ts
```diff
@@ -87,21 +87,21 @@ const getToastElement = (template?: string): HTMLElement => {
...
 
 // We need to use a template, in order to generate the toast markup on the fly
-const cloneToastTemplate = (containerToAppend: HTMLElement, toastContainer?: HTMLElement): HTMLElement | false => {
+const cloneToastTemplate = (toastContainer: HTMLElement, overrideContainer?: HTMLElement): HTMLElement | false => {
   let toastTemplate;
 
-  if (toastContainer) {
-    toastTemplate = toastContainer.querySelector<HTMLTemplateElement>(selectorsMap.toast.template);
+  if (overrideContainer) {
+    toastTemplate = overrideContainer.querySelector<HTMLTemplateElement>(selectorsMap.toast.template);
```
NeOMakinG, line 94: I think this will cause some problems

Try to initialise a toast with the container already present in the page

Then try to override it with this

Then show both toast at the same time

NeOMakinG: For example:

```js
const {instance: test} = window.Theme.components.useToast('test')
const {instance: overrideTest} = window.Theme.components.useToast('overrideTest', options: {<put the overriden markup>})

test.show();
overrideTest.show();
```

One of these toasts will be under the other, because the container is duplicated :p

It works if you change the overridden container position, bottom: 0 for example, but not in the case you place the container at the same place than the already present one :P

NeOMakinG:
I'm not a big fan of the current implementation, I really feel like we should split the container logic from the toast logic, so we could be able to override one or the other, actually, the toast logic is inside the container logic in most of the code, wdyt?

src/js/components/useToast.ts
mparvazi: We create two instances and then call, so we should had two toasts at stack:

Screenshot from 2022-03-04 07-50-14

What's the problem?

src/js/components/useToast.test.ts
```diff
@@ -103,26 +103,26 @@ describe('useToast', () => {
...
-    it('should insert overriden template in existing empty container in the DOM', () => {
+    it('should insert overridden template in existing empty container in the DOM', () => {
       resetHTMLBodyContent(Toastify.WithContainerWithoutTemplate);
 
-      const toast = useToast('', {type: 'success', template: Toastify.Override});
+      const toast = useToast('', Toastify.TestTemplateOption);
       toast.instance.show();
       const toastContainer = toast.element.parentElement;
-      const overridenToast = toastContainer?.querySelectorAll<HTMLElement>(Toastify.OverridenToastClass);
+      const overrirdenToast = toastContainer?.querySelector<HTMLElement>(Toastify.OverriddenToastSelector);
```
NeOMakinG, line 122:
```suggestion
      const overriddenToast = toastContainer?.querySelector<HTMLElement>(Toastify.OverriddenToastSelector);
```

src/js/components/useToast.test.ts
```diff
@@ -103,26 +103,26 @@ describe('useToast', () => {
...
 
-      expect(toastContainer?.classList.contains(Toastify.FallbackContainerClass)).toBeFalsy();
-      expect(overridenToast?.length).toBe(1);
+      expect(toastContainer?.classList.contains(Toastify.OverriddenContainerClass)).toBeFalsy();
+      expect(overrirdenToast).toBeInstanceOf(HTMLElement);
```
NeOMakinG, line 125:
```suggestion
      expect(overriddenToast).toBeInstanceOf(HTMLElement);
```

mparvazi:
> I'm not a big fan of the current implementation, I really feel like we should split the container logic from the toast logic, so we could be able to override one or the other, actually, the toast logic is inside the container logic in most of the code, wdyt?

I think the container placement should be defined just by toast-container.tpl (and if not exists by useToast).
It shouldn't be changed by components override template.
Maybe a component wants to place toast at bottom-right, but this placement causes some accessibility problems in third-party theme.

src/js/components/useToast.ts
NeOMakinG: Oh ok, just understood that you didn't change anything about the logic I introduced, so it's fine

The overridden container will never be used, we should probably get rid of it as it's useless

NeOMakinG:
> > I'm not a big fan of the current implementation, I really feel like we should split the container logic from the toast logic, so we could be able to override one or the other, actually, the toast logic is inside the container logic in most of the code, wdyt?
> 
> I think the container placement should be defined just by toast-container.tpl (and if not exists by useToast). It shouldn't be changed by components override template. Maybe a component wants to place toast at bottom-right, but this placement causes some accessibility problems in third-party theme.

Ok, agree, but then we shouldn't be forced to send a container in the override then, we should probably change this in the implementation

mparvazi:
Important update:
Only the <template> tag in override-template option will be used (if it has container then will be dismissed).

When useToast not works:
1. The toast-container markup is not exists in the DOM and The JS fallback container is not valid
2. The template option used but its structure is not valid (it is not based on the BS5 standard toast structure)

mparvazi:
1. The override template has container but it will be dismissed, and the markup's container or JS fallback will be used, but the `<template>` will be used for toast clone:
```
const OverrideTemplate = `
  <div class="toast-container position-fixed bottom-0 end-0 p-2" id="js-toast-container">
    <template class="js-toast-template">
      <div class="toast">
        <div class="toast-header">Override</div>
        <div class="toast-body"></div>
      </div>
    </template>
  </div>
`;
```

2. The override template has not `<template>` then try to use current template in the DOM or if not exists use JS fallback:
```
const OverrideTemplate = `
  <div class="toast-container position-fixed bottom-0 end-0 p-2" id="js-toast-container">
  </div>
`;

**Comment by NeOMakinG** [comment]
> 1. The override template has container but it will be dismissed, and the markup's container or JS fallback will be used, but the `<template>` will be used for toast clone:
> 
> ```
> const OverrideTemplate = `
>   <div class="toast-container position-fixed bottom-0 end-0 p-2" id="js-toast-container">
>     <template class="js-toast-template">
>       <div class="toast">
>         <div class="toast-header">Override</div>
>         <div class="toast-body"></div>
>       </div>
>     </template>
>   </div>
> `;
> ```
> 
> 2. The override template has not `<template>` then try to use current template in the DOM or if not exists use JS fallback:
> 
> ```
> const OverrideTemplate = `
>   <div class="toast-container position-fixed bottom-0 end-0 p-2" id="js-toast-container">
>   </div>
> `;
> ```

Sending a container that will never be used feels really weird, don't you think?

**Comment by mparvazi** [comment]
@NeOMakinG 
> Sending a container that will never be used feels really weird, don't you think?

It's weird, but need to consider these weird things to make useToast stable.

Even I'm thinking about should we check override template structure before use or not?
When override template is wrong useToast throws an error, so what should we do in this situation?
1. Throw an error (as it does currently)
2. Using the fallback template

**Comment by mparvazi** [comment]
Can we use `console.error()` or `console.warn()` instead of `throw`?

Screenshot from 2022-03-07 13-14-43

In this way useToast doesn't stop theme script due to an error.

**Comment by NeOMakinG** [comment]
> Can we use `console.error()` or `console.warn()` instead of `throw`?
> 
> Screenshot from 2022-03-07 13-14-43
> 
> In this way useToast doesn't stop theme script due to an error.

yeah feel free :) way better

**Comment by NeOMakinG** [comment]
> @NeOMakinG
> 
> > Sending a container that will never be used feels really weird, don't you think?
> 
> It's weird, but need to consider these weird things to make useToast stable.
> 
> Even I'm thinking about should we check override template structure before use or not? When override template is wrong useToast throws an error, so what should we do in this situation?
> 
> 1. Throw an error (as it does currently)
> 2. Using the fallback template
> 
> For example:
> 
> ```
> const OverrideTemplate = `
>     <template class="js-toast-template">
>       <div class="toast">
>         <div class="toast-header"></div>
>       </div>
>     </template>
> `;
> ```
> 
> The `toast-body` in override template is not exists, so useToast throws an error.

Can't we consider sending the toast markup directly and use it instead of using a never-used container with template?
I think we shouldn't use the JS fallback if the override is invalid, just throw an error, people should remove the override if they send something invalid, or fix it!

File: src/js/components/useToast.ts [hunk 4]
```diff
@@ -43,20 +61,24 @@ const useToast = (message: string, options?: Toaster.Options): Toaster.Result |
...
-      const toastObject: Toaster.Result = {
-        // It's a Bootstrap.Toast instance, will be used to show, hide and remove the toast
-        instance: new Toast(toastElement, {autohide: clonedOptions.autohide, delay: clonedOptions.delay}),
-        // In case someone wants to modify the toast markup afterwhile
-        element: toastElement,
-        // In case someone wants to modify the text content afterwhile
-        content: toastElementBody,
+      // It's a Bootstrap's Toast instance, will be used to show, hide and remove the toast
+      const instance = new Toast(toastElement, {autohide: clonedOptions.autohide, delay: clonedOptions.delay});
+      toastObject = {
```
NeOMakinG, line 66:
```suggestion
      toastObject = {
        instance,
```

src/js/components/useToast.test.ts
```diff
@@ -37,92 +37,94 @@ describe('useToast', () => {
 
     it('should display the HTML markup message when used', () => {
       const toast = useToast(Toastify.TestMarkupMessage);
-      toast.instance.show();
```
NeOMakinG, line 40:
```
const {instance: toast} = useToast(Toastify.TestMarlupMessage);
toast.show()
```

Is it that hard to understand destructuring?

src/js/components/useToast.test.ts
```diff
@@ -37,92 +37,94 @@ describe('useToast', () => {
...
       expect(toastMessage === expectedToastMessage).toBeTruthy();
     });
 
     it('should style the toast as same as the type when is set', () => {
       const toast = useToast('', Toastify.TestTypeOption);
-      toast.instance.show();
-
-      const toastClassList = toast.element.classList.toString();
+      toast.show();
+      const toastElement = toast.element();
```
NeOMakinG, line 51:
```suggestion
      const toastElement = toast.element();
```

Why would we need `toast.element()` while you can get the toast element using destructuring when you initialize the toast like

```
const {element} = useToast(...)
```

NeOMakinG:
@mparvazi 

I don't like the fact that we need to use a getter to get the `element` like `toast.element()` while we could get it when we initialise our `useToast` composable like `const {element, instance: myToast} = useToast(...)`

src/js/components/useToast.test.ts
mparvazi: In above code, you used bootstrap show(), if the useToast not initialized correctly (invalid template or invalid markup) it will stop JS execution.

mparvazi: Bootstrap doesn't control toast instance before  show() or hide();

NeOMakinG: Inside React/Vue composables, the good practice is to have an error that you can get while destructuring which is undefined if there are no errors

```
const {instance: toast, element, error} = useToast(Toastify.TestMarlupMessage);

if (!error) {
  toast.show()
  element.remove()
}
```

where `error` would be something like `'The toast markup override is invalid.'`

NeOMakinG: You can also verify that the element exist

```
if (!toast) return;
```

or 
```
toast?.show();
```

mparvazi: You can not get error after initializing, Bootstrap throws error in runtime.
`new Toast('.dummy');`
Instance will be created without error by BS Toast, but ".dummy" is not exists in the DOM.
So using `show()` throws an error.

src/js/components/useToast.ts
```diff
@@ -27,112 +27,141 @@ import {Toast} from 'bootstrap';
...
+    }
+  }
+
+  // This happens only if someone removed the markup AND removed the fallback
+  printConsoleError(
+    'Failed to clone toast template.',
+    'Check the toast template in theme markup or JS fallback.',
+  );
+
+  return undefined;
```
NeOMakinG, line 131:
```suggestion
  return;
```

It will do the same I think so

src/js/components/useToast.ts
```diff
@@ -27,112 +27,141 @@ import {Toast} from 'bootstrap';
...
   }
 
-  // This happen only if someone removed the markup AND removed the JS fallback
-  throw new DOMException('The object can not be cloned.');
+  printConsoleError(
+    'The override toast template is not valid.',
+    'Reference: https://getbootstrap.com/docs/5.0/components/toasts/',
+  );
+
+  return undefined;
```
NeOMakinG, line 152:
```suggestion
  return;
```

src/js/components/useToast.ts
```diff
@@ -142,24 +171,18 @@ const useFallbackToastContainer = (template?: string): HTMLElement => {
     }
   }
 
-  throw new DOMException('The object can not be cloned.');
+  return undefined;
```
NeOMakinG, line 174:
```suggestion
  return;
```

src/js/components/useToast.test.ts
NeOMakinG: Ok, just tested

Can we at least make these override methods (show, hide, dispose and more) inside a `toast` property that the `useToast` object return ?

So you will be able to use 

`const {element, toast, content} = useToast(...)` where toast === `{show, hide, dispose, instance}` and instance === Bootstrap.Toast.instance ?

src/js/components/useToast.ts
NeOMakinG: If it's the case, ok don't worry

src/js/components/useToast.test.ts
```diff
@@ -36,93 +36,88 @@ describe('useToast', () => {
     });
 
     it('should display the HTML markup message when used', () => {
-      const toast = useToast(Toastify.TestMarkupMessage);
-      toast.instance.show();
-
-      const toastMessage = toast.content.innerHTML;
+      const {instance, element, content} = useToast(Toastify.TestMarkupMessage);
+      instance?.show();
```
NeOMakinG, line 40:
```suggestion
      const {instance, element, content, show: showToast} = useToast(Toastify.TestMarkupMessage);
      showToast();
```

Use every methods like this, so you don't need to use the optional chaining

NeOMakinG (approved):

NeOMakinG:
Thanks a lot @mparvazi, great work!

PR #197, opened 2022-03-11
Introduce useAlert

mparvazi:
Bootstrap Alert API is simpler than Toast, so I tried to add some capabilities to make it similar as Toast.

**Just for review, not tested completely.**

Screenshot 2022-03-11 at 16-27-39 Cart


NeOMakinG (approved):
Looks fine to me

PR #206, opened 2022-03-16
Add toasts and useToast documentation

NeOMakinG:
| Questions         | Answers
| Description?      | We need to document toasts and useToast
| BC breaks?        | no
| Deprecations?     | no

## Quickview
image

### DIFF
--- a/stories/theme/javascript/components/useToast/index.stories.mdx
+++ b/stories/theme/javascript/components/useToast/index.stories.mdx
@@ -4,7 +4,7 @@ import useMultipleContent from './multiple.html'
 
 <Meta title="Theme/JavaScript/Components/useToast" />
 
-# useToast JavaScript usage
+# useToast
 
 **useToast** is accessible through `window.Theme.components.useToast`.
 
--- a/stories/theme/javascript/components/useToast/index.stories.mdx
+++ b/stories/theme/javascript/components/useToast/index.stories.mdx
@@ -8,6 +8,18 @@ import useMultipleContent from './multiple.html'
 
 **useToast** is accessible through `window.Theme.components.useToast`.
 
+## Concept
+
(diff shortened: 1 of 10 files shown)


NeOMakinG:
@Hlavtox @mparvazi @SharakPL

Feel free to review if you want (you're not forced)

stories/theme/javascript/components/useToast/index.stories.mdx
```diff
@@ -0,0 +1,191 @@
...
+  content: HTMLElement | null;
+  // Bootstrap methods wrapped
+  show: () => boolean;
+  hide: () => boolean;
+  dispose: () => boolean;
+  remove: () => boolean;
+  message: (markup?: string) => string | boolean;
+}
+```
+
```
NeOMakinG, line 47: This is probably the most important part to review

mparvazi:
dispose() will be useful when you just have a toast inside the DOM (without template):

<div class="toast-container">
    <div class="toast">...</div>
</div>

So when you call dispose() then you can create another toast instance with this div.

But in useToast we used template for cloning and appending in the DOM.
I mean when you call useToast you have a corresponding toast element in the DOM.

I think BS developers forget that something like <template> is more flexible so they defined something like dispose()

NeOMakinG:
> dispose() will be useful when you just have a toast inside the DOM (without template):
> 
> ```
> <div class="toast-container">
>     <div class="toast">...</div>
> </div>
> ```
> 
> So when you call dispose() then you can create another toast instance with this div.
> 
> But in useToast we used template for cloning and appending in the DOM. I mean when you call useToast you have a corresponding toast element in the DOM.
> 
> I think BS developers forget that something like `<template>` is more flexible so they defined something like dispose()

At least, we leave the liberty to anyone, to use whatever they wants :)

NeOMakinG:
This is not blocking anything, let's just merge this and improve it later if needed

PR #208, opened 2022-03-17
Improve readme title

matks:
| Questions         | Answers
| Description?      | Improve readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# PrestaShop 8.0
+# Hummingbird theme
 
-This theme is the new theme of PrestaShop. Please, if you work on this theme, use the `develop` branch of PrestaShop, because this theme will be available on 8.0
+This is the new theme of PrestaShop we are working on. Please, if you work on this theme, use the `develop` branch of PrestaShop, because this theme will be available from version 8.0.
 
 [Read more](https://build.prestashop.com/news/new-theme-announce/) about this theme on the blog.
 


NeOMakinG (approved):

kpodemski:
@matks, shouldn't the repo name be `hummingbird-theme`?

NeOMakinG:
> @matks, shouldn't the repo name be `hummingbird-theme`?

If we do this, we are forced to `git clone <urlrpo> hummingbird` because of the config.yml, and add the word `theme` to a subfolder of `themes` looks wrong

NeOMakinG:
@kpodemski if you want to talk about it, open an issue or slack!

kpodemski:
@NeOMakinG 

it's exactly the same with the `classic`, this is why I thought we want to follow this naming convention 👍 

not that important tbh

NeOMakinG:
> @NeOMakinG
> 
> it's exactly the same with the `classic`, this is why I thought we want to follow this naming convention 👍
> 
> not that important tbh

Yeah and this is exactly why I said to @matks that classic repository shouldn't be named like this :(

NeOMakinG:
> Maybe we need to rename `classic-theme` to `classic` then 🤔

@eternoendless wdyt? Do we have some constraint about this?

eternoendless:
I named classic-theme that way because "classic" is a generic word and I thought it could be confusing otherwise. I'm not sure if we can rename the source of a composer package in packagist. Perhaps we can.

The package is still named `prestashop/classic` because the package name determines the installation directory and that unfortunately cannot be changed. Plus, the translation system does not support dashes in the module/theme name either.

The git clone issue is just a small inconvenience in my opinion, anyway the normal way to include a theme should be through composer (unless you're planning on working on the theme itself).

PR #213, opened 2022-03-18
Add steps component to checkout page

NeOMakinG:
| Questions         | Answers
| Description?      | The new checkout page has some steps on the top
| BC breaks?        | no
| Deprecations?     | no

### Desktop
image

### Mobile
image

## Progress Circle component

We needed to introduce a Progress Circle component because the JS will change it on the top on mobile

It's not testable because JSDom is not supporting SVGs elements...

image

### DIFF
--- a/src/scss/partials/variables/bootstrap/_variable-overrides.scss
+++ b/src/scss/partials/variables/bootstrap/_variable-overrides.scss
@@ -1,7 +1,6 @@
 // IT IS FORBIDDEN TO USE REFERENCES TO BS5 VARIABLES BECAUSE THEY ARE NOT YET DECLARED.
 $primary: #5434d0;
 $secondary: #f39d72;
-$info: #f39d72;
 $success: #21834D;
 $warning: #ff9a52;
 $danger: #ff4c4c;
--- a/src/scss/partials/variables/bootstrap/_variable-overrides.scss
+++ b/src/scss/partials/variables/bootstrap/_variable-overrides.scss
@@ -1,7 +1,8 @@
 // IT IS FORBIDDEN TO USE REFERENCES TO BS5 VARIABLES BECAUSE THEY ARE NOT YET DECLARED.
 $primary: #5434d0;
 $secondary: #f39d72;
-$success: #4cbb6c;
+$info: #f39d72;
+$success: #21834D;
 $warning: #ff9a52;
 $danger: #ff4c4c;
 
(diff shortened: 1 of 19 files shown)


src/scss/partials/variables/bootstrap/_variable-overrides.scss
```diff
@@ -1,7 +1,6 @@
 // IT IS FORBIDDEN TO USE REFERENCES TO BS5 VARIABLES BECAUSE THEY ARE NOT YET DECLARED.
```
SharakPL, line 1: The way I always do it is copy all bootstrap _variables.scss into theme and use this **instead** original variables, not before them. This way you can use references to BS5 variables already declared in this file without problems. Plus it gives better overview of what can be changed within BS theme variables and helps to keep the project clean, without duplicated code and useless variables like it was with classic.

NeOMakinG: This is something we should discuss in a proper issue, and it will probably change a lot of files in our variable folder

NeOMakinG:
@SharakPL @mparvazi @Hlavtox 

I prefer to calculate the initial state of the component inside the smarty file so:
- It doesn't force you to use some javascript in case you want to display a progress circle
- It greatly simplifies the JS codebase

I'm forced to set some CSS styles inside the markup in order to achieve this, but I think it's worth honestly.

Feel free to review, every feedback is welcome!

mparvazi:
Screenshot 2022-03-21 at 15-21-42 develop

## My suggestion:

**.checkout__steps:**
border-top => none
background => white
border-bottom-color => $gray-200 (as like as header-border-color)

**#notifications:**

Screenshot from 2022-03-21 15-43-50


Screenshot 2022-03-21 at 15-46-13 develop

Screenshot 2022-03-21 at 15-47-14 develop

NeOMakinG:
> IMO it's better to use number only for checked steps:
> 
> Screenshot 2022-03-21 at 15-58-35 develop

Applied this

NeOMakinG:
@mparvazi As I don't wan't to decide on design things what I did it this:
image

And I'll try to catch up @TristanLDD in order to speak about your suggestion

NeOMakinG:
If someone has something against this PR, please create issues in order to speak about it or feel free to contribute

Hlavtox:
It's awesome, merge it. Polishing later :D

anjohnb:
Thank you

PR #215, opened 2022-03-21
Refacto qty input

mparvazi:
Screenshot 2022-03-21 at 17-28-44 Cart

### DIFF
--- a/src/js/constants/selectors-map.ts
+++ b/src/js/constants/selectors-map.ts
@@ -55,8 +55,12 @@ export const cart = {
 
 const selectorsMap = {
   qtyInput: {
-    default: '.js-quantity-button input',
-    modal: '.modal-dialog .js-quantity-button input',
+    default: '.js-quantity-button',
+    modal: '.modal-dialog .js-quantity-button',
+    increment: '.js-increment-button',
+    decrement: '.js-decrement-button',
+    icon: '.material-icons',
+    spinner: '.spinner-border',
   },
   alert: {
     selector: '#notifications .container',
@@ -83,7 +87,9 @@ const selectorsMap = {
   facetedsearch,
   pageLoader,
   listing,
-  cart,
+  cart: {
+    alert: '#js-product-line-alert--{pid}',
+  },
 };
 
 export default selectorsMap;
--- a/src/js/qty-input.ts
+++ b/src/js/qty-input.ts
@@ -23,23 +23,106 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
 
-export default function initQuantityInput(selector: string) {
-  $(document).ready(() => {
-    $(selector).inputSpinner({
-      decrementButton: '<i class="material-icons">expand_more</i>',
-      incrementButton: '<i class="material-icons">expand_less</i>',
-      buttonsClass: '',
-      buttonsWidth: '1.25rem',
-      /* eslint-disable */
-      template: 
-          '<div class="input-group ${groupClass}">' +
-            '<input type="text" inputmode="decimal" style="text-align: ${textAlign}" class="form-control form-control-text-input"/>' +
-            '<div class="qty-right">' +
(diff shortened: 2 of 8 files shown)


src/js/constants/selectors-map.ts
```diff
@@ -83,7 +87,9 @@ const selectorsMap = {
   facetedsearch,
   pageLoader,
   listing,
-  cart,
+  cart: {
+    alert: '#js-product-line-alert--{pid}',
```
NeOMakinG, line 91: What do you want to do there? Looks like something is wrong

NeOMakinG: Also, if we could avoid to use an id, I would prefer

src/js/qty-input.ts
```diff
@@ -23,23 +23,106 @@
...
+
+      toggleSpinner(targetButton, targetButtonIcon, targetButtonSpinner);
+
+      sendUpdateCartRequest(requestUrl)
+        .then((response) => {
+          toggleSpinner(targetButton, targetButtonIcon, targetButtonSpinner);
+
+          return response.json();
+        })
+        .then((data) => {
```
NeOMakinG, line 71: Let's use async/await instead of chaining promise's callback

src/js/qty-input.ts
```diff
@@ -23,23 +23,106 @@
...
+          const {productId: pid} = qtyInput.dataset;
+          let alertSelector: string;
+
+          if (pid) {
+            alertSelector = selectorsMap.cart.alert.replace('{pid}', pid);
```
NeOMakinG, line 76: Please, prefer using an arrow function inside the selectors map so you can use it like `selectorsMap.cart.alert(pid)` and insert it into the selector like `#my-selector-${pid}`

There are several examples in the map

src/js/qty-input.ts
```diff
@@ -23,23 +23,106 @@
...
+              useAlert(error, {type: 'danger', selector: alertSelector}).show();
+            });
+          } else {
+            prestashop.emit('updateCart', {
+              reason: qtyInput.dataset,
+              resp: data,
+            });
+          }
+        })
+        .catch((error) => {
```
NeOMakinG, line 96: It will be replaced by a try/catch when using async/await

NeOMakinG (changes requested):
I think we can improve it by making it reusable

Imagine that a module want's to use a qty input for something else than a product qty

What do you think of making it reusable like we did for `useProgressRing` for example? We don't need to generate the markup, but you can pass your own selector, with your own callback value, so you can execute your own logic when increasing/decreasing

Because I'm pretty sure some modules will wants to use the quantity input I think

Also, at the same time, some comments about the current implementation, they may become useless depending on the new implementation

src/js/qty-input.ts
mparvazi: What's the problem behind this implementation?

sendUpdateCartRequest is a async/await function.

first of all `.then((response)` will be called, spinners will be reset:
1. if there is a response then the json data will be sent to `.then((data)`
2. else `.catch((error)` will be called.

mparvazi:
> I think we can improve it by making it reusable
> 
> Imagine that a module wants to use a qty input for something else than a product qty
> 
> What do you think of making it reusable as we did for `useProgressRing` for example? We don't need to generate the markup, but you can pass your own selector, with your own callback value, so you can execute your own logic when increasing/decreasing
> 
> Because I'm pretty sure some modules will want to use the quantity input
> 
> Also, at the same time, some comments about the current implementation, may become useless depending on the new implementation

Ok, just need to think about that.

src/js/qty-input.ts
NeOMakinG: Overall, it's a good practice to use async/await now because you absolutely never end by having callbacks hell, and it's easier to read

As you said `sendUpdateCartRequest` is an async function, it means that you can do something like:

```ts
try {
  const response = await sendUpdateCartRequest(requestUrl);
  const datas = await response.json();
  
  return <whateverYouWant>
} catch(error) {
  return new Error('Something went wrong');
}
```

NeOMakinG: It's way more readable than having multiple callbacks

PR #217, opened 2022-03-23
feat: checkout content adjustement (not full rework yet)

NeOMakinG:
## How does it look?
https://www.awesomescreenshot.com/video/8042555?key=6b104d500f9d58909ac56ecb1ba7edc9


templates/components/checkout-steps.tpl
```diff
@@ -82,7 +82,7 @@
         </p>
       </div>
 
-      <div class="{$componentName}__step" data-step="2">
+      <div class="{$componentName}__step" data-step="checkout-addresses-step">
```
Hlavtox, line 85: @NeOMakinG Is this correct? The d-none thing on steps

NeOMakinG: The d-none? Steps are hidden if it's not the current and it's toggled using JS if you switch to another one

Hlavtox: Yes, but why is specifically the second one displayed by default?

NeOMakinG: Did hide every steps by default now

matthieu-rolland (commented):
One question, lgtm otherwise 👍

NeOMakinG:
It's now ready for review

I would like to pass some time with @Hlavtox on this in order to check that the markup is fine enough, pretty sure we can find some improvements on it!

NeOMakinG:
As we're currently peer programming on it with Hlavtox, we're merging this, if you want to review it, do it, it will be taken under consideration inside another PR

PR #221, opened 2022-03-28
Improve useProgressRing

mparvazi:


src/js/components/useProgressRing.ts
```diff
@@ -33,19 +34,38 @@ export interface ProgressRingReturn {
...
-    const circle = element.querySelector<SVGCircleElement>(ProgressRingMap.checkout.circle);
+export const TextType = {
+  enum: 'enum',
+  percent: 'percent',
+};
+
+export interface ProgressRingOptions {
+  steps: number;
+  text: keyof typeof TextType;
+}
```
NeOMakinG, line 45:
```suggestion
export enum TextType {
  enum: 'enum',
  percent: 'percent',
};

export interface ProgressRingOptions {
  steps: number;
  text: TextType;
}
```

Does it works like this?

src/js/components/useProgressRing.ts
```diff
@@ -33,19 +34,38 @@ export interface ProgressRingReturn {
   error?: Error
```
NeOMakinG, line 34:
```suggestion
  error?: Error
  progressElement: HTMLElement
```

src/js/components/useProgressRing.ts
```diff
@@ -33,19 +34,38 @@ export interface ProgressRingReturn {
...
         circle.style.strokeDashoffset = offset.toString();
+        circle.dataset.percent = String(percent);
+
+        if (progressText) {
+          const text = (options.text === TextType.enum) ? `${index} / ${options.steps}` : `${percent}%`;
+          progressText.innerHTML = text;
+        }
       };
 
       return {
```
NeOMakinG, line 71:
```suggestion
      return {
        progressElement,
```

templates/components/progress-circle.tpl
```diff
@@ -48,14 +46,12 @@
...
     stroke="currentColor"
     stroke-width="{$stroke}"
-    data-percent="50"
+    data-percent="0"
     fill="transparent"
     r="{$radius}"
     cx="{$size / 2}"
     cy="{$size / 2}"
-    style="stroke-dasharray: {$circumference}, {$circumference}; stroke-dashoffset: {$dashoffset};"
+    style="stroke-dasharray: {$circumference}, {$circumference}; stroke-dashoffset: {$circumference};"
```
NeOMakinG, line 54: Are you sure? It's working fine?

src/js/components/useProgressRing.ts
mparvazi: In modern TypeScript, you may not need an enum when an object with as const could suffice:
https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums

but here i forget to add `as const`  to the TextType object!

templates/components/progress-circle.tpl
mparvazi: In initial state (0%) the circle should be gray.
{$dashoffset = $circumference - (($percent / 100 ) * $circumference)}
so when $percent is 0 => $dashoffset = $circumference

src/js/components/useProgressRing.ts
NeOMakinG: But using an enum for such a thing looks like a good practice on different projects

mparvazi: I've tested, there is a problem with eslint when using enum here.
I think it's a bug in eslint:
https://stackoverflow.com/questions/63961803/eslint-says-all-enums-in-typescript-app-are-already-declared-in-the-upper-scope

NeOMakinG: Isn't it related to the enum keyword that should be named differently?

mparvazi: Using object with `as const` is safer:
Screenshot from 2022-03-29 20-35-33

NeOMakinG:
@mparvazi what's up with this?

mparvazi:
Can not use enum with eslint.

NeOMakinG:
> Can not use enum with eslint.

I guess it's not related, it was because it was already defined?

Do you want me to get a look and open this PR again? Why did you close it?

mparvazi:
@NeOMakinG 

Try to use an enum inside a component and test with eslint.

NeOMakinG:
@mparvazi Fixed it :)

I guess we're near the merge right?

mparvazi:
no-shadow: true

then:
```
var a = 3;
function b() {
    var a = 10;
}
```
In this case, the variable a inside of b() is shadowing the variable a in the global scope. 
no problem about this?

mparvazi:
OK, so let me check again.

mparvazi:
@NeOMakinG 
Ready to merge.

NeOMakinG:
@mparvazi let me test it tomorrow and it looks pretty fine to me now :)

NeOMakinG (approved):
The implementation looks fine to me, also tested it and it works fine

As we discussed in private weeks ago, we should think about testing `useProgress` and `useAlert` if it's possible

PR #222, opened 2022-03-28
Fix: language and currency selectors, checkout js bugs and improvements

NeOMakinG:
| Questions         | Answers
| Description?      | Language and currency selectors, miss placed div in checkout, checkout back button, first tab rename of checkout personal information, use alert on order confirm message
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no

### DIFF
--- a/src/js/modules/blockcart.ts
+++ b/src/js/modules/blockcart.ts
@@ -47,7 +47,7 @@ prestashop.blockcart.showModal = (html: string) => {
   const mainElement = document.createElement('div');
   mainElement.innerHTML = html;
   // Select the template of the modal
-  const modalTemplate = document.querySelector<HTMLElement>(BlockcartMap.modal);
+  const modalTemplate = mainElement.querySelector<HTMLElement>(BlockcartMap.modal);
 
   if (modalTemplate) {
     // If the template exist, append it to the body
--- a/src/js/modules/blockcart.ts
+++ b/src/js/modules/blockcart.ts
@@ -26,11 +26,12 @@ import {Modal} from 'bootstrap';
 import {isHTMLElement} from '@helpers/typeguards';
 
 const {prestashop} = window;
-const {blockcart: BlockcartMap} = prestashop.themeSelectors;
 
 prestashop.blockcart = prestashop.blockcart || {};
 
 prestashop.blockcart.showModal = (html: string) => {
+  const {blockcart: BlockcartMap} = prestashop.themeSelectors;
+
   function getBlockCartModal() {
     const blockCartModal = document.querySelector<HTMLElement>(BlockcartMap.modal);
 
--- a/src/js/modules/blockcart.ts
+++ b/src/js/modules/blockcart.ts
@@ -23,14 +23,16 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
 import {Modal} from 'bootstrap';
+import {isHTMLElement} from '@helpers/typeguards';
 
 const {prestashop} = window;
+const {blockcart: BlockcartMap} = prestashop.themeSelectors;
 
 prestashop.blockcart = prestashop.blockcart || {};
 
 prestashop.blockcart.showModal = (html: string) => {
   function getBlockCartModal() {
-    const blockCartModal = <HTMLElement>document.querySelector('#blockcart-modal');
+    const blockCartModal = document.querySelector<HTMLElement>(BlockcartMap.modal);
 
     return blockCartModal;
   }
@@ -43,23 +45,30 @@ prestashop.blockcart.showModal = (html: string) => {
 
   const mainElement = document.createElement('div');
   mainElement.innerHTML = html;
+  // Select the template of the modal
+  const modalTemplate = document.querySelector<HTMLElement>(BlockcartMap.modal);
 
-  document.querySelector('body')?.append(<HTMLElement>mainElement.querySelector('#blockcart-modal'));
+  if (modalTemplate) {
+    // If the template exist, append it to the body
+    document.querySelector('body')?.append(modalTemplate);
+  }
 
   blockCartModal = getBlockCartModal();
 
-  const modal = new Modal(blockCartModal);
+  if (blockCartModal) {
+    const modal = new Modal(blockCartModal);
 
-  modal.show();
+    modal.show();
 
-  blockCartModal.addEventListener('hidden.bs.modal', (event: Event) => {
-    const target = <HTMLElement>event.currentTarget;
+    blockCartModal.addEventListener('hidden.bs.modal', (event: Event) => {
+      const target = event.currentTarget;
 
-    if (target) {
-      prestashop.emit('updateProduct', {
-        reason: target.dataset,
-        event,
-      });
-    }
-  });
+      if (isHTMLElement(target)) {
+        prestashop.emit('updateProduct', {
+          reason: target.dataset,
+          event,
+        });
+      }
+    });
+  }
 };
--- a/src/js/modules/ps_currencyselector.ts
+++ b/src/js/modules/ps_currencyselector.ts
@@ -38,7 +38,7 @@ const initCurrencySelector = () => {
 document.addEventListener('DOMContentLoaded', () => {
   initCurrencySelector();
 
-  prestashop.on('responsive update', () => {
+  prestashop.on('responsiveUpdate', () => {
     initCurrencySelector();
   });
 });
--- a/src/js/modules/ps_currencyselector.ts
+++ b/src/js/modules/ps_currencyselector.ts
@@ -23,10 +23,9 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
 
-const {prestashop} = window;
-const {currencySelector: CurrencySelectorMap} = prestashop.themeSelectors;
-
 const initCurrencySelector = () => {
+  const {prestashop} = window;
+  const {currencySelector: CurrencySelectorMap} = prestashop.themeSelectors;
   const languageSelector = document.querySelector<HTMLElement>(CurrencySelectorMap.currencySelector);
 
   languageSelector?.addEventListener('change', (event) => {
@@ -35,13 +34,5 @@ const initCurrencySelector = () => {
     window.location.href = option.value;
   });
 };
-
-document.addEventListener('DOMContentLoaded', () => {
-  initCurrencySelector();
-
-  prestashop.on('responsiveUpdate', () => {
-    initCurrencySelector();
-  });
-});
 
 export default initCurrencySelector;
--- a/src/js/modules/ps_currencyselector.ts
+++ b/src/js/modules/ps_currencyselector.ts
@@ -24,9 +24,10 @@
  */
 
 const {prestashop} = window;
+const {currencySelector: CurrencySelectorMap} = prestashop.themeSelectors;
 
 const initCurrencySelector = () => {
-  const languageSelector = document.querySelector<HTMLElement>('.js-currency-selector');
+  const languageSelector = document.querySelector<HTMLElement>(CurrencySelectorMap.currencySelector);
 
   languageSelector?.addEventListener('change', (event) => {
     const option = event.target as HTMLOptionElement;
--- a/src/js/modules/ps_currencyselector.ts
+++ b/src/js/modules/ps_currencyselector.ts
@@ -1,4 +1,4 @@
-{**
+/**
  * Copyright since 2007 PrestaShop SA and Contributors
  * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
  *
@@ -21,28 +21,26 @@
  * @author    PrestaShop SA and Contributors <contact@prestashop.com>
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
- *}
+ */
 
-<section id="js-checkout-summary" class="accordion-collapse collapse d-lg-block card cart-summary mb-4 js-cart" data-refresh-url="{$urls.pages.cart}?ajax=1&action=refresh">
-  <div class="card-block">
-    {block name='hook_checkout_summary_top'}
-      {include file='checkout/_partials/cart-summary-top.tpl' cart=$cart}
-    {/block}
+const {prestashop} = window;
 
-    {block name='cart_summary_products'}
-      {include file='checkout/_partials/cart-summary-products.tpl' cart=$cart}
-    {/block}
+const initCurrencySelector = () => {
+  const languageSelector = document.querySelector<HTMLElement>('.js-currency-selector');
 
-    {block name='cart_summary_subtotals'}
-      {include file='checkout/_partials/cart-summary-subtotals.tpl' cart=$cart}
-    {/block}
-  </div>
+  languageSelector?.addEventListener('change', (event) => {
+    const option = event.target as HTMLOptionElement;
 
-  {block name='cart_summary_totals'}
-    {include file='checkout/_partials/cart-summary-totals.tpl' cart=$cart}
-  {/block}
+    window.location.href = option.value;
+  });
+};
 
-  {block name='cart_summary_voucher'}
-    {include file='checkout/_partials/cart-voucher.tpl'}
-  {/block}
-</section>
+document.addEventListener('DOMContentLoaded', () => {
+  initCurrencySelector();
+
+  prestashop.on('responsive update', () => {
+    initCurrencySelector();
+  });
+});
+
+export default initCurrencySelector;
--- a/src/js/modules/ps_languageselector.ts
+++ b/src/js/modules/ps_languageselector.ts
@@ -23,10 +23,9 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
 
-const {prestashop} = window;
-const {languageSelector: LanguageSelectorMap} = prestashop.themeSelectors;
-
 const initLanguageSelector = () => {
+  const {prestashop} = window;
+  const {languageSelector: LanguageSelectorMap} = prestashop.themeSelectors;
   const languageSelector = document.querySelector<HTMLElement>(LanguageSelectorMap.languageSelector);
 
   languageSelector?.addEventListener('change', (event) => {
@@ -35,13 +34,5 @@ const initLanguageSelector = () => {
     window.location.href = option.value;
   });
 };
-
-document.addEventListener('DOMContentLoaded', () => {
-  initLanguageSelector();
-
-  prestashop.on('responsiveUpdate', () => {
-    initLanguageSelector();
-  });
-});
 
 export default initLanguageSelector;
--- a/src/js/modules/ps_languageselector.ts
+++ b/src/js/modules/ps_languageselector.ts
@@ -24,9 +24,10 @@
  */
 
 const {prestashop} = window;
+const {languageSelector: LanguageSelectorMap} = prestashop.themeSelectors;
 
 const initLanguageSelector = () => {
-  const languageSelector = document.querySelector<HTMLElement>('.js-language-selector');
+  const languageSelector = document.querySelector<HTMLElement>(LanguageSelectorMap.languageSelector);
 
   languageSelector?.addEventListener('change', (event) => {
     const option = event.target as HTMLOptionElement;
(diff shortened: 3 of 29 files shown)


matthieu-rolland (approved):

mparvazi:
The `next-button` should be on the right AND `forward-icon` should be on the right side of `next-button`:

Screenshot 2022-03-29 at 15-19-08 develop

mparvazi:
@NeOMakinG 

I think it's better to show this alert in top of the action buttons:

Screenshot 2022-03-29 at 20-52-55 develop

`templates/checkout/_partials/steps/payment.tpl`

src/js/theme.ts
```diff
@@ -38,14 +38,16 @@ import initResponsiveToggler from './responsive-toggler';
...
+import initCheckout from './pages/checkout';
 import useToast from './components/useToast';
 import useProgressRing from './components/useProgressRing';
 import './modules/blockcart';
 import initProductBehavior from './product';
 import './mobile-menu';
 import './modules/ps_searchbar';
 import './modules/facetedsearch';
+import './modules/ps_languageselector';
+import './modules/ps_currencyselector';
```
kpodemski, line 50: @NeOMakinG, is there any chance to force webpack to export these files as separated JS files to proper directories? I'm wondering what if one decides to disable these modules. Will we end up having a bunch of not needed JS?

mparvazi: @kpodemski maybe someone wants to disable facetedsearch, mobile-menu, ...
I think having just 10 lines of JS code in bundle doesn't impact the performance.

NeOMakinG: This is something we need to think about, It should be possible

matthieu-rolland (approved):

templates/checkout/_partials/modal-terms.tpl
```diff
@@ -0,0 +1,39 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ *}
+
+ <div id="checkout-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="checkout-modal" aria-hidden="true">
```
Progi1984, line 26:
```suggestion
<div id="checkout-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="checkout-modal" aria-hidden="true">
```

Progi1984 (commented):
A small comment

src/js/theme.ts
Hlavtox: Yes, we need to revert to this. In 1.6, due to separation of css files, only the global.css + bundles related to the controller were loaded. In 1.7, at least modules stayed “modular”.

We cannot make it even worse and make everything into one huge bundle loading everything even if it's not needed. 😬

CCC system is ready for this, it caches unique packs of assets.

mparvazi:
At "final summery":
1. Terms of service is **disabled** in BO
2. No payment method selected
So we will have this alert (but it's wrong):
develop

The alert message should be separated  in context:

develop2

It seems must be fixed here first:
`PS/themes/_core/js/checkout-payment.js`

Hlavtox:
Good find!

@mparvazi What do you think about making the **Place order** button full width, green and with paperplane icon for example?

mparvazi:
@Hlavtox 

We don't need to make it full-width, would be scary in xxlarge screen!
Just floating to the right is enough (in BS standard float-end).

IMO the place order button should has "cart-check" icon:
For example in BS icons:
1. `cart` icon for "Shopping cart"
2. `cart-plus` icon for "Add to cart"
3. `cart-check` icon for "Place order"
Screenshot 2022-04-06 at 14-09-10 Bootstrap Icons

In material-icons we have:
Screenshot 2022-04-06 at 14-14-27 Material Design Icons DX

Btw need to update to latest version for using them ([material-design-icons-iconfont](https://github.com/jossef/material-design-icons-iconfont) v6.5)

mparvazi:
If need to have an icon, doesn't need to has right arrow icon (because it's the last step), 
so my suggestion:
Screenshot 2022-04-06 at 14-26-13 develop

it seems material-icons doesn't have "paper-plane" icon.

mparvazi:
@NeOMakinG 
`
templates/checkout/_partials/order-confirmation-table.tpl`
Ln:116 & Ln:120
`</td>` => `</div>`

templates/checkout/_partials/order-confirmation-table.tpl
```diff
@@ -89,8 +88,7 @@
       </div>
       
       <div class="item__prices col-md-2 col-sm-12 col-12">
-        <div class="text-md-end">{$product.price}</div>
-        <div class="text-md-end">{$product.quantity}</div>
+        <div class="text-md-end">{$product.price} {l s='x%s' sprintf=[$product.quantity] d='Shop.Theme.Catalog'}</div>
```
mparvazi, line 91:
```suggestion
        <div class="text-md-end">({l s='x%s' sprintf=[$product.quantity] d='Shop.Theme.Catalog'}) {$product.price}</div>
```

mparvazi: Screenshot 2022-04-06 at 14-47-55 develop

src/js/modules/ps_currencyselector.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
+
+const {prestashop} = window;
+
+const initCurrencySelector = () => {
+  const languageSelector = document.querySelector<HTMLElement>('.js-currency-selector');
```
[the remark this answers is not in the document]
NeOMakinG: We do, it's a wip

src/js/modules/ps_languageselector.ts
```diff
@@ -0,0 +1,46 @@
...
+    const option = event.target as HTMLOptionElement;
+
+    window.location.href = option.value;
+  });
+};
+
+document.addEventListener('DOMContentLoaded', () => {
+  initLanguageSelector();
+
+  prestashop.on('responsive update', () => {
```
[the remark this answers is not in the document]
NeOMakinG: Seems like you're right, we should rename it

src/js/theme.ts
NeOMakinG: @jolelievre you got it, we changed the module markup, so we need to handle things with our own markup now

src/js/pages/checkout.ts
```diff
@@ -120,6 +127,42 @@ const initCheckout = () => {
...
+
+              termsModal.show();
+            }
+          } catch (e) {
+            prestashop.emit('handleError', {eventType: 'clickOnTermsLink', error: e});
+          }
+        })();
+      }
+
+      $(prestashop.themeSelectors.modal).modal('show');
```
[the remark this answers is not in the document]
NeOMakinG: Was not needed at all, it was probably a test

NeOMakinG:
> At "final summery":
> 
> 1. Terms of service is **disabled** in BO
> 2. No payment method selected
>    So we will have this alert (but it's wrong):
>    develop
> 
> The alert message should be separated in context:
> 
> develop2
> 
> It seems must be fixed here first: `PS/themes/_core/js/checkout-payment.js`

Do we have a world where `$conditions_to_approve` contains something else than terms? Overwise we can handle it directly in the template

@Hlavtox @mparvazi

mparvazi:
@NeOMakinG 

> Do we have a world where `$conditions_to_approve` contains something else than terms? Overwise we can handle it directly in the template
> 
The `checkout-payment.js` controls the alert container for show/hide, so need to have conditional content for alert:

Screenshot from 2022-04-08 12-40-49

NeOMakinG:
> @NeOMakinG
> 
> > Do we have a world where `$conditions_to_approve` contains something else than terms? Overwise we can handle it directly in the template
> 
> The `checkout-payment.js` controls the alert container for show/hide, so need to have conditional content for alert:
> 
> Screenshot from 2022-04-08 12-40-49

Yeah but as it's an array, I've the feeling that the terms message can be shown even if terms are disabled, with such a condition

Hlavtox:
There is a hook for product terms and conditions. https://github.com/PrestaShop/PrestaShop/issues/26608 Can you try the hook and test the behavior?

mparvazi:
When "Terms of service" in BO is disabled, `$conditions_to_approve|count` returns 0
Btw I've tested this code in several situation and it works.

NeOMakinG:
No one of you responded :p

Is there something else than terms and services in $condition_to_approve? I guess yes as it's an array, then $conditions_to_approve isn't directly pegged to Terms, but can contain another conditions, so we can't use it for this

Hlavtox:
@NeOMakinG `conditions_to_approve` can contain another conditions from hooked modules, so there can be multiple conditions :D

See here - https://github.com/PrestaShop/PrestaShop/blob/develop/classes/checkout/CheckoutPaymentStep.php
And the finder here - https://github.com/PrestaShop/PrestaShop/blob/develop/classes/checkout/ConditionsToApproveFinder.php

NeOMakinG:
Then we need to find another way, like the hook you talked about, I'll search it

Hlavtox:
And if you can make it so it can be optional, it would be good, I would make the core change required. :-)))

So the modules can hook also optional conditions and check their state by JS.

NeOMakinG:
I'm ok with it if we make sure that the theme will be compatible with the latest 1.7 version :p

mparvazi:
> So I think even if the default ones are disabled, there can be conditions from a module.

Is this a problem?
I mean if "Terms of service" in BO is disabled, then other conditions must not be displayed?

NeOMakinG:
> > So I think even if the default ones are disabled, there can be conditions from a module.
> 
> Is this a problem? I mean if "Terms of service" in BO is disabled, then other conditions must not be displayed?

If Terms of service is disabled, it will remove Terms of service from the array, but not others conditions to approve, which aren't forced to be pegged to terms of services, it can be any conditions a module adds

mparvazi:
@NeOMakinG 

To fix undefiend `selectorsMap` in componenets:

Can remove this line: `window.prestashop.themeSelectors = selectorsMap;`
from `theme.ts` and add at the end of `selectors-map.ts`

templates/checkout/_partials/modal-terms.tpl
```diff
@@ -0,0 +1,39 @@
...
+       <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="{l s='Close' d='Shop.Theme.Global'}"></button>
+     </div>
+     <div class="modal-body text-center">
+      <div class="spinner-border text-primary" role="status">
+        <span class="visually-hidden">{l s='Loading...' d='Shop.Theme.Global'}</span>
+      </div> 
+     </div>
+   </div>
+ </div>
+</div>
```
Progi1984, line 39:
```suggestion
<div id="checkout-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="checkout-modal" aria-hidden="true">
  <div class="modal-dialog modal-lg" role="document">
    <div class="modal-content">
      <div class="modal-header">
        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="{l s='Close' d='Shop.Theme.Global'}"></button>
      </div>
      <div class="modal-body text-center">
        <div class="spinner-border text-primary" role="status">
          <span class="visually-hidden">{l s='Loading...' d='Shop.Theme.Global'}</span>
        </div> 
      </div>
    </div>
  </div>
</div>
```

Progi1984 (commented):
@NeOMakinG Just fixing your indentation in a template.

src/js/modules/blockcart.ts
```diff
@@ -43,23 +46,30 @@ prestashop.blockcart.showModal = (html: string) => {
 
   const mainElement = document.createElement('div');
   mainElement.innerHTML = html;
+  // Select the template of the modal
+  const modalTemplate = document.querySelector<HTMLElement>(BlockcartMap.modal);
```
mparvazi, line 50:
```suggestion
  const modalTemplate = mainElement.querySelector<HTMLElement>(BlockcartMap.modal);
```

src/js/theme.ts
```diff
@@ -37,16 +34,18 @@ import 'bootstrap-input-spinner/src/bootstrap-input-spinner';
 import initResponsiveToggler from './responsive-toggler';
 import initQuickview from './quickview';
 import initCart from './pages/cart';
-import initCheckout from './checkout';
+import initCheckout from './pages/checkout';
 import useToast from './components/useToast';
 import useAlert from './components/useToast';
```
mparvazi, line 39:
```suggestion
import useAlert from './components/useAlert';
```

NeOMakinG: Will be fixed inside another PR

templates/checkout/_partials/steps/addresses.tpl
```diff
@@ -113,15 +113,15 @@
...
 
       <div class="mt-4">
-        <button class="btn btn-outline-primary btn-with-icon d-none d-md-inline-block me-2 js-back" data-step="checkout-personal-information-step">
+        <button class="btn btn-outline-primary btn-with-icon d-block d-md-inline-block me-2 w-full w-md-auto mb-3 mb-md-0 js-back" data-step="checkout-personal-information-step">
           <div class="material-icons">arrow_backward</div>
           {l s='Continue to Personal Information' d='Shop.Theme.Actions'}
         </button>
 
         {if !$form_has_continue_button}
             <button type="submit" class="btn btn-primary btn-with-icon d-block d-md-inline-block w-full w-md-auto continue" name="confirm-addresses" value="1">
```
mparvazi, line 122:
```suggestion
            <button type="submit" class="btn btn-primary btn-with-icon d-block d-md-inline-block w-full w-md-auto float-end continue" name="confirm-addresses" value="1">
```

templates/checkout/_partials/steps/payment.tpl
```diff
@@ -114,8 +114,23 @@
...
+          '[/2]' => '</a>'
+        ]
+        d='Shop.Theme.Checkout'
+      }
+    </article>
+  {/if}
+
   <div class="payment__actions d-block d-md-flex">
-    <button class="btn btn-outline-primary btn-with-icon d-none d-md-inline-block me-2 js-back" data-step="checkout-delivery-step">
+    <button class="btn btn-outline-primary btn-with-icon d-block d-md-inline-block me-2 w-full w-md-auto mb-3 mb-md-0 js-back" data-step="checkout-delivery-step">
```
mparvazi, line 133:
```suggestion
    <button class="btn btn-outline-primary btn-with-icon d-block d-md-inline-block me-auto w-full w-md-auto mb-3 mb-md-0 js-back" data-step="checkout-delivery-step">
```

templates/checkout/_partials/steps/shipping.tpl
```diff
@@ -100,30 +100,30 @@
...
         </div>
 
         <div class="shipping__actions">
-          <button class="btn btn-outline-primary btn-with-icon d-none d-md-inline-block me-2 js-back" data-step="checkout-addresses-step">
+          <button class="btn btn-outline-primary btn-with-icon d-block d-md-inline-block me-2 w-full w-md-auto mb-3 mb-md-0 js-back" data-step="checkout-addresses-step">
             <div class="material-icons">arrow_backward</div>
             {l s='Back to Addresses' d='Shop.Theme.Actions'}
           </button>
 
           <button type="submit" class="btn btn-primary btn-with-icon d-block d-md-inline-block w-full w-md-auto" name="confirmDeliveryOption" value="1">
```
mparvazi, line 124:
```suggestion
          <button type="submit" class="btn btn-primary btn-with-icon d-block d-md-inline-block w-full w-md-auto float-end" name="confirmDeliveryOption" value="1">
```

mparvazi (commented):
Just a suggestion:
Place the "next buttons" on the right.

Screenshot 2022-04-15 at 23-35-17 develop

templates/checkout/_partials/order-final-summary.tpl
```diff
@@ -22,70 +22,125 @@
...
-          <h4 class="h5 black addresshead">{l s='Your Delivery Address' d='Shop.Theme.Checkout'}</h4>
-          {$customer.addresses[$cart.id_address_delivery]['formatted'] nofilter}
+<hr />
+<section id="order-summary-content" class="order__summary">
+  <h4 class="mb-4">{l s='Please check your order before payment' d='Shop.Theme.Checkout'}</h4>
+
+  <div class="order__summary__addresses mb-4">
+    <h5 class="mb-3">
+      {l s='Addresses' d='Shop.Theme.Checkout'}
+      <span class="step-edit step-to-addresses fs-6 text-gray js-edit-addresses" data-step="checkout-addresses-step"><i class="material-icons edit">mode_edit</i> {l s='edit' d='Shop.Theme.Actions'}</span>
```
mparvazi, line 33:
```suggestion
      <span class="btn step-edit step-to-addresses fs-6 text-gray js-edit-addresses" data-step="checkout-addresses-step"><i class="material-icons fs-6 edit">mode_edit</i> {l s='Edit' d='Shop.Theme.Actions'}</span>
```

templates/checkout/_partials/order-final-summary.tpl
```diff
@@ -22,70 +22,125 @@
...
     </div>
   </div>
 
   {if !$cart.is_virtual}
-    <h4 class="h4">
+    <h5 class="mb-3">
       {l s='Shipping Method' d='Shop.Theme.Checkout'}
-      <span class="step-edit step-to-delivery js-edit-delivery"><i class="material-icons edit">mode_edit</i> {l s='edit' d='Shop.Theme.Actions'}</span>
-    </h4>
+      <span class="step-edit step-to-addresses fs-6 text-gray js-edit-shipping" data-step="checkout-delivery-step"><i class="material-icons edit">mode_edit</i> {l s='edit' d='Shop.Theme.Actions'}</span>
```
mparvazi, line 98:
```suggestion
      <span class="btn step-edit step-to-addresses fs-6 text-gray js-edit-shipping" data-step="checkout-delivery-step"><i class="material-icons fs-6 edit">mode_edit</i> {l s='Edit' d='Shop.Theme.Actions'}</span>
```

mparvazi (commented):
Fix "Edit button" for sections on "order final summary" page.

Screenshot 2022-04-15 at 23-51-49 develop

templates/checkout/_partials/order-final-summary.tpl
```diff
@@ -22,70 +22,125 @@
...
+      <div class="col-12 col-sm-6 mb-2">
+        <div class="address card">
+          <div class="card-body">
+            <div class="address__content col-10">
+              <h6 class="address__alias card-title">{l s='Your Delivery Address' d='Shop.Theme.Checkout'}</h6>
+              <address class="address__content">{$customer.addresses[$cart.id_address_delivery]['formatted'] nofilter}</address>
+
+              <div class="address__actions">
+                <a
+                  class="address__edit text-muted ps-0 pb-0"
```
mparvazi, line 46:
```suggestion
                  class="address__edit ps-0 pb-0"
```

templates/checkout/_partials/order-final-summary.tpl
```diff
@@ -22,70 +22,125 @@
...
+      <div class="col-12 col-sm-6 mb-2">
+        <div class="address card">
+          <div class="card-body">
+            <div class="address__content col-10">
+              <h6 class="address__alias card-title">{l s='Your Invoice Address' d='Shop.Theme.Checkout'}</h6>
+              <address class="address__content">{$customer.addresses[$cart.id_address_delivery]['formatted'] nofilter}</address>
+
+              <div class="address__actions">
+                <a
+                  class="address__edit text-muted ps-0 pb-0"
```
mparvazi, line 74:
```suggestion
                  class="address__edit ps-0 pb-0"
```

mparvazi (commented):
We should not use "text-muted" for buttons or links. The hover state will be off for link (text-muted uses !important for color in BS).
```
.text-muted {
  color: #6c757d !important;
}
```

templates/checkout/_partials/steps/addresses.tpl
Hlavtox: I would use flex, justify content space between. The floats sometime cause weird parent div height behavior. 👍

NeOMakinG:
Fixed last feedback, can we finally merge this? So we can provide every fixes to the theme :p

NeOMakinG:
As noone from @mparvazi and @Hlavtox disapproved, I'll merge this one and iterate on a new PR in order to fix last fixes from the big issue list

PR #228, opened 2022-04-19
Change ps_mainmenu to make it work dropdowns

rodriciru:
| Questions         | Answers
| Description?      | Change dropdowns to make it work in menu
| Type?             | bug fix 
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     | 
| How to test?      | Just update and check submenus whit more submenus
| Possible impacts? | Nope

Hi.
When a menu child has submenus, these submenus don't have margin, so looks like they don't have any relation.
And you can't hide/show this sub menus.
This PR fix that
Old behavior
imagen

New  behavior
imagen

Collapsed submenus
imagen

You can go to the parent URL just by clicking the name, and to collapse the submenu, click the right arrows
Will be great to change the arrows according if it's open or not

### DIFF
--- a/modules/ps_mainmenu/ps_mainmenu.tpl
+++ b/modules/ps_mainmenu/ps_mainmenu.tpl
@@ -1,86 +1,88 @@
 {assign var=_counter value=0}
 {function name="menu" nodes=[] depth=0 parent=null}
-    {if $nodes|count}
-      <ul class="{if $depth == 0}main-menu__tree{/if}" {if $depth == 0}id="top-menu"{/if} data-depth="{$depth}">
-        {foreach from=$nodes item=node}
-            <li class="{$node.type}{if $node.current} current {/if}{if $depth == 0} main-menu__tree__item{/if}{if $node.children|count} dropdown{/if}" id="{$node.page_identifier}">
-            {assign var=_counter value=$_counter+1}
-              <a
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
I will test it later, accepting it but I guess the menu needs more work to finish it

mparvazi:
Screenshot from 2022-04-20 12-29-34

Screenshot from 2022-04-20 12-26-53

Screenshot from 2022-04-20 12-26-35

Screenshot from 2022-04-20 12-26-16

mparvazi:
I think we should not use collapse buttons, all categories should be visible.
We can use flex for rendering the menu items:

Screenshot from 2022-04-20 13-14-32

NeOMakinG:
I guess we should follow these mockups:
https://www.figma.com/file/LfVl5leeSKcVUhSaYwhbtM/New-Theme?node-id=4%3A461

rodriciru:
> I guess we should follow these mockups: https://www.figma.com/file/LfVl5leeSKcVUhSaYwhbtM/New-Theme?node-id=4%3A461

Agreed, but in the meantime this fixes this problem

NeOMakinG:
Let's just merge this, as anyway we will work on it later

NeOMakinG:
Thanks @rodriciru

rodriciru:
> Thanks @rodriciru

Nope. Thank you all for the hard work

PR #232, opened 2022-04-23
Fix dropdown menu for RTL languages

mparvazi:
Using static display for dropdown menu (popper.js will be disabled so bugs for RTL will be fixed)
The goal of this change is that not to use `!important` for fixing bugs in RTL (popper.js uses inline style for dropdown positioning).

Screenshot from 2022-04-23 14-06-39

Screenshot from 2022-04-23 14-08-20

Screenshot from 2022-04-23 14-16-18

## RTL fixed

Screenshot from 2022-04-23 14-08-36

### DIFF
--- a/modules/ps_brandlist/views/templates/_partials/brand_form.tpl
+++ b/modules/ps_brandlist/views/templates/_partials/brand_form.tpl
@@ -27,6 +27,7 @@
   <button
     class="btn-unstyle select-title"
     rel="nofollow"
+    data-bs-display="static"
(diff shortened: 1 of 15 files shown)


mparvazi:
https://user-images.githubusercontent.com/85633460/164891490-e59da2f9-65eb-4942-8b75-9ae6670505f4.mp4


https://user-images.githubusercontent.com/85633460/164891502-fae4ad5e-ced9-4a92-b475-abf913300406.mp4

PR #243, opened 2022-04-28
Fix typographic base for BS

mparvazi:
Some components use `$line-height-base` for alignment, so it must be defined.
Using `line-height` in body definition leads to some bugs in components (e.g. label):

| Current | Fixed |
| :---:        |     :---:      |
|    |   |

### DIFF
--- a/src/scss/partials/_commons.scss
+++ b/src/scss/partials/_commons.scss
@@ -10,7 +10,6 @@ html {
 
 body {
   font-family: $font-family-base;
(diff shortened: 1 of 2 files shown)


NeOMakinG:
I'll need to test is personally, removing the line-height on the body can cause some side effect, just to be sure

NeOMakinG (approved):
Couldn't find any regression, looks fine to me 👍

PR #245, opened 2022-04-28
Fix image slider control buttons

mparvazi:


NeOMakinG (approved):
Nitpick, I'll accept it but I think that the slider will probably move to another lib in the future

Thanks 🚀

PR #269, opened 2022-05-06
Ps_banner module image display fix

### LINKED ISSUE #268: Usage of ps_banner inside theme
We shouldn't set fixed with/height and force image ratio inside template.
Merchant should be able to use any image.

Oksydan:
| Questions         | Answers
| Description?      | Issue describes it enough. We should able to use `banner_height` and `banner_width` when new version of ps_banner will be released (https://github.com/PrestaShop/ps_banner/pull/37)
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | Fixes #268 
| How to test?      | Set new banner and check FO.
| Possible impacts? | Nope


kpodemski (approved):

NeOMakinG (approved):
Great 👍

NeOMakinG:
@Oksydan require a small rebase due to your first PR!

Oksydan:
@NeOMakinG rebased

NeOMakinG:
@Oksydan something went wrong with the rebase

Oksydan:
@NeOMakinG yea, you are right I didn't notice that banner import is back 🙈

mparvazi:
So as I saw in this PR, for styling it's better to use BS class utilities (as much as possible) rather than writing everything with Sass. right?

Oksydan:
@mparvazi IMO yes, you can do a lot just with utilities and BS components. 
Bootstrap contains a lot of utility code with breakpoints and like 90% of them are never used.

SharakPL:
> Bootstrap is not designed to be an utility lib like Tailwind can be

But it is and it's always been! It has less utility classes than Tailwind but it's still best part of Bootstrap. I never understood why Classic duplicates a lot instead of just using BS classes. It's the same with components. Custom collapse, custom modals, custom navbar. Why? All of it is already there. Most of the work should be setting up bootstrap variables, not making new components.
It would be so much easier to just change the class in the template instead of setting up dev env and compiling everything again even for a small change.

NeOMakinG:
Thanks @Oksydan 

The module is currently being released, overwise let's just work with the dev version!

PR #272, opened 2022-05-08
Fix a11y for Desktop Menu

mparvazi:
### Keyboard navigation:
- **Main menu links [depth=0]**
 1- Change focus by TAB
 2- Open menu with ARROW_DOWN (focus on first link [depth=1]) and ARROW_UP (focus on last link [depth=1])
- **Inside opened menu**
 1- Change focus by TAB
 2- Close menu with ESC
 3- Open dropdown with ARROW_RIGHT and ARROW_LEFT

https://user-images.githubusercontent.com/85633460/167289613-2f07b536-4ad6-41c5-8f80-44ae0603d652.mp4

### DIFF
--- a/modules/ps_mainmenu/ps_mainmenu.tpl
+++ b/modules/ps_mainmenu/ps_mainmenu.tpl
@@ -27,7 +27,7 @@
           </a>
           {if $node.children|count}
             {if $depth !== 1}
-              <div class="{if $depth === 0}menu-container js-sub-menu{/if}{if $depth > 1 && $node.children|count}dropdown-menu{/if}">
+              <div class="{if $depth === 0}menu-container shadow-sm js-sub-menu{/if}{if $depth > 1 && $node.children|count}dropdown-menu{/if}">
             {/if}
               {if $depth === 0}
                 <div class="container">
--- a/modules/ps_mainmenu/ps_mainmenu.tpl
+++ b/modules/ps_mainmenu/ps_mainmenu.tpl
@@ -2,12 +2,12 @@
   {if $nodes|count}
     <ul
       {if $depth === 0}id="top-menu"{/if}
-      class="{if $depth === 0}main-menu__tree{else if $depth === 1}row gy-3{else}submenu{/if}"
+      class="{if $depth === 0}main-menu__tree{else if $depth === 1}row row-cols-4 gy-3{else}submenu{/if}"
       data-depth="{$depth}"
     >
     {foreach from=$nodes item=node}
       <li
-        class="{$node.type}{if $node.current} current{/if}{if $depth === 0} main-menu__tree__item{/if}{if $depth === 1} col-3{/if}"
+        class="{$node.type}{if $node.current} current{/if}{if $depth === 0} main-menu__tree__item{/if}"
         id="{$node.page_identifier}"
       >
         {if $depth > 1 && $node.children|count}
--- a/modules/ps_mainmenu/ps_mainmenu.tpl
+++ b/modules/ps_mainmenu/ps_mainmenu.tpl
@@ -1,32 +1,50 @@
-{function name="menu" nodes=[] depth=0 parent=null}
+{function name="desktopMenu" nodes=[] depth=0 parent=null}
   {if $nodes|count}
-    <ul class="{if $depth == 0}main-menu__tree{/if}{if $depth == 1} row g-3{/if}{if $depth > 1} submenu{/if}"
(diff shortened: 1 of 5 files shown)


mparvazi:
@NeOMakinG 

Give me a screenshot for focus elements on Safari.
Need to compare with Firefox and Chrome.
Should define outline style for hummingbird to make it similar on all browsers. 

Hopefully Firefox in new version made outline style as like as Chrome.

modules/ps_mainmenu/ps_mainmenu.tpl
```diff
@@ -1,104 +1,140 @@
...
   </button>
 </div>
 
-<div class="main-menu__offcanvas offcanvas offcanvas-start js-menu-canvas" tabindex="-1" id="mobileMenu"
-  aria-labelledby="mobileMenuLabel">
+<div
+  class="main-menu__offcanvas offcanvas offcanvas-start js-menu-canvas" 
+  tabindex="-1" 
+  id="mobileMenu"
+  aria-labelledby="mobileMenuLabel"
```
NeOMakinG, line 126: I guess no content is linked to that label, we should either use aria-label or use a good element to describe this canvas, wdyt?

src/js/modules/ps_mainmenu.ts
```diff
@@ -1,45 +1,91 @@
...
+  /**
+   * Handle Keyboard events for main menu links
+   * Find all links with [data-depth: 0]
+   */
+  const mainMenuLinks = document.querySelectorAll(desktopMenuMap.dropdownItemAnchor(0));
+  mainMenuLinks.forEach((mainMenuLinkElement: HTMLElement) => {
+    const menuContainerElement = mainMenuLinkElement.nextElementSibling;
+
+    /* Open the menu container when Arrow Up or Arrow Down key is pressed */
+    mainMenuLinkElement.addEventListener('keydown', (event: KeyboardEvent) => {
```
NeOMakinG, line 62: Do we really need this? Bootstrap dropdowns should naturally handle keyboard navigation

mparvazi: The current structure is not based on BS standard for dropdowns,
If we add 'dropdown' selector to main links parent container to enable BS keyboard navigation,
then we will have problem for controlling the menu container, so need to do a refacto and change current structure to BS navbar.

modules/ps_mainmenu/ps_mainmenu.tpl
mparvazi: Yep!
It's just a container, I think something like this is enough:
`<div id="mobileMenu" class="main-menu__offcanvas offcanvas offcanvas-start js-menu-canvas">`

NeOMakinG: Probably, as the id will be taken in account when screen readers talks?

src/js/modules/ps_mainmenu.ts
NeOMakinG: Do you see some constraints on moving to the BS structure? If we can use the exact same style but with a cleaner bootstrap structure, that could be a great maintainable solution

I would probably accept your PR as a first workaround, but if we can do something more cleaner, sticking to BS, maybe this is the right choice

wdyt @Hlavtox ?

NeOMakinG (approved):
Not blocking, it's a great working thing with the current menu implementation, well-done.

But I'm wondering if we didn't do a mistake by not using the bootstrap navbar markup directly, and trying to have the same mockup styling, maybe this would be a great improvement in the future

src/js/modules/ps_mainmenu.ts
mparvazi: The current structure is fine,
If you try to use BS Navbar with dropdowns in main links, then you can not have menu-container with rows and columns.

In this PR I tried to simulate BS dropdown behavior for dropdowns in main-links.

modules/ps_mainmenu/ps_mainmenu.tpl
mparvazi: `<div class="main-menu__offcanvas offcanvas offcanvas-start js-menu-canvas" id="mobileMenu">`

mparvazi: I don't have any idea about this one!

NeOMakinG: I guess yes

mparvazi:
@Hlavtox 
Wdyt?
Any changes needed?

PR #276, opened 2022-05-10
Improve Customization form

mparvazi:

### DIFF
--- a/templates/catalog/_partials/product-customization.tpl
+++ b/templates/catalog/_partials/product-customization.tpl
@@ -2,40 +2,79 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  *}
+ 
 <section class="product-customization js-product-customization">
   {if !$configuration.is_catalog}
-    <div class="card card-body bg-light">
-      <p class="h4 card-title">{l s='Product customization' d='Shop.Theme.Catalog'}</p>
-
-      {block name='product_customization_form'}
-        <form method="post" action="{$product.url}" enctype="multipart/form-data">
-          {foreach from=$customizations.fields item="field"}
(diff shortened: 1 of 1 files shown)


Hlavtox (changes requested):
One small thing - we are doubling the paddings on the sides due to extra card. On mobile, we will have 3x the padding on both sides, WDYT?

Otherwise, I would unify these parts of code (duplicated for image and text) and it's a go for me!

```
<div class="card border-1 mb-3">
<div class="card-header bg-transparent fw-bold border-1">{$field.label}</div>
<div class="card-body">
```

```
<div class="card-footer bg-transparent border-1">
<div class="d-flex">
<div class="flex-grow-1">
<small class="text-muted">{l s='250 char. max' d='Shop.Forms.Help'}</small> // CONDITION HERE
</div>
{if !$field.required}
<small class="text-muted">{l s='Optional' d='Shop.Forms.Help'}</small>
{/if}
</div>
</div>
```

NeOMakinG (approved):
Let's approve this, and see if peoples complains about it, I think it's usable like it is, but I may be wrong

mparvazi:
> Pretty worried about smallest phones which are using `320px` layouts, specially when you added an image
> 
> But I guess it should stay usable anyway

Screenshot from 2022-05-10 12-15-43

mparvazi:
Btw for small display phones we can use something like this for Card padding:
`<div class="card-body px-2 px-sm-3">`

Screenshot from 2022-05-10 13-21-50

mparvazi:
Optimize cards for very small screen:

Screenshot from 2022-05-10 14-45-15

mparvazi:
Safari:

NeOMakinG (approved):
Looks like we can give it a go 👍

PR #279, opened 2022-05-11
Move window.prestashop.themeSelectors selectors to window.Theme.selectors

### LINKED ISSUE #274: Move every selectors to the theme side instead of the `prestashop` object
Every selectors from the theme should be exported by webpack to `window.Theme.selectors` instead of `window.prestashop.themeSelectors`

We decided to split the theme components and whatever from the core, in order to make it clearer, everything exported from `theme.ts` is naturally usable inside `window.Theme`, we can export a `selectors` object from `theme.ts` and use it everywhere in our codebase instead of `window.prestashop`

MeKeyCool:
| Questions         | Answers
| Description?      | Split Theme and core elements under window variable.
| Type?             | refacto
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | Fixes #274.
| How to test?      | All the theme should work.
| Possible impacts? | Every selector may be impacted so all the theme could be broken.

### DIFF
--- a/src/js/pages/checkout.ts
+++ b/src/js/pages/checkout.ts
@@ -7,8 +7,7 @@ import useProgressRing from '@js/components/useProgressRing';
 
 const initCheckout = () => {
   const {prestashop} = window;
-  const {Theme} = window;
-  const {selectors} = Theme;
+  const {Theme: {selectors}} = window;
   const {progressRing: ProgressRingMap, checkout: CheckoutMap} = selectors;
 
   const steps = document.querySelectorAll<HTMLElement>(CheckoutMap.steps.item);
--- a/src/js/pages/checkout.ts
+++ b/src/js/pages/checkout.ts
@@ -7,8 +7,9 @@ import useProgressRing from '@js/components/useProgressRing';
 
 const initCheckout = () => {
   const {prestashop} = window;
-  const {themeSelectors: selectorsMap} = prestashop;
-  const {progressRing: ProgressRingMap, checkout: CheckoutMap} = selectorsMap;
+  const {Theme} = window;
+  const {selectors} = Theme;
+  const {progressRing: ProgressRingMap, checkout: CheckoutMap} = selectors;
 
   const steps = document.querySelectorAll<HTMLElement>(CheckoutMap.steps.item);
   const actionButtons = document.querySelectorAll<HTMLElement>(CheckoutMap.actionsButtons);
@@ -120,8 +121,8 @@ const initCheckout = () => {
             const content = await response.text();
             const contentElement = document.createElement('div');
             contentElement.innerHTML = content;
-            const modalBody = termsModalElement.querySelector(selectorsMap.modalBody);
-            const sanitizedContent = contentElement.querySelector(selectorsMap.pageCms);
+            const modalBody = termsModalElement.querySelector(selectors.modalBody);
+            const sanitizedContent = contentElement.querySelector(selectors.pageCms);
 
             if (sanitizedContent && modalBody) {
               modalBody.innerHTML = sanitizedContent.innerHTML;
--- a/src/js/theme.ts
+++ b/src/js/theme.ts
@@ -2,7 +2,7 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-import selectors from './constants/selectors-map';
+import themeSelectors from './constants/selectors-map';
 import initEmitter from './prestashop';
 import initResponsiveToggler from './responsive-toggler';
 import initQuickview from './quickview';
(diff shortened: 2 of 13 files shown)


types/global.d.ts
```diff
@@ -1,5 +1,6 @@
 interface Window { 
   prestashop: any;
+  Theme: any;
```
NeOMakinG, line 3: As this is an intern dependency, can't we import the `selectors` object and use something like `Theme: typeof selectors`?

NeOMakinG:
@MeKeyCool Looks fine to me, but I noticed something.

There are some components importing these selectors directly instead of using the window object (useQuantityInput for example).

This is wrong because we want child themes and modules to be able to override these selectors without any bundlers...

Would mind improving this at the same time? So we can get rid of these selectors issues

types/global.d.ts
NeOMakinG: We should definitely type it somehow, it will have a lot of benefits

src/js/theme.ts
```diff
@@ -24,7 +24,7 @@ import initDropdownToggleMenu from './modules/ps_mainmenu';
 
 initEmitter();
 
-window.prestashop.themeSelectors = selectorsMap;
+window.Theme.selectors = selectorsMap;
```
NeOMakinG, line 27:
```suggestion
```

src/js/pages/checkout.ts
```diff
@@ -7,7 +7,8 @@ import useProgressRing from '@js/components/useProgressRing';
 
 const initCheckout = () => {
   const {prestashop} = window;
-  const {themeSelectors: selectorsMap} = prestashop;
+  const {Theme} = window;
+  const {selectors: selectorsMap} = Theme;
```
NeOMakinG, line 11:
```suggestion
  const {selectors} = Theme;
```

src/js/theme.ts
NeOMakinG: and add `selectors` to the `export default`, it will be magically exported to `window.Theme`

NeOMakinG (changes requested):
2 small requests:

- Type the selectors object on window
- Remove the assignment to window, export directly using webpack (see comment)

types/global.d.ts
mparvazi: `src/js/constants/selectors-map.ts`
```javascript
export interface Selector {
  [key: string]: string | object;
}

export const facetedsearch: Selector = {
  range: '.js-faceted-slider',
  rangeContainer: '.js-faceted-slider-container',
  filterSlider: '.js-faceted-filter-slider',
};
```

`types/global.d.ts`
```javascript
import {Selector} from "@constants/selectors-map";

interface Window { 
  prestashop: any;
  Theme: Selector;
  $: JQueryStatic;
  jQuery: JQueryStatic;
}

```

NeOMakinG: @mparvazi don't use `object`, it's almost like using `any`

and I think this interface would be too generic, we could type it stronger somehow

NeOMakinG: Let's ignore that for now, I don't have the time to take a look rn, I'll create an issue when this gets merged

src/js/theme.ts
```diff
@@ -62,4 +60,5 @@ export default {
   initResponsiveToggler,
   initQuickview,
   initProductBehavior,
+  selectorsMap,
```
MeKeyCool, line 63: @NeOMakinG , was it that you wanted ? Shouldn't I rename `selectorsMap` to `selectors` ?
Is it thanks to this export default that we can `const {selectors} = Theme;` ?

mparvazi:
@NeOMakinG 

Screenshot 2022-05-12 134406

src/js/theme.ts
NeOMakinG: Yep, rename it to `selectors` (don't forget to rename the import

Yep, this will be naturally exported to `window.Theme` by webpack, check the webpack configuration, you will be able to find the magic

NeOMakinG:
> @NeOMakinG
> 
> Screenshot 2022-05-12 134406

This would allow only 2 levels right?

NeOMakinG (approved):
Looks fine to me, I'll need to test it before merging as it can be buggy somehow!

NeOMakinG (approved):
Tested it, found some bugs, fixed it, looks fine now

NeOMakinG:
@Hlavtox @mparvazi if you want to get a look and feel like it's mergeable, merge it or tell me to merge it! Thanks

mparvazi (approved):

Hlavtox:
Can't approve due to my low JS skills, I leave this one up to you. :-)

NeOMakinG:
Thanks @MeKeyCool

PR #288, opened 2022-05-20
Improve useQuantityInput and fix minor bug

mparvazi:
Bug:

If we have multiple product lines for one product, then corresponding alert container can not be found by useQuantityInput.
It's possible to have one product with customizations, so we need to define correct ID for each one.

Improvements:

1. Removed event handling from` theme.ts` and `quickview.ts`
2. Added the update listener only if quantity input has update URL.

## Bug:

https://user-images.githubusercontent.com/85633460/169544498-fc912c16-493a-4b29-a6f1-0c4c2b21fb96.mp4

## Fixed:

https://user-images.githubusercontent.com/85633460/169544521-c05aca1a-62d7-4c83-9f5f-86dbff4fccbc.mp4

### DIFF
--- a/src/js/components/useQuantityInput.test.ts
+++ b/src/js/components/useQuantityInput.test.ts
@@ -4,7 +4,7 @@
  */
 
 import initEmitter from '@js/prestashop';
-import selectorsMap from '@constants/selectors-map';
+import {quantityInput as quantityInputMap} from '@constants/selectors-map';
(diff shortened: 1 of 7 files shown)


NeOMakinG (approved):
Didn't test it yet but looks really fine to me! Great work

Hlavtox (approved):
Nice work dude

NeOMakinG:
Test and works like a charm!

PR #292, opened 2022-05-23
Add customization modal for product line

mparvazi:
It's confusing for users to see multiple lines for same product in Cart or Order details, without showing the difference between them (Customization data).

https://user-images.githubusercontent.com/85633460/169803231-e265cdf8-6cf7-4243-98fe-916eb2d418e8.mp4

### DIFF
--- a/templates/catalog/_partials/product-customization-modal.tpl
+++ b/templates/catalog/_partials/product-customization-modal.tpl
@@ -20,32 +20,30 @@
     <div class="modal-dialog modal-dialog-scrollable modal-fullscreen-md-down">
       <div class="modal-content">
         <div class="modal-header">
-          <h5 class="modal-title">{l s='Product customization' d='Shop.Theme.Checkout'}</h5>
-          <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+          <h4 class="modal-title">{l s='Product customization' d='Shop.Theme.Checkout'}</h4>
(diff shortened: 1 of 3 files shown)


templates/catalog/_partials/product-customization-modal.tpl
```diff
@@ -0,0 +1,53 @@
...
+                    </p>
+                  {elseif $field.type == 'image'}
+                    <img class="align-self-start rounded-3" src="{$field.image.small.url}">
+                  {/if}
+                </div>
+                {/foreach}
+              </div>
+            {/foreach}
+          </div>
+        </div>
```
Hlavtox, line 49: I think we can simplify the code a bit, WDYT

```suggestion
        <div class="modal-body">
          {foreach from=$product.customizations item="customization"}
            {foreach from=$customization.fields item="field"}
              <div class="mb-3">
                <p class="mb-1 text-dark">{$field.label}</p>
                {if $field.type == 'text'}
                  <p class="mb-0">
                    {if (int)$field.id_module}
                      {$field.text nofilter}
                    {else}
                      {$field.text}
                    {/if}
                  </p>
                {elseif $field.type == 'image'}
                  <img class="rounded-3" src="{$field.image.small.url}">
                {/if}
              </div>
            {/foreach}
          {/foreach}
        </div>
```

mparvazi: Yep! much better :+1:

Hlavtox (approved):

NeOMakinG (approved):
Looks fine

PR #293, opened 2022-05-24
TypeScript types improvements

NeOMakinG:
| Questions         | Answers
| Description?      | We had some types problems in the theme, most of the `window.Theme` object wasn't typed, it was hiding some types problems
| BC breaks?        |  no
| Deprecations?     | no
| How to test?      | JS behavior should be as usual, tests should be green

- Updated and changed jest to ts-jest
- Move every reusable components types to `.d.ts` files and made them globals
- I added every events to an `enum` so it's way more strict and will avoid mistakes in the future, as a typo will break the build
- I added Typed props to `window.Theme` so it's strict: events, selectors, components.


mparvazi (approved):
Awesome! neat and clean :+1:

PR #294, opened 2022-05-24
Improve quickview modal

mparvazi:
Modal is scrollable, so need to add border for header and footer.

The customization block should be displayed if product is customizable.
If there is a required field for customization, "Add to cart" button is disabled and without customization block it's confusing.

Screenshot from 2022-05-24 21-00-48


Hlavtox:
@NeOMakinG Merged this one as it's a tiny change

PR #295, opened 2022-05-24
Add quickview touch button for mobile

### LINKED ISSUE #291: Different display for quickview on touch screen
In screens that hover not works, need to show quickview button in product box.
```scss
// hide quickview touch button

@media (hover: none) and (pointer: coarse) {
  // hide hover quickview box
  // show quickview touch button
}
```

https://user-images.githubusercontent.com/85633460/169580931-bf914e64-b7d7-4445-a358-47909cde0766.mp4

mparvazi:
Fixes #291

### DIFF
--- a/src/scss/custom/components/product/_product-miniature.scss
+++ b/src/scss/custom/components/product/_product-miniature.scss
@@ -65,6 +65,10 @@ $component-name: product-miniature;
     padding: 0.5rem;
     background-color: $white;
     transition: 0.25s ease-out;
+
+    @media (hover: none) and (pointer: coarse) {
+      display: none;
+    }
   }
 
   & &__quickview_button {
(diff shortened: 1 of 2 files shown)


Hlavtox (approved):
Nice work, just a comment to think about - what would you say about making the button styling universal, like `product-miniature__action-button` - so we can have the same styling for wishlist and others?

mparvazi:
> Nice work, just a comment to think about - what would you say about making the button styling universal, like `product-miniature__action-button` - so we can have the same styling for wishlist and others?

Currently we don't have any access to wishlist button template in theme.

@NeOMakinG wdyt?

Hlavtox:
@NeOMakinG We should return that hook dude... it was removed for performance reasons, but isn't calling a php function 24 times still faster than loading a javascript for that?

NeOMakinG:
> @NeOMakinG We should return that hook dude... it was removed for performance reasons, but isn't calling a php function 24 times still faster than loading a javascript for that?

I guess no, overwise we would have kept it

Calling a dumb function doesn't cost so much, but calling `numberOfProduct * numberOfModulesUsingTheHook * some SQL request` and more can be really dangerous

But there is another way to be able to override that button for our styling guys, we should add some changes this to `blockwishlist`

NeOMakinG (approved):
Nice thing 👍

Issue #296, opened 2022-05-25
Add modules to config.yml


rodriciru:
I find some modules missing on certain or all hooks on config.yml.
- blockwishlist
- ps_featuredproducts
- ps_newproducts
- ps_bestsellers

Should we add it? They are so used, and in classic they appear on that file, but they are not core modules, you have to install it manually (blockwishlist is not the case)

The other option will be to add a ~ to keep that hook as is before the change.
https://devdocs.prestashop.com/1.7/themes/getting-started/theme-yml/

PR #297, opened 2022-05-25
Add modules to config.yml hook

rodriciru:
| Questions         | Answers
| Type?             | bug fix 
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/296
| How to test?      | Reinstall theme and check if module is hooked to that hook
| Possible impacts? | Noooope

I chose a `~` for displayFooter because I think many devs will use this hook to add some modals, scripts...., so we must keep as store admin wants, with all modules working

### DIFF
--- a/config/theme.yml
+++ b/config/theme.yml
@@ -67,9 +67,11 @@ global_settings:
         - ps_searchbar
       displayHome:
         - ps_imageslider
+        - ps_customtext
         - ps_featuredproducts
         - ps_banner
-        - ps_customtext
+        - ps_newproducts
+        - ps_bestsellers
       displayFooterBefore:
         - ps_emailsubscription
         - ps_socialfollow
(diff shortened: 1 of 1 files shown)


config/theme.yml
```diff
@@ -67,16 +67,19 @@ global_settings:
...
+        - ps_newproducts
+        - ps_bestsellers
       displayFooterBefore:
         - ps_emailsubscription
         - ps_socialfollow
       displayFooter:
         - ps_linklist
         - ps_customeraccountlinks
         - ps_contactinfo
+        - ~
```
NeOMakinG, line 82: Oh, what does this means?

rodriciru: Keep whatever is currently hooked to it.
In this very case, first ps_linklist, then ps_customeraccountlinks, then ps_contactinfo, then whatever is currently hooked to it

rodriciru: https://devdocs.prestashop.com/1.7/themes/getting-started/theme-yml/

NeOMakinG (approved):

PR #299, opened 2022-05-25
Fix some scss vars and tpls

rodriciru:
| Questions         | Answers
| Description?      | IF we customize card-bg color, quick view It's still white, so it looks like a Picasso. Also, i add some changes and new classes to other elements
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     |  no

By default, $card-bg it's white, so there are no visible changes.

Problem:
imagen

Solution (if we choose this wonderful red 😀):
imagen

Account identity old (same as old address):
imagen

Account identity new (bg and padding) (same as new address)
imagen

### DIFF
--- a/src/scss/abstract/variables/bootstrap/_accordion.scss
+++ b/src/scss/abstract/variables/bootstrap/_accordion.scss
@@ -1,7 +1,7 @@
 $accordion-padding-y:                     0;
 $accordion-padding-x:                     0;
 
-$accordion-border-width:                  0;
+$accordion-border-width:                  2px;
 
 $accordion-body-padding-y:                0.5rem;
 
--- a/src/scss/abstract/variables/bootstrap/_accordion.scss
+++ b/src/scss/abstract/variables/bootstrap/_accordion.scss
@@ -6,10 +6,10 @@ $accordion-border-width:                  0;
 $accordion-body-padding-y:                0.5rem;
 
 $accordion-button-padding-y:              0.5rem;
-$accordion-button-active-bg:              $white;
+$accordion-button-active-bg:              $body-bg;
 $accordion-button-active-color:           $body-color;
 
-$accordion-button-focus-border-color:     $white;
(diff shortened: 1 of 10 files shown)


NeOMakinG:
@rodriciru 

https://www.figma.com/file/LfVl5leeSKcVUhSaYwhbtM/New-Theme?node-id=236%3A10538

We should follow the mockups as much as possible, except if we have a really needed choice to do.

I guess the background on `My Account` pages is just a design choice which we shouldn't take the liberty to decide ourself while some designers worked on it

rodriciru:
Well, I'm not so sure, but I can't guarantee it.

First, I use this light gray (SCSS lighten($gray-100, 2%)) because I see on other forms, but maybe it should be white as Figma. My bad.

Supposing I choose white instead of this gray:
As the bg in the page is white, you can have a bg on the form, and because it's white you won't notice it. In other parts of the customer account, you have this white bg, so I deduced it's a CSS mistake o whatever. But as I say, I can't guarantee it.

And please, keep in mind that this imgs that I posted on this PR, **has my own bg color** for the body (gray), so when stock theme you won't notice the bg, just when you decide to change the bg

I do this, to maintain a consistence between pages on the customer account, when change the bg to other than white

Hlavtox:
@rodriciru I think I get your idea, but changing the background usually requires to add some paddings around content with white background. So it will never be just 'change one var and done'...

rodriciru:
> @rodriciru I think I get your idea, but changing the background usually requires to add some paddings around content with white background. So it will never be just 'change one var and done'...

That's right. And in this case it is. This was one of my problems, and I add some padding ( I copy from other form), because I think the drawbacks are less than the benefit in pro of customization.

You are the bosses on this theme, so whatever yours decide will be the best.

NeOMakinG:
It didn't move for a while, as we switched to default bootstrap colors and CSS vars, it's probably over already

if @rodriciru want's to change something again, feel free to pick it up, reopen it or open a new PR

PR #302, opened 2022-05-27
img responsive on modal shopingcart

rodriciru:
| Questions         | Answers
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     |  no

On no img product, when added to car, img has no responsive class, so its out of div element.
imagen
FIXED:
imagen

### DIFF
--- a/modules/ps_shoppingcart/modal.tpl
+++ b/modules/ps_shoppingcart/modal.tpl
@@ -27,7 +27,7 @@
                   <img
                     src="{$urls.no_picture_image.bySize.medium_default.url}"
                     loading="lazy"
-                    class="product-image"
+                    class="img-fluid product-image"
                   />
                 {/if}
               </div>


Hlavtox:
Nice catch!

NeOMakinG:
Did we even work on this modal yet?

Hlavtox:
No, because it doesn't work if I build it on WSL :D

mparvazi:
> Did we even work on this modal yet?

IMO needs a redesign.

Hlavtox:
Yeah, and not use like 5 different fonts/weights/color like in classic... 👍

mparvazi:
@Hlavtox 

Use [Windows Terminal](https://apps.microsoft.com/store/detail/windows-terminal/9N0DX20HK701)

Check WSL version:





Add npm and lampp directory to Antivirus or Windows Security exclude list for scan:

Screenshot 2022-05-30 115404

PR #304, opened 2022-05-30
Add BS form validation

mparvazi:
https://user-images.githubusercontent.com/85633460/170999553-ab95e479-8e4f-4fd4-a9c7-4962549ac9a8.mp4

### DIFF
--- a/src/js/constants/selectors-map.ts
+++ b/src/js/constants/selectors-map.ts
@@ -107,6 +107,10 @@ export const quantityInput = {
   alert: (param: string): string => `#js-product-line-alert--${param}`,
 };
 
+export const formValidation = {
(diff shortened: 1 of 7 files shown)


src/js/constants/selectors-map.ts
```diff
@@ -107,6 +107,10 @@ export const quantityInput = {
   alert: (param: string): string => `#js-product-line-alert--${param}`,
 };
 
+export const formValidation = {
+  default: '.needs-validation',
```
NeOMakinG, line 111: Mmmmh, not a big fan of this naming... nitpick but I would see something like `form-validation`

NeOMakinG (approved):
LGTM anyway

Issue #305, opened 2022-05-30
Product details shows detail that can confuse customers


rodriciru:
![imagen](https://user-images.githubusercontent.com/10741347/171025016-fcd6b104-7214-4944-92ff-2b5245aeb257.png)
this detail is the hook `actionProductOutOfStock` but the product has stock. Some customers can misunderstand this and think there is no stock

rodriciru:
I'm a little confused.
I see that action hooks only executes actions, no display nothing.
But devdocs say this:

> actionProductOutOfStock
>     This hook displays new action buttons if a product is out of stock 

So it displays something or not?. If not, I think it shouldn't be any li.details and only the hook outside the accordion

NeOMakinG:
I guess we could remove it in favor of product ability 

https://github.com/PrestaShop/PrestaShop/pull/25062

PR #306, opened 2022-05-30
Remove actionProductOutOfStock from product-details

rodriciru:
| Questions         | Answers
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/305

This hook is action type, so it must not display anything, so it makes no sense to put on product-details.tpl with markup code.
Also, I think this hook is useless. I can't find any hook::exec on PrestaShop code since 1.6 version. Anyway, I keep it just in case...

### DIFF
--- a/templates/catalog/_partials/product-details.tpl
+++ b/templates/catalog/_partials/product-details.tpl
@@ -1,11 +1,11 @@
 
-<div 
+<div
   class="info js-product-details accordion-item"
   id="product-details"
   data-product="{$product.embedded_attributes|json_encode}"
 >
   <h5 class="info__title accordion-header" id="product-details-heading">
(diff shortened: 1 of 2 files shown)


NeOMakinG:
@Hlavtox wdyt?

NeOMakinG:
Myself I would say to remove it, but I may miss something so

Hlavtox:
I would remove it, this hook has no purpose.

rodriciru:
> I would remove it, this hook has no purpose.

Complete remove from hummingbird or just from product-details.tpl?

NeOMakinG (approved):

NeOMakinG:
Thanks @rodriciru

PR #317, opened 2022-06-09
Improve button UI with icon

### LINKED ISSUE #241: Material Icons are not aligned vertically inside the buttons
**Current (the icon and the text are not aligned vertically when icons are not inside Flex container):**
Screenshot from 2022-04-28 03-59-13

**How BS fix this problem for [its icons](https://icons.getbootstrap.com/icons/cart-fill/) inside the components:**
```
.bi {
  display: inline-block;
  vertical-align: -.125em;
}
```

**For hummingbird (material icons) can be:**
```
.material-icons {
  font-size: 150%;
  vertical-align: -0.25em; 
}
```
_em is relative to the font-size of its direct or nearest parent._
bs

**Another example:**
Cart icon is a flex item so icon is aligned vertically.
User icon is NOT a flex item so `vertical-align: middle;` will be used for it, but it's not vertically aligned.
Screenshot from 2022-04-28 04-50-48

Fixed by `vertical-align: -0.25em;`
Screenshot from 2022-04-28 04-52-32

mparvazi:
Fixes #241

**btn-with-icon:**
Vertical alignment for icons inside buttons (inline-flex)

**The size of icon and gap will be changed based on button size (btn-sm, btn, btn-lg):**

Screenshot from 2022-06-11 17-43-16

Screenshot from 2022-06-11 17-42-13

## Variables

```scss
$default-button-icon-size: 1.5rem;
$default-button-icon-gap: 0.25rem;

$small-button-icon-size: 1.25rem;
$small-button-icon-gap: 0.125rem;

$large-button-icon-size: 1.875rem;
$large-button-icon-gap: 0.375rem;
```

## SCSS

```scss
.btn-with-icon {
  display: inline-flex;
  gap: $default-button-icon-gap;
  align-items: center;
  justify-content: center;

  .material-icons {
    font-size: $default-button-icon-size;
  }

  &.btn-sm {
    gap: $small-button-icon-gap;

    .material-icons {
      font-size: $small-button-icon-size;
    }
  }

  &.btn-lg {
    gap: $large-button-icon-gap;

    .material-icons {
      font-size: $large-button-icon-size;
    }
  }
}
```


SharakPL:
All of this paddings and rtl can be avoided if you simply use `gap` on the button 😉

NeOMakinG (approved):
I guess it looks fine, but didn't tested it yet so!

NeOMakinG:
@Hlavtox @SharakPL if you find something wrong with this one afterwhile, please open an issue or a PR :)

Looked fine and wants to clean a bit the current PR flow of this repository in order to avoid stacking too much stuff

SharakPL:
Looks good for me.

One thing though. I see some duplications throughout the theme (not specific to this PR):

https://github.com/PrestaShop/hummingbird/blob/5c5319ffc7deb4775e39618a7a8731d498a48c09/src/scss/partials/_helpers.scss#L39-L49

Can I remove these and just use Bootstrap classes `w-100` and `h-100` instead?  `text-muted` isn't really the same as `text-gray` but I'm guessing the purpose of it is the same, so it also could be used here.

https://github.com/PrestaShop/hummingbird/blob/5c5319ffc7deb4775e39618a7a8731d498a48c09/src/scss/partials/_helpers.scss#L51-L53

This also could be removed. Adding ` role="button"` to the element does exactly the same.

NeOMakinG:
> Looks good for me.
> 
> One thing though. I see some duplications throughout the theme (not specific to this PR):
> 
> https://github.com/PrestaShop/hummingbird/blob/5c5319ffc7deb4775e39618a7a8731d498a48c09/src/scss/partials/_helpers.scss#L39-L49
> 
> Can I remove these and just use Bootstrap classes `w-100` and `h-100` instead? `text-muted` isn't really the same as `text-gray` but I'm guessing the purpose of it is the same, so it also could be used here.
> 
> https://github.com/PrestaShop/hummingbird/blob/5c5319ffc7deb4775e39618a7a8731d498a48c09/src/scss/partials/_helpers.scss#L51-L53
> 
> This also could be removed. Adding ` role="button"` to the element does exactly the same.

Feel free, you're welcome!

PR #318, opened 2022-06-09
Remove override rules for blockreassurance

### LINKED ISSUE #277: Do not override the blockreassurance appearance colors
If we try to override inline color rules for blockreassurance by `!important `,
so what's the benefit of **Appearance** config in this module?

Screenshot from 2022-05-10 11-51-56

mparvazi:
Fixes #277

The [blockreassurance](https://github.com/PrestaShop/blockreassurance) module has appearance config.

Screenshot 2022-06-09 at 20-32-45 Hummingbird printed sweater

### DIFF
--- a/src/scss/custom/modules/_blockreassurance.scss
+++ b/src/scss/custom/modules/_blockreassurance.scss
@@ -1,20 +1,7 @@
 .blockreassurance_product {
-  .item-product {
-    // stylelint-disable-next-line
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
Ok! Nice catch

matthieu-rolland:
thank you @mparvazi

PR #320, opened 2022-06-09
Style social follow module

rodriciru:
| Questions         | Answers
| Description?      | This module has wrong style
| BC breaks?        | no
| Deprecations?     |  no

image


Hlavtox (changes requested):
- container should be used only to limit page width
- No need to have extra row and col-12

src/scss/custom/modules/_social-follow.scss
```diff
@@ -0,0 +1,48 @@
...
+        width: 2.5rem;
+        height: 2rem;
+        margin-right: 1.3rem;
+        margin-bottom: 1.3rem;
+        background-size: contain;
+      }
+
+      &:hover {
+        background-color: #24b9d7;
+      }
```
SharakPL, line 33: I don't like this. `a` is the active element so that's the element that should be styled. You can even skip `cursor: pointer` on it, because it's default. `li` is just the container.

IMO it shouldn't even be `ul` in the first place, because it serves no purpose here. Just a simple `div` with links.

Another problem is the explicit size. The best option here is to use SVGs inlined in the template with `height="1em"` and determine the size with `font-size`. I wouldn't even use `width` or `height` on links. Just `gap` to determine spaces between them. Unless they're going to be styled as buttons with borders, then:
```
a {
  display: inline-flex;
  align-items: center;
  justify-content: center;
  width: 1.5em;
  height: 1.5em;
  border: 1px solid currentcolor;
  // optionally
  border-radius: 50%;
}
```

SharakPL (changes requested):

NeOMakinG:
@SharakPL are you on slack? Would like to discuss 5 mins about something with you

rodriciru:
Please feel free to make changes on your own if you have access.
I just copy and paste the code from classic theme, and adapt it just the necessary (and one thing more) to make it work.
I suppose, that the designer will make a new design, and it won't be like this, so I do just the minimum.
Also, I'm getting married the next week, so I will be off for a long time. :)

matks:
> How do I join prestashop.slack.com? It only accepts @prestashop.com emails

https://www.prestashop-project.org/slack/

NeOMakinG:
> Please feel free to make changes on your own if you have access. I just copy and paste the code from classic theme, and adapt it just the necessary (and one thing more) to make it work. I suppose, that the designer will make a new design, and it won't be like this, so I do just the minimum. Also, I'm getting married the next week, so I will be off for a long time. :)

Wow, what a great news ser! Have fun :)

src/scss/custom/modules/_social-follow.scss
NeOMakinG: @SharakPL we need to deal with the module, we can't use an inline SVG unfortunately :/

src/scss/custom/modules/_social-follow.scss
```diff
@@ -0,0 +1,20 @@
+.block-social {
+  background-color: $primary;
```
NeOMakinG, line 2: I want to avoid that so it stay reusable

src/scss/custom/modules/_social-follow.scss
SharakPL: > we can't use an inline SVG unfortunately

Why not? I changed my theme to use inline SVG in this module with no problem.

NeOMakinG: Well, I would agree, and I'll push it using your svgs, but it adds some maintenance, as when the module get updated like adding a new social network, the theme needs to deal with it!

SharakPL: It's still much easier to simply add another condition and path. Icons are from FontAwesome so if this font would be used, maybe instead of Material Icons, then we could avoid svgs altogether and just use simple classes. Bootstrap Icons also have social media icons, though they don't match as good as those in FA (eg. all squares, like in example above).

Best option would be to not hardcode any social media and just allow user to add any link and pick the icon from FA font. Similar like _Icon_ component in _CreativeElements_ module:
obraz

Progi1984 (approved):

src/scss/custom/modules/_social-follow.scss
NeOMakinG: Good point, but I guess we should bump it to a major version, as it's a huge BC for every user of this module

Until it's not done, I guess we can just accept this PR, and then remove it when the module has been updated!

mparvazi:
Screenshot from 2022-07-08 15-05-50

The `fs-0` used for hiding labels, but we will have over-height on 320px because of wrapping.
I think it's better to use `d-none `or `visually-hidden` for labels.

SharakPL:
@NeOMakinG does it even need these labels? I use `title` instead in my example, which is also required for html validation and is good for screen readers.

NeOMakinG:
@SharakPL @mparvazi @Hlavtox removed labels, replaced by titles, feel free to merge whenever you want :)

src/scss/custom/modules/_social-follow.scss
```diff
@@ -0,0 +1,11 @@
+.block-social {
+  span {
+    font-size: 0;
+  }
+
```
SharakPL, line 5: This is obsolete now. I mean `span` styling.

src/scss/abstract/variables/bootstrap/_grid.scss
```diff
@@ -0,0 +1 @@
+$enable-cssgrid: true;
```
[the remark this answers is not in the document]
NeOMakinG: No it's not, but that could be useful for theme developers isn't it? It enables grid helpers

SharakPL (approved):

PR #321, opened 2022-06-10
Add row class fallback to productlist.tpl and remove every css override

rodriciru:
| Questions         | Answers
| Description?      | .row class is missing so BS wont work
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     | no

Before:
imagen
After:
imagen

### DIFF
--- a/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
+++ b/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
@@ -5,7 +5,7 @@
 <section class="best-sellers-products mt-3">
     <div class="container">
         {include file="components/section-title.tpl" title={l s="Best Sellers" d="Shop.Theme.Catalog"}}
-        {include file="catalog/_partials/productlist.tpl" products=$products cssClass="row" productClass="col-6 col-lg-4 col-xl-3"}
+        {include file="catalog/_partials/productlist.tpl" products=$products productClass="col-6 col-lg-4 col-xl-3"}
         <div class="best-sellers-products-footer text-center">
             <a class="all-product-link btn btn-outline-primary" href="{$allBestSellers}">
                 {l s='All best sellers' d='Shop.Theme.Catalog'}<i class="material-icons rtl-flip">&#xE315;</i>
--- a/modules/ps_categoryproducts/views/templates/hook/ps_categoryproducts.tpl
+++ b/modules/ps_categoryproducts/views/templates/hook/ps_categoryproducts.tpl
@@ -9,5 +9,5 @@
     {include file="components/section-title.tpl" title={l s='%s other products in the same category' sprintf=[$products|@count] d='Shop.Theme.Catalog'}}
   {/if}
 
-  {include file="catalog/_partials/productlist.tpl" products=$products cssClass="row" productClass="col-6 col-lg-4 col-xl-3"}
+  {include file="catalog/_partials/productlist.tpl" products=$products productClass="col-6 col-lg-4 col-xl-3"}
 </section>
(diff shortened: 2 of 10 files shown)


matthieu-rolland (approved):

Hlavtox:
We should make some kind of a fallback for modules not passing this parameter I think. 🤔

NeOMakinG:
@Hlavtox feel free to merge when you feel like :)

jolelievre (approved):
Thanks @rodriciru

jolelievre:
@Hlavtox @NeOMakinG 

I don't really know the process for validating/merging PRs on this repo, is there some QA done at some point? If yes by who?

mparvazi:
Hello @rodriciru 

For modules, the templates inside `<section>` it's better to be wrapped by `<div class="container"> `

Currently missed on:
ps_categoryproducts.tpl
ps_crossselling.tpl
ps_viewedproduct.tpl


Screenshot from 2022-07-11 15-10-40

mparvazi (changes requested):
Hello @rodriciru 

For modules, the templates inside `<section>` must be wrapped by `<div class="container"> `

Currently missed on:
ps_categoryproducts.tpl
ps_crossselling.tpl
ps_viewedproduct.tpl

NeOMakinG:
> For modules, the templates inside `<section>` block must be wrapped by `<div class="container"> `
> 
> Currently missed on:
> 
> * ps_categoryproducts.tpl
> * ps_crossselling.tpl
> * ps_viewedproduct.tpl

I'll continue taking over this PR, will add this later in the day, thanks for testing @mparvazi

NeOMakinG:
> @Hlavtox @NeOMakinG
> 
> I don't really know the process for validating/merging PRs on this repo, is there some QA done at some point? If yes by who?

There is no QA process for the moment, it will probably be set up after the first beta release, most of the time, contributors and myself are still testing the PR before merging

And I really love that honestly, I'll try to keep a similar workflow in the future as a test, if the QA agree, would be awesome as it's really flexible and seems like it's not a pain for the moment

Hlavtox:
@jolelievre Yeah, for now, the development phase, we just check each others work and merge it. It would develop 10x slower if we had to wait for QA every time.

mparvazi (approved):

mparvazi:
@jolelievre 

No problem, I try to test everything before merge.

Hlavtox:
@mparvazi @NeOMakinG @SharakPL @rodriciru I am still thinking, can we do it somehow, that we define the columns only in productlist.tpl? 🤔 Is it possible to detect the layout in that file and have 3 configs to automatically use? So the tpls that include it don't have to think about it.

So it can look only like: `{include file="catalog/_partials/productlist.tpl" products=$products}`

NeOMakinG:
> @Hlavtox does it mean there is at least a developer who gets the branch and tests it a bit? or it's only review, approve and merge?

Most of the time we are testing PRs ourself, it doesn't mean that we merge blindly

NeOMakinG:
> @mparvazi @NeOMakinG @SharakPL @rodriciru I am still thinking, can we do it somehow, that we define the columns only in productlist.tpl? 🤔 Is it possible to detect the layout in that file and have 3 configs to automatically use? So the tpls that include it don't have to think about it.
> 
> So it can look only like: `{include file="catalog/_partials/productlist.tpl" products=$products}`
> 
> I know we implemented this in classic together, but MAYBE there is a better solution.

There is 2 different layout, listing from homepage and listing inside a container, I guess this is why we implemented this

Maybe there is a better solution as @SharakPL we could use some layouts with grids

Btw, as this is another refactor discussion, I would suggest merging this PR, and create an issue about improving that system, or we take the risk to get stuck between 2 buggy states if no one takes the refactoring subject :)

SharakPL:
@NeOMakinG I'm on it! 🙋

PR #325, opened 2022-06-21
Clean helpers, add programmable media queries

SharakPL:
| Questions         | Answers
| Description?      | Removes some obsolete classes, makes media queries programmable to match changes in variables and limit the size of the output
| Type?             | improvement / refacto
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | Please indicate how to best verify that this PR is correct.

If you need additional helper classes for each breakpoint, don't create another media query. Just add your class to the loop.

### DIFF
--- a/src/scss/partials/_helpers.scss
+++ b/src/scss/partials/_helpers.scss
@@ -40,10 +40,6 @@
   color: $gray-700;
 }
 
-.h-full {
-  height: 100%;
-}
-
 .cursor-pointer {
   cursor: pointer;
 }
--- a/src/scss/partials/_helpers.scss
+++ b/src/scss/partials/_helpers.scss
@@ -18,15 +18,17 @@ $container-max-width in $container-max-widths {
     .container--limited-#{$breakpoint} {
       max-width: $container-max-width;
     }
+
+    .w-#{$breakpoint}-auto {
+      width: auto;
+    }
+
+    .w-#{$breakpoint}-100 {
+      width: 100%;
+    }
   }
 }
 
 .text-gray {
   color: $gray-700;
-}
-
-.w-md-auto {
-  @include media-breakpoint-up(md) {
-    width: auto;
-  }
 }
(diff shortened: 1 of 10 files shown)


Hlavtox (approved):
Nice work dude!!!

src/scss/partials/_helpers.scss
```diff
@@ -12,48 +12,22 @@
...
+      max-width: $container-max-width;
+    }
+
+    .w-#{$breakpoint}-auto {
+      width: auto;
+    }
+
+    .w-#{$breakpoint}-100 {
+      width: 100%;
+    }
```
NeOMakinG, line 27: I bet this is something that would probably interest the bootstrap team  isn't it?

NeOMakinG (approved):

NeOMakinG:
Thanks @SharakPL

src/scss/partials/_helpers.scss
SharakPL: [They already use it](https://github.com/twbs/bootstrap/blob/main/scss/_containers.scss) :)
And that's what I've been trying to say from the start of this theme. Bootstrap has a lot already done just waiting to be used. No need to create duplications, when almost whole theme could be done just by setting up default variables. Not much places where you actually need some custom css.

SharakPL:
```scss
@each $name, $gray-variant in $grays {
  .text-gray-#{$name} {
    color: $gray-variant;
  }
}
```
This would generate `.text-gray-100`, `.text-gray-200`, etc. just like Tailwind. Would this also be useful or is `.text-light`, `text-dark` (both default BS) and `.text-gray` (theme helpers) enough?

SharakPL: That's why I'm only asking if this would be helpful right now. I can imagine few places where you would want text with different shades of gray, but that's just a suggestion if it's ever needed.

PR #331, opened 2022-07-05
Remove fixed sizes of ps_banner

NeOMakinG:
| Questions         | Answers
| Description?      | We don't want to force some sizes on any images
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/268.
| How to test?      | See with the 2.1.2 version of ps_banner if it's displayed properly
| Possible impacts? | ps_banner module usage


modules/ps_banner/ps_banner.tpl
```diff
@@ -4,7 +4,7 @@
  *}
 <a class="banner" href="{$banner_link}" title="{$banner_desc}">
```
mparvazi, line 5:
```suggestion
<a class="d-block" href="{$banner_link}" title="{$banner_desc}">
```

SharakPL: Adding `d-block` class is still good. No more css needed.

NeOMakinG:
Closing in favor of https://github.com/PrestaShop/hummingbird/pull/269/files

PR #332, opened 2022-07-05
Add !important to our custom helpers

NeOMakinG:
| Questions         | Answers
| Description?      | Bootstrap core helpers are using !important, our own helpers should use it too, overwise it won't work
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | You can compare with the Product Added To Cart modal if needed

### DIFF
--- a/src/scss/partials/_helpers.scss
+++ b/src/scss/partials/_helpers.scss
@@ -19,10 +19,12 @@
     }
 
     .w-#{$breakpoint}-auto {
+      // stylelint-disable-next-line declaration-no-important
       width: auto !important;
     }
 
     .w-#{$breakpoint}-100 {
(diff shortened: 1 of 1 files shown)


SharakPL:
Thanks. I forgot about it.
Shouldn't there be space required by stylelint before additions like `!important` or `!default`?

NeOMakinG:
> Thanks. I forgot about it. Shouldn't there be space required by stylelint before additions like `!important` or `!default`?
> 
> Actually it already is. That's why linter is not happy. Also `declaration-no-important` should be disabled for this file.

Yeah forgot to run it 🤦

kpodemski:
"Make linter great again" ❤️ 

thx @NeOMakinG

Issue #256, opened 2022-05-02
Categories with subcategories on category page are wrongly displayed


NeOMakinG:
![image](https://user-images.githubusercontent.com/14963751/166239927-b84f856e-ffca-4513-a2d4-1d0233c47138.png)

We should probably add a proper collapse there, we definitely can't show every categories by default because some big websites won't agree with that choice, but it should be easily adaptable

Hlavtox:
We should add an option to the the module called "Open the current category". ;-)

SharakPL:
How about using 1 chevron arrow and just rotating it 180deg instead of showing/hiding + and -?

mparvazi:
With chevron it will be as same as faceted search UI.

NeOMakinG:
> With chevron it will be as same as faceted search UI.

and it looks pretty fine to me, but this is obviously something that the product team will notice while reviewing

PR #333, opened 2022-07-06
Use accordions as categories with childs on categorytree

NeOMakinG:
| Questions         | Answers
| Description?      | On ps_categorytree, we want a cool design, easy to maintain, bootstrap collapse with the accordion style is a perfect fit for it!
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/256.
| How to test?      | Go on category page, see screenshot
| Possible impacts? | categorytree

image

### DIFF
--- a/modules/ps_categorytree/views/templates/hook/ps_categorytree.tpl
+++ b/modules/ps_categorytree/views/templates/hook/ps_categorytree.tpl
@@ -8,38 +8,20 @@
 {function name="categories" nodes=[] depth=0}
   {strip}
     {if $nodes|count}
-      <ul class="{$componentName}-list">
+      <ul class="{$componentName}__list" data-depth="{$depth}">
         {foreach from=$nodes item=node name="categories"}
-          <li class="{$componentName}-item" data-depth="{$depth}">
-            {if $depth===0}
+          <li class="{$componentName}__item">
+            <div class="{$componentName}__item__header{if $node.children} {$componentName}__item__header--parent{/if}">
+              <a class="{$componentName}__item__link" href="{$node.link}">{$node.name}</a>
               {if $node.children}
-                <div class="d-flex align-items-center justify-content-space-between">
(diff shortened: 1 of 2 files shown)


Hlavtox (changes requested):

SharakPL:
Some improvements to consider:
1. Utility classes are great, but I would avoid them inside loops. Especially big loops like menu or category tree. It's better to create a simple set of styles and target html tags or data-depth attributes than repeat a long lists of classes for each item.
2. `ul` determines the depth so that's where `data-depth` attribute should be places. No need to repeat it for every `li` element.
3. I never liked that only links of nodes with children are wrapped in a div. This makes whole thing harder to style, especially when only some nodes of the same depth have children. Every link should be wrapped the same way.
4. It's required to have current category branch shown by default so we need more states eg. `node.current=true` for current category and `node.active=true` for current category and all it's parents.

```smarty
<ul class="{$componentName}-list" data-depth="{$depth}">
  {foreach from=$nodes item=node name="categories"}
    <li class="{$componentName}-item {if $node.current}current{/if}">
      <div class="{$componentName}-item-header">
        <a class="{$componentName}-{if $depth===0}item{else}child{/if}-link" href="{$node.link}">{$node.name}</a>
        {if $node.children}
          <button class="collapsed" data-bs-toggle="collapse" data-bs-target="#exCollapsingNavbar{$node.id}"></button>
        {/if}
      </div>
      {if $node.children}
        <div class="collapse {if $node.active}show{/if}" id="exCollapsingNavbar{$node.id}">
          {categories nodes=$node.children depth=$depth+1}
        </div>
      {/if}
    </li>
  {/foreach}
</ul>
```

Hlavtox:
@SharakPL @NeOMakinG We should add a function to keep the current category open, a lot of client want this.

It was a nice feature of 1.6

NeOMakinG:
> @SharakPL @NeOMakinG We should sometime add a function to keep the current category open, a lot of client want this.
> 
> It was a nice feature of 1.6

We need to add this to a list somewhere or open an issue

PR #334, opened 2022-07-07
fix return table

JBLach:
| Questions         | Answers
| Description?      | I have fixed return table
| BC breaks?        |  no
| Deprecations?     |  no
| How to test?      | see screenshots below.

## Before
image

## After
image

### DIFF
--- a/templates/customer/_partials/order-detail-return.tpl
+++ b/templates/customer/_partials/order-detail-return.tpl
@@ -135,65 +135,86 @@
       </table>
     </div>
 
-    <div class="order-items d-block d-sm-block d-md-none box">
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
Gm @JBLach and thanks for the contrib

From a code point of view, I couldn't find something blocking us from merging your work, looks quite healthy to me

From a design point of view, it doesn't look perfect, but I guess we don't have any mockups and the PR is improving it a lot, it's acceptable enough.

🚀

NeOMakinG:
@Hlavtox @SharakPL @mparvazi feel free to merge if you feel like it

Issue #340, opened 2022-07-08
Viewport check: 320px


mparvazi:
## Header logo
The logo container display must be defined as block, in current implementation there is overlap with search icon if the logo width is more than 150px.

![logo](https://user-images.githubusercontent.com/85633460/177984701-d550c855-d78c-42bd-ba8c-84f8865ed63a.png)

## User menu button (Fixed by #343)
When customer is not logged-in there is an unwanted right padding, so we have extra space on the right side.

![user](https://user-images.githubusercontent.com/85633460/177984976-45425e70-d12b-4fca-8861-c0a1a90f9f97.png)

## Newsletter block
Need to use full-width style for input elements.

![newsletter](https://user-images.githubusercontent.com/85633460/177985175-f3a5d562-3bd9-4e77-9d88-ac210a9860e8.png)

## Pagination
We have overflow when Next and Previous buttons are displayed at the same time. I think we need to hide text label for mobile screen.

![pagination](https://user-images.githubusercontent.com/85633460/177985378-75f92bcb-4b45-40bd-ac26-d17aeb2541c1.png)

 ## Account page (Fixed by #357)
It's better to use an unstyle button for Back and place it under the page title.

![back](https://user-images.githubusercontent.com/85633460/177985595-afe1c645-68f4-4af6-9173-1a316c4d5ca9.png)

## Products list
Just a suggestion: showing one product per column on mobile screen.

![products](https://user-images.githubusercontent.com/85633460/177985636-95c6248a-5e5c-42dc-a6c4-9ce08e6cdbb5.png)

mparvazi:
I tried to change Pagination display for XS screen. Not bad if page numbers aren't too much:

![Screenshot from 2022-07-08 23-38-35](https://user-images.githubusercontent.com/85633460/178055757-7f4eaf94-8d56-4f21-9f34-c957eef24da7.png)

But what about this?!

![Screenshot from 2022-07-08 23-35-16](https://user-images.githubusercontent.com/85633460/178055706-17ff076b-3f61-4d1a-90ad-51af0062afcd.png)

mparvazi:
Wdyt about this?
A compact version for XS screen:

![Screenshot from 2022-07-09 00-11-10](https://user-images.githubusercontent.com/85633460/178059786-5ad8eb2e-a293-4b30-9454-af1a394beb0d.png)

Support page numbers until 999 :P

mparvazi:
[...]

PR #342, opened 2022-07-08
Fix header logo block

mparvazi:
Part of #340 (Header logo)

### DIFF
--- a/src/scss/custom/layout/_header.scss
+++ b/src/scss/custom/layout/_header.scss
@@ -32,10 +32,6 @@ $component-name: header;
       border-bottom: var(--header-bottom-border-bottom);
     }
 
-    .navbar-brand {
-      padding: 0;
-    }
-
     .nav-link {
       &:active,
       &.active {
--- a/templates/_partials/header.tpl
+++ b/templates/_partials/header.tpl
@@ -51,11 +51,9 @@
   <div class="{$headerBottomName}">
     <div class="container {$headerBottomName}__container">
       <div id="_desktop_logo" class="logo logo-desktop order-1 order-xl-0">
-        {if $page.page_name == 'index'}<h1>{/if}
-        <a class="navbar-brand" href="{$urls.pages.index}">
+        <a class="navbar-brand d-block" href="{$urls.pages.index}">
           <img class="logo img-fluid" src="{$shop.logo_details.src}" alt="{$shop.name}" loading="lazy" width="{$shop.logo_details.width}" height="{$shop.logo_details.height}">
         </a>
-        {if $page.page_name == 'index'}</h1>{/if}
       </div>
 
       {hook h='displayTop'}


Hlavtox (approved):
Approve, but thinking about that H1 practice….

However, the shop developer should be responsible and put a proper H1 somewhere into homepage. 👍

mparvazi:
@Hlavtox 

If you think it's important, we can keep h1 wrapper and fix padding for it.

NeOMakinG:
> Approve, but thinking about that H1 practice….
> 
> However, the shop developer should be responsible and put a proper H1 somewhere into homepage. 👍

That h1 makes no sense to me honestly

mparvazi:
> > Approve, but thinking about that H1 practice….
> > However, the shop developer should be responsible and put a proper H1 somewhere into homepage. +1
> 
> That h1 makes no sense to me honestly

This is what Google do in its pages for <h1>:

Result page (when it doesn't need to be shown visually):

```html
<h1 class="Uo8X3b">Accessibility links</h1>
```

```css
.Uo8X3b {
  clip: rect(1px,1px,1px,1px);
  height: 1px;
  overflow: hidden;
  position: absolute;
  white-space: nowrap;
  width: 1px;
  z-index: -1000;
  user-select: none;
}
```

In some pages (when visually need to show):
```html
<div>
   <h1>
        <a href="https://www.google.com/?hl=en_GB" class="chtwBf">
            <img alt="Google" src="...">
        </a>
    </h1>
   <h1>page title</h1>
</div>
```


So currently suspend this PR.
I think need to choose a way for using <h1> tag.

mparvazi:
My suggestion:

```
<img src="shop_logo"/>
<h1 class="visually-hidden">{$page.meta.title}</h1>
```
There is no problem for multiple `<h1>` tags, if developer wants to add own h1 tag.

`With HTML5, it’s not a problem at all. It is common to have multiple H1 heading tags in different parts of your page, in your website’s template, theme or other sections. Regardless of whether you use HTML5 or not, having multiple H1 elements or multiple heading tags of the same type on a page is completely fine. `

NeOMakinG:
> My suggestion:
> 
> ```
> 
> <h1 class="visually-hidden">{$page.meta.title}</h1>
> ```
> 
> There is no problem for multiple `<h1>` tags, if developer wants to add own h1 tag.
> 
> `With HTML5, it’s not a problem at all. It is common to have multiple H1 heading tags in different parts of your page, in your website’s template, theme or other sections. Regardless of whether you use HTML5 or not, having multiple H1 elements or multiple heading tags of the same type on a page is completely fine. `

This is what I was about to say

NeOMakinG:
Maybe we can just keep this one, no one ever created an issue about this

SharakPL:
It is fine for html validation, but it's not fine for SEO. Afaik it's a rule to always have just one h1 per page.

SharakPL:
@mparvazi why did you close it?

mparvazi:
@SharakPL 
let me think more about that.
just want to test different situations for logo size.

btw the related issue is still open.

SharakPL:
IMHO no need to think about that. This header should be completely different anyway. There should be just the logo on top and maybe social media icons or switch for showing all contact information in a modal. All the icons should be in the action bar fixed to the bottom - mobile app style 🤘

This way you'll have all the space for the logo and even additional icons like wishlist.

PR #344, opened 2022-07-08
Fix email subscription form UX on XS screen

mparvazi:
Part of #340 (Newsletter block)

## Before
Screenshot from 2022-07-08 20-51-41

## After (With last commit)
Screenshot from 2022-07-25 11-27-11

### DIFF
--- a/modules/ps_emailsubscription/views/templates/hook/ps_emailsubscription.tpl
+++ b/modules/ps_emailsubscription/views/templates/hook/ps_emailsubscription.tpl
@@ -5,8 +5,8 @@
 
 {$componentName = 'email-subscription'}
 
-<div class="{$componentName} py-4">
-  <div class="container">
+<div class="{$componentName} px-0 py-4">
+  <div class="container px-1">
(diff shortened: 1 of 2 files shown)


Hlavtox:
I think we can put them like this by default (only padding-bottom) and put them on a line with bigger screen with media query. Less code. ;-)

Hlavtox:
Also, WDYT?
178031718-0eac72e5-853b-4127-9c5d-10751bb5af38

mparvazi:
Like this?

Screenshot from 2022-07-09 12-43-07

SharakPL:
How about full width button when it's under the input? Otherwise [this version](https://github.com/PrestaShop/hummingbird/pull/344#issuecomment-1179501684) is good. Could be also `>` instead of `Subscribe` for mobile.

mparvazi:
Screenshot from 2022-07-25 11-27-11

NeOMakinG (approved):

atomiix (approved):

PR #345, opened 2022-07-08
Fix product list column counts on XS screen

mparvazi:
Part of #340 (Products list)

https://user-images.githubusercontent.com/85633460/178060181-71f3c1b4-c47a-41b8-8ecc-8db7d4b8f92a.mp4

### DIFF
--- a/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
+++ b/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
@@ -5,7 +5,7 @@
 <section class="best-sellers-products mt-3">
     <div class="container">
         {include file="components/section-title.tpl" title={l s="Best Sellers" d="Shop.Theme.Catalog"}}
-        {include file="catalog/_partials/productlist.tpl" products=$products cssClass="row" productClass="col-6 col-lg-4 col-xl-3"}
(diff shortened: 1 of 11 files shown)


Hlavtox:
I liked the original behavior more - it's specified like this in the design also - https://www.figma.com/file/LfVl5leeSKcVUhSaYwhbtM/New-Theme?node-id=173%3A12443.

You can fit more things like this. I would only reduce the gap between the columns to save space. So we have same padding between items as on the container.

SharakPL:
@Hlavtox on 320px fitting 2 in a row doesn't look good at all. It may look good enough on mockup which is made to look good, but in real life with longer product names, more digits in prices and additional content on the product image like more labels or blockwishlist's _like_ button it will look badly. IMHO product-list item should be at least 200px wide.

Hlavtox:
@SharakPL iPhone XS, one of our shops.

B3253645-E46C-4F8B-B1D8-C82034DDBF2E

NeOMakinG:
It would be too hard to take over and the lists should be switched to the grid, it would be easier

Hlavtox:
That's why I leave a padding on the top, for a label ;-)

Snímek obrazovky 2022-12-21 153345

Issue #346, opened 2022-07-09
Viewport check: 768px


mparvazi:
## Products filters row
Dropdown menu and label have text wrapping.

![Screenshot from 2022-07-09 11-22-41](https://user-images.githubusercontent.com/85633460/178096156-4e404dcd-1569-4235-8ed5-3b17822dc163.png)

## Quickview modal
Thumbnails width will be calculated based on viewport size, so need to use fullscreen modal on < 992px.

![Screenshot from 2022-07-09 11-37-48](https://user-images.githubusercontent.com/85633460/178096190-2ddeedf2-ef14-464a-b608-d2119c0fe27d.png)

SharakPL:
IMO sorting should be in new line and modal definitely should be wider here. Although I'm not sure about viewport related size of thumbs. It should be a carousel with auto-width.

NeOMakinG:
> IMO sorting should be in new line and modal definitely should be wider here. Although I'm not sure about viewport related size of thumbs. It should be a carousel with auto-width or even better something like swiper's [freemode](https://swiperjs.com/demos#freemode)

Mmh, about thumbs, we discussed it with the @PrestaShop/product-team 

Did we take a decision on this? I guess we wanted these thumbs on the left of the product page image

This topic is on the table and keeps coming back, it looks like there are some merchant's preferences, I wonder if we should make it configurable in multiple ways to everyone would be happy!

SharakPL:
Thumbs on the left just use more space and leave a lot of free space under the main image. Plus on mobile you would still have to move the thumbs under, that would require more css and maybe even some js, so there's really no point. I would keep it **always** under the main image.

mparvazi:
> Thumbs on the left just use more space and leave a lot of free space under the main image. Plus on mobile you would still have to move the thumbs under, that would require more css and maybe even some js, so there's really no point. I would keep it **always** under the main image.

I think we had discuss with @Hlavtox 
Some merchants wants to display all thumbs at the same time.
Definitely we can not use vertical thumbs in this situation.

Hlavtox:
Vertical thumbnails are just bad, period.

Regarding horizontal under image - some clients want a carousel, some clients want them all.
I vote for showing them all, because that 40px is not that much space, you don't have to click any arrows and you immediately see what you want to show.

NeOMakinG:
But, why don't we think about making it configurable? I guess both ways are possible and this is probably one of the place where being flexible makes sense

[...]

PR #355, opened 2022-07-20
fixes on 768px viewport

JBLach:
| Questions         | Answers
| Description?      | Fixed quickview modal and sorting  in 768px viewport
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | Fixes [#346]

##Before
image
image

##After
image
image

### DIFF
--- a/templates/catalog/_partials/products-top.tpl
+++ b/templates/catalog/_partials/products-top.tpl
@@ -4,7 +4,7 @@
  *}
 <div id="js-product-list-top" class="products-selection">
   <div class="products-selections-filters row">
-    <div class="col-12 col-md-6 total-products order-2 order-md-1 mt-4 mt-md-0">
+    <div class="col-12 col-md-12 col-lg-6 total-products order-2 order-lg-1 mt-4 mt-lg-0">
       {if $listing.pagination.total_items> 1}
         <p>{l s='There are %product_count% products.' d='Shop.Theme.Catalog' sprintf=['%product_count%' => $listing.pagination.total_items]}</p>
       {elseif $listing.pagination.total_items> 0}
@@ -12,7 +12,7 @@
       {/if}
     </div>
 
-    <div class="col-md-6 order-1 order-md-2">
+    <div class="col-md-12 col-lg-6 order-1 order-lg-2">
       <div class="d-flex align-items-center justify-content-md-end sort-by-row">
         {block name='sort_by'}
           {include file='catalog/_partials/sort-orders.tpl' sort_orders=$listing.sort_orders}
--- a/templates/catalog/_partials/products-top.tpl
+++ b/templates/catalog/_partials/products-top.tpl
@@ -4,7 +4,7 @@
  *}
 <div id="js-product-list-top" class="products-selection">
   <div class="products-selections-filters row">
-    <div class="col-12 col-lg-6 total-products order-2 order-lg-1 mt-4 mt-lg-0">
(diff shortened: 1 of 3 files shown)


templates/catalog/_partials/products-top.tpl
```diff
@@ -4,15 +4,15 @@
  *}
 <div id="js-product-list-top" class="products-selection">
   <div class="products-selections-filters row">
-    <div class="col-12 col-md-6 total-products order-2 order-md-1 mt-4 mt-md-0">
+    <div class="col-12 col-md-12 col-lg-6 total-products order-2 order-lg-1 mt-4 mt-lg-0">
```
Hlavtox, line 7: @JBLach Hi Jakub, can we please simplify this? `col-12 col-md-12 col-lg-6` can be just `col-lg-6`

The same please on line 15 `col-md-12 col-lg-6` -> `col-lg-6`.

Otherwise, I think it's good. Maybe just there will be a better way to specify the modal width. I would just say that modal will be `100% - some padding`, up to certain max width. WDYT @NeOMakinG @mparvazi

JBLach: Hi, I followed the shopingcart modal example, maybe we could create a helper class for modals below the 992px viewport so we don't have to overwrite every modal that has too much content, something similar to BS class `modal-fullscreen-lg-down` but only with maximum width with padding.  WDYT @Hlavtox @NeOMakinG @mparvazi

Hlavtox: > maybe we could create a helper class for modals below the 992px viewport so we don't have to overwrite every modal that has too much content, something similar to BS class modal-fullscreen-lg-down but only with maximum width with padding.

Yes, that would be good!

Change this also please - `col-12 col-lg-6` to `col-lg-6`

mparvazi:
Hello @JBLach 

Please add this for sm view:
```scss
    @include media-breakpoint-only(sm) {
      max-width: 100%;
      margin: 1rem;
    }
```

Screenshot from 2022-07-21 12-28-37

mparvazi (approved):
Thanks @JBLach

Hlavtox:
```
@include media-breakpoint-down(md) {
    max-width: 100%;
    margin: 1.75rem;
}
```

I think we can do something like this to make it cleaner, no?

mparvazi:
> ```css-scss
> @include media-breakpoint-down(md) {
>     max-width: 100%;
>     margin: 1.75rem;
> }
> ```
> 
> I think we can do something like this to make it cleaner, no?

For sm-down media, BS controls the modal with different style.
Just need to change the BS default for (sm -> md) and (md -> lg).

templates/catalog/_partials/products-top.tpl
NeOMakinG: > Hi, I followed the shopingcart modal example, maybe we could create a helper class for modals below the 992px viewport so we don't have to overwrite every modal that has too much content, something similar to BS class `modal-fullscreen-lg-down` but only with maximum width with padding. WDYT @Hlavtox @NeOMakinG @mparvazi

If we have multiple modals with the same behavior, a helper class could be better than duplicating media queries of course!

NeOMakinG (approved):
Pre-stamping 🍬 

We've to think about a helper class in case we need this behavior in multiple places

mparvazi:
@NeOMakinG 

I've tested, works fine!

NeOMakinG:
Thanks @JBLach

PR #357, opened 2022-07-25
account-page-back-button-fix

### LINKED ISSUE #340: Viewport check: 320px

JBLach:
| Questions         | Answers
| Description?      | Account back button correction, by issue #340
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | Fixes #340

## Before
image

## After
image

### DIFF
--- a/src/scss/custom/components/customer/_customer-menu.scss
+++ b/src/scss/custom/components/customer/_customer-menu.scss
@@ -50,6 +50,6 @@ $component-name: account-menu;
 
   &__back {
     display: block;
-    margin-bottom: 1.2rem;
+    margin-bottom: 1rem;
   }
 }
--- a/templates/components/account-menu.tpl
+++ b/templates/components/account-menu.tpl
@@ -78,9 +78,3 @@
     </a>
   </div>
 {/block}
-
-{block name='account_link'}
-  <a class="btn btn-outline-primary hidden-on-desktop {$componentName}__back" href="{$urls.pages.my_account}">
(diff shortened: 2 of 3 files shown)


Hlavtox:
I liked the current implementation better 🤔

mparvazi:
> I liked the current implementation more 🤔

In current implementation the back button is too bold (I mean it's not an important action button and it's a duplication for breadcrumb link).
In other pages we used unstyle button for back links, so somehow is related to consistency in design.

NeOMakinG (approved):
Stamping because of the consistency, agree with both

I would call @PrestaShop/ux-team but looks like we won't get any feedbacks for years

Hlavtox:
> In other pages we used unstyle button for back links, so somehow is related to consistency in design.

OK ;-)

mparvazi:
Thanks @JBLach

Issue #129, opened 2021-12-01
Module CSS should not be included in the base CSS file


Hlavtox:
Module styling should be in separate styles, so they are loaded only when needed. This is how it was done in Prestashop 1.6.

People usually disable many of the default modules, their styling should not be loaded with theme.css.

CCC feature will take care of merging it into 1 file.

rodriciru:
1 filer per module or 1 file for all modules?

NeOMakinG:
Little question, is there a something to check if a module is installed? So we can disable the stylesheet include if the module isn't enabled (because if peoples uninstall a module or disable it, I would like to avoid them to need to remove the stylesheet import from the smarty markup)

NeOMakinG:
And we need to find a way to avoid having a lot of stylesheet link request I think so

Hlavtox:
The stylesheet from `modules/modulename/xxxx.css` or overriden by` theme-refacto/modules/modulename/xxxx.css` will be only loaded, if the module is active and includes it.

SharakPL:
> Little question, is there a something to check if a module is installed? So we can disable the stylesheet include if the module isn't enabled (because if peoples uninstall a module or disable it, I would like to avoid them to need to remove the stylesheet import from the smarty markup)

Yeah, this is covered:
https://github.com/PrestaShop/ps_imageslider/blob/8fcc41e8297a40541255c6bb1882baa9eb8527e0/ps_imageslider.php#L536-L541

PR #359, opened 2022-08-04
Split JS/CSS bundles

NeOMakinG:
| Questions         | Answers
| Description?      | We want to split our bundles to avoid loading unused code on different pages
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/129.
| How to test?      | The theme should look as usual
| Possible impacts? | Style and JS


SharakPL:
What about CCC? Won't it still make one big css and js file from all this?

Hlavtox:
`###         path: assets/css/modules/shoppingcart.css`

This has no effect and we are where we started. The css file must be in the module folder overriding the original CSS, so it's loaded only when the module says so.

TheGera (approved):

NeOMakinG:
> What about CCC? Won't it still make one big css and js file from all this?

I guess I tested it without CCC activated, or I missed something, but you should be right

NeOMakinG:
> `### path: assets/css/modules/shoppingcart.css`
> 
> This has no effect and we are where we started. The css file must be in the module folder overriding the original CSS, so it's loaded only when the module says so.

Yeah but then tell me the solution, because some modules doesn't have any CSS stylesheets, what about them? 😢

PR #360, opened 2022-08-08
Psgdpr module checkbox fix

JBLach:
| Questions         | Answers
| Description?      | I have fixed psgdpr module checkbox
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no

## Before
image

## After
image
image

### DIFF
--- a/modules/psgdpr/views/templates/hook/displayGDPRConsent.tpl
+++ b/modules/psgdpr/views/templates/hook/displayGDPRConsent.tpl
@@ -16,68 +16,14 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  * International Registered Trademark & Property of PrestaShop SA
  *}
-
-<div id="gdpr_consent" class="mt-2 gdpr_module_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
-  <span class="form-check">
-    <input id="psgdpr_consent_checkbox_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}" name="psgdpr_consent_checkbox" type="checkbox" value="1" class="form-check-input psgdpr_consent_checkboxes_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
-    <label class="form-check-label psgdpr_consent_message">
-      {$psgdpr_consent_message nofilter}{* html data *}
-    </label>
-  </span>
-</div>
-{literal}
-  <script type="text/javascript">
-    var psgdpr_front_controller = "{/literal}{$psgdpr_front_controller|escape:'htmlall':'UTF-8'}{literal}";
-    psgdpr_front_controller = psgdpr_front_controller.replace(/\amp;/g,'');
-    var psgdpr_id_customer = "{/literal}{$psgdpr_id_customer|escape:'htmlall':'UTF-8'}{literal}";
-    var psgdpr_customer_token = "{/literal}{$psgdpr_customer_token|escape:'htmlall':'UTF-8'}{literal}";
-    var psgdpr_id_guest = "{/literal}{$psgdpr_id_guest|escape:'htmlall':'UTF-8'}{literal}";
-    var psgdpr_guest_token = "{/literal}{$psgdpr_guest_token|escape:'htmlall':'UTF-8'}{literal}";
-
-    document.addEventListener('DOMContentLoaded', function() {
-      let psgdpr_id_module = "{/literal}{$psgdpr_id_module|escape:'htmlall':'UTF-8'}{literal}";
-      let parentForm = $('.gdpr_module_' + psgdpr_id_module).closest('form');
-
-      let toggleFormActive = function() {
-        let parentForm = $('.gdpr_module_' + psgdpr_id_module).closest('form');
-        let checkbox = $('#psgdpr_consent_checkbox_' + psgdpr_id_module);
-        let element = $('.gdpr_module_' + psgdpr_id_module);
-        let iLoopLimit = 0;
-
-        // by default forms submit will be disabled, only will enable if agreement checkbox is checked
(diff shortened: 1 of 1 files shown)


NeOMakinG:
Thanks @JBLach you're on fire these days 🔥 

I have a question, do you see any way to avoid that override?

JBLach:
> I have a question, do you see any way to avoid that override?

Hi @NeOMakinG
Does this override cause any problems? Alternatively we can overwrite the front.js file and add these classes using js, but i don't think it's better solution, maybe someone else has an idea?

JBLach:
> > > I have a question, do you see any way to avoid that override?
> > 
> > 
> > Hi @NeOMakinG Does this override cause any problems? Alternatively we can overwrite the front.js file and add these classes using js, but i don't think it's better solution, maybe someone else has an idea?
> 
> Adding an override means that if the module get updated, this override will probably not be updated in the future, it's always a bad idea in case we can avoid it, this is why I asked

Sure, i understand. I will think about a solution tomorrow

JBLach:
Hi @NeOMakinG 

We can try to create a `block` in main module and extends the file, only changing the content in the block. WDYT?
```smarty
{extends file="modules/psgdpr/views/templates/hook/displayGDPRConsent.tpl"}
{block name='gdpr_checkbox'}
  <div id="gdpr_consent" class="mt-2 gdpr_module_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
    <span class="form-check">
      <input id="psgdpr_consent_checkbox_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}" name="psgdpr_consent_checkbox" type="checkbox" value="1" class="form-check-input psgdpr_consent_checkboxes_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
      <label class="form-check-label psgdpr_consent_message">
        {$psgdpr_consent_message nofilter}{* html data *}
      </label>
    </span>
  </div>
{/block}
```

NeOMakinG:
> Hi @NeOMakinG
> 
> We can try to create a `block` in main module and extends the file, only changing the content in the block. WDYT?
> 
> ```
> {extends file="modules/psgdpr/views/templates/hook/displayGDPRConsent.tpl"}
> {block name='gdpr_checkbox'}
>   <div id="gdpr_consent" class="mt-2 gdpr_module_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
>     <span class="form-check">
>       <input id="psgdpr_consent_checkbox_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}" name="psgdpr_consent_checkbox" type="checkbox" value="1" class="form-check-input psgdpr_consent_checkboxes_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
>       <label class="form-check-label psgdpr_consent_message">
>         {$psgdpr_consent_message nofilter}{* html data *}
>       </label>
>     </span>
>   </div>
> {/block}
> ```

Does it work if the template comes from the module directly?

I guess it would be cleaner than overriding the whole file yes! I would agree.

@Hlavtox wdyt?

JBLach:
> > Hi @NeOMakinG
> > We can try to create a `block` in main module and extends the file, only changing the content in the block. WDYT?
> > ```
> > {extends file="modules/psgdpr/views/templates/hook/displayGDPRConsent.tpl"}
> > {block name='gdpr_checkbox'}
> >   <div id="gdpr_consent" class="mt-2 gdpr_module_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
> >     <span class="form-check">
> >       <input id="psgdpr_consent_checkbox_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}" name="psgdpr_consent_checkbox" type="checkbox" value="1" class="form-check-input psgdpr_consent_checkboxes_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
> >       <label class="form-check-label psgdpr_consent_message">
> >         {$psgdpr_consent_message nofilter}{* html data *}
> >       </label>
> >     </span>
> >   </div>
> > {/block}
> > ```
> 
> Does it work if the template comes from the module directly?
> 
> I guess it would be cleaner than overriding the whole file yes! I would agree.
> 
> @Hlavtox wdyt?

I tested this solution and it works

NeOMakinG:
> > > Hi @NeOMakinG
> > > We can try to create a `block` in main module and extends the file, only changing the content in the block. WDYT?
> > > ```
> > > {extends file="modules/psgdpr/views/templates/hook/displayGDPRConsent.tpl"}
> > > {block name='gdpr_checkbox'}
> > >   <div id="gdpr_consent" class="mt-2 gdpr_module_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
> > >     <span class="form-check">
> > >       <input id="psgdpr_consent_checkbox_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}" name="psgdpr_consent_checkbox" type="checkbox" value="1" class="form-check-input psgdpr_consent_checkboxes_{$psgdpr_id_module|escape:'htmlall':'UTF-8'}">
> > >       <label class="form-check-label psgdpr_consent_message">
> > >         {$psgdpr_consent_message nofilter}{* html data *}
> > >       </label>
> > >     </span>
> > >   </div>
> > > {/block}
> > > ```
> > 
> > 
> > Does it work if the template comes from the module directly?
> > I guess it would be cleaner than overriding the whole file yes! I would agree.
> > @Hlavtox wdyt?
> 
> I tested this solution and it works

Go for it! Seems way cleaner :)

NeOMakinG (approved):
So quick 🍬

NeOMakinG:
I guess we need towait for psgdpr to be released before merging this

NeOMakinG (changes requested):
Waiting for psgdpr release

kpodemski (approved):
gdpr module has been released, we can merge this :)

PR #362, opened 2022-08-12
Fix import overrides scss file order according to boostrap doc

ga-devfront:
| Questions         | Answers
| Description?      | Change import order of `overrides.scss` according to [bootstrap documentation](https://getbootstrap.com/docs/5.1/customize/sass/#importing).
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | No ticket.
| How to test?      | Check the diff.
| Possible impacts? | Possibility to use manipulation colors, SVGs, calc, etc in overrides.

### DIFF
--- a/src/scss/abstract/variables/_index.scss
+++ b/src/scss/abstract/variables/_index.scss
@@ -1,6 +1,7 @@
+@import "~bootstrap/scss/functions";
+
 @import "overrides";
 
-@import "~bootstrap/scss/functions";
 @import "~bootstrap/scss/variables";
 @import "~bootstrap/scss/mixins";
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):

mparvazi (approved):
I think it doesn't make any problem, because there is not predefined vars in `functions.scss`

NeOMakinG:
> I think it doesn't make any problem, because there is not predefined vars in `functions.scss`

This is what I was initially thinking, but let's follow the doc anyway 👍

NeOMakinG:
Thanks @ga-devfront

PR #364, opened 2022-08-17
order return table fix

JBLach:
| Questions         | Answers
| Description?      | I have fixed return table on return details page
| BC breaks?        |  no
| Deprecations?     |  no
| How to test?      | See screenshots below.

## Before
image
image

## After
image
image

## Screenshot update
image
image
image

### DIFF
--- a/templates/customer/order-return.tpl
+++ b/templates/customer/order-return.tpl
@@ -37,38 +37,69 @@
           }
         </p>
         <p>{l s='List of items to be returned:' d='Shop.Theme.Customeraccount'}</p>
-        <table class="table table-striped table-bordered">
-          <thead class="thead-default">
-            <tr>
-              <th>{l s='Product' d='Shop.Theme.Catalog'}</th>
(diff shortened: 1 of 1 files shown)


SharakPL:
What does it fix exactly? TBH I like when attributes are always visible. I would maybe change the placement (a list under the title would be better).

JBLach:
> What does it fix exactly? TBH I like when attributes are always visible. I would maybe change the placement (a list under the title would be better).

i improved the look of the table like we have on other user pages, i can see that i forgot to add the reference code on mobile view, I made a screenshot of the products without attributes, so you can't see them , if exist should be visible next to the product name as before

JBLach:
@kpodemski yes i will add what i forgot and change the position of the table as suggested by SharakPL, I'll take care of it now

NeOMakinG (approved):
🍬

kpodemski (approved):

PR #373, opened 2022-09-20
Upgrade autoprefixer to latest version

lramos-prolaser:
I was having a `Term expected` error in compiled `assets/css/theme.css`. It was some `ms-grid-` prefix added by `autoprefixer`. So, I upgraded autoprefixer to the latest version and error is gone.


SharakPL (approved):
We should probably upgrade other packages too...

NeOMakinG (approved):

NeOMakinG:
Thanks @lramos-prolaser 

@SharakPL You're right, what about adding dependabot too?

PR #383, opened 2022-10-27
Fix product page overflow in mobile view

### LINKED ISSUE #382: Product page overflow in mobile view
In `product.tpl` there is a `g-5` class in `<div class="row g-5 product js-product-container">`  that causes a slight overflow in mobile view:
image

Changing it to `g-lg-5` fixes that:
image

lucrp:
| Questions         | Answers
| Description?      | In `product.tpl` there is a `g-5` class in `<div class="row g-5 product js-product-container">` that causes a slight overflow. Changing it to `g-lg-5` fixes that.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #382
| How to test?      | Mobile view in product page 

### DIFF
--- a/templates/catalog/product.tpl
+++ b/templates/catalog/product.tpl
@@ -29,7 +29,7 @@
 
 {block name='content'}
   {* FIRST PART - PHOTO, NAME, PRICES, ADD TO CART*}
-  <div class="row g-5 product js-product-container">
+  <div class="row g-lg-5 product js-product-container">
     <div class="product__left col-12 col-lg-6 col-xl-7">
       {block name='product_cover_thumbnails'}
         {include file='catalog/_partials/product-cover-thumbnails.tpl'}
--- a/templates/catalog/product.tpl
+++ b/templates/catalog/product.tpl
@@ -29,14 +29,14 @@
 
 {block name='content'}
   {* FIRST PART - PHOTO, NAME, PRICES, ADD TO CART*}
-  <div class="row g-lg-5 product js-product-container">
-    <div class="product__left col-lg-6 col-xl-7">
+  <div class="row g-5 product js-product-container">
+    <div class="product__left col-12 col-lg-6 col-xl-7">
       {block name='product_cover_thumbnails'}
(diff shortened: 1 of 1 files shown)


Hlavtox:
Hi @lucrp, can you post a screenshot of the overflow issue please? 🤔

Hlavtox:
This is weird, I just built fresh develop branch of the theme and I have no issue like this. The gap between columns should not affect the row width at all.

Are you sure you didn't modify anything?

lucrp:
Hi @Hlavtox !
I tested again with a fresh develop branch install and I have the same overflow (both in Chrome and Firefox).
image

I found this problem with a little JS snippet that I use to find overflows like this, if you want to test it:
```javascript
document.querySelectorAll('*').forEach(el => {
  if (el.offsetWidth > document.documentElement.offsetWidth) {
      console.log('Found the worst element ever: ', el);
  }
});
````

lucrp:
I found another way of solving that. If we add a `col-12` in `product__left` and `col` in `product__col` overflow is fixed. Wdyt @Hlavtox ?

NeOMakinG:
@lucrp we did investigate it, the col fix isn't the right one at all

The parent container has an inferior gutter than the row of these columns, I guess the right fix is probably about reducing the gutters on mobile or both desktop and mobile, we should really take care about the fact that he gutter helper is fine with the parent container

lucrp:
So, if I use `p-4` or lower, it fixes the overflow. Am I right ? Otherwise, I did not understand how to change that. If you can explain me @Hlavtox, I'll appreciate it. Thanks in advance. 😄

NeOMakinG:
> So, if I use `p-4` or lower, it fixes the overflow. Am I right ? Otherwise, I did not understand how to change that. If you can explain me @Hlavtox, I'll appreciate it. Thanks in advance. 😄

`g-4 g-xl-5` should fix it properly :)

NeOMakinG (approved):

NeOMakinG:
@Hlavtox feel free to merge if you feel it's good, should do the job!

atomiix (approved):

PR #394, opened 2022-11-22
Put products page thumbnails on the left

NeOMakinG:
| Questions         | Answers
| Description?      | It's scrollable and scrolling in case the element is out of the scrollable div
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | Go on a product page with a ton of thumbnails
| Possible impacts? | Product page images

image

### DIFF
--- a/src/js/constants/selectors-map.ts
+++ b/src/js/constants/selectors-map.ts
@@ -145,8 +145,8 @@ const selectorsMap = {
     carousel: '.js-product-carousel',
     miniature: '.js-product-miniature',
     thumbnail: '.js-thumb-container',
-    thumbnailsContainer: '.js-thumbnails-container',
-    activeThumbail: (id: number): string => `.js-thumb-container:nth-child(${id + 1})`,
+    thumbsContainer: '.js-thumbnails-container',
+    activeThumbnail: (id: number): string => `.js-thumb-container:nth-child(${id + 1})`,
   },
   order: {
     returnForm: '.js-order-return-form',
--- a/src/js/constants/selectors-map.ts
+++ b/src/js/constants/selectors-map.ts
@@ -145,6 +145,7 @@ const selectorsMap = {
     carousel: '.js-product-carousel',
     miniature: '.js-product-miniature',
     thumbnail: '.js-thumb-container',
+    thumbnailsContainer: '.js-thumbnails-container',
     activeThumbail: (id: number): string => `.js-thumb-container:nth-child(${id + 1})`,
   },
   order: {
--- a/src/js/product.ts
+++ b/src/js/product.ts
@@ -20,6 +20,7 @@ export default () => {
 
   function onProductSlide(event: ProductSlideEvent): void {
     const thumbnails = document.querySelectorAll(SelectorsMap.product.thumbnail);
+    const thumbnailsContainer = document.querySelector(SelectorsMap.product.thumbnailsContainer);
 
     thumbnails.forEach((e: Element) => {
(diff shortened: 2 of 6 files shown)


src/js/product.ts
```diff
@@ -28,6 +29,36 @@ export default () => {
     const activeThumbnail = document.querySelector(SelectorsMap.product.activeThumbail(event.to));
 
     if (activeThumbnail) {
+      const activeThumbailRect = activeThumbnail.getBoundingClientRect();
```
SharakPL, line 32:
```suggestion
      const activeThumbnailRect = activeThumbnail.getBoundingClientRect();
```
This and all references below. IMO we should switch to term `thumb` instead of `thumbnail` just to shorten the code a bit.

src/scss/core/pages/product/_product.scss
```diff
@@ -1,7 +1,10 @@
 .thumbnails__container {
-  margin-top: 1rem;
+  max-height: 38.125rem;
```
SharakPL, line 2: I very much don't like thumbs on the left. It's hard to make the thumbs carousel responsive 😭
IMO we should use a component with thumbs built-in and on the top/bottom, similar to [THIS](https://fancyapps.com/playground/vu)

src/js/constants/selectors-map.ts
```diff
@@ -145,6 +145,7 @@ const selectorsMap = {
     carousel: '.js-product-carousel',
     miniature: '.js-product-miniature',
     thumbnail: '.js-thumb-container',
+    thumbnailsContainer: '.js-thumbnails-container',
     activeThumbail: (id: number): string => `.js-thumb-container:nth-child(${id + 1})`,
```
SharakPL, line 149:
```suggestion
    activeThumbnail: (id: number): string => `.js-thumb-container:nth-child(${id + 1})`,
```
Another typo.

src/js/product.ts
```diff
@@ -20,15 +20,46 @@ export default () => {
 
   function onProductSlide(event: ProductSlideEvent): void {
     const thumbnails = document.querySelectorAll(SelectorsMap.product.thumbnail);
+    const thumbContainer = document.querySelector(SelectorsMap.product.thumbContainer);
```
SharakPL, line 23: `thumbContainer` is single `thumbnail` in selectors-map. I believe this should be `thumbsContainer` although selectors-map still uses `thumbnailsContainer`

src/scss/core/pages/product/_product.scss
Hlavtox: I don't like it either :D

NeOMakinG:
@PrestaShop/product-team 2 committers prefer the thumbnails under the product image, it still simplifies the codebase, and peoples are able to change it on their personal themes following this PR or adding a new library, I've the feeling that we should close that and keep our codebase sane

eshraw:
> @PrestaShop/product-team 2 committers prefer the thumbnails under the product image, it still simplifies the codebase, and peoples are able to change it on their personal themes following this PR or adding a new library, I've the feeling that we should close that and keep our codebase sane

@NeOMakinG would it be possible to open a Gh discussion on this subject ? I think it would be interesting to discuss it further.
In the meantime I agree, let's to go with the thumbnails under the product image considering the mentioned advantages. I believe there would be an impact on the potential release date as well right ?

NeOMakinG:
Peoples gave their opinion, let's close this for the moment, let's see in the future! https://github.com/PrestaShop/hummingbird/discussions/410

PR #395, opened 2022-11-23
Fix: changing the country creates a new address

NeOMakinG:
| Questions         | Answers
| Description?      | When changing the country in the form, an ajax call is made, and it needs to include the `$id_address`.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes [#30375](https://github.com/PrestaShop/PrestaShop/issues/30375)
| How to test?      | Follow the steps to reproduce in #30375. Notice that the form action URL still contains the address id after changing the country.
| Possible impacts? | This should not have any other impact than the one described.

### DIFF
--- a/templates/checkout/_partials/address-form.tpl
+++ b/templates/checkout/_partials/address-form.tpl
@@ -13,7 +13,7 @@
       method="POST"
       action="{url entity='order' params=['id_address' => $id_address]}"
       data-id-address="{$id_address}"
-      data-refresh-url="{url entity='order' params=['ajax' => 1, 'action' => 'addressForm']}"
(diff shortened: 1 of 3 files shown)


nicosomb (approved):

0x346e3730 (approved):

matthieu-rolland (approved):

matthieu-rolland:
@NeOMakinG should we reopen the issue ? I'm setting the PR as ready for QA since your how to test filled and the issue was apparently tested

NeOMakinG:
@matthieu-rolland No, we need to original issue to be reproduced, we've to talk with the QA

lartist (approved):

nicosomb:
Thank you @NeOMakinG !

PR #403, opened 2022-12-07
Switch some images to picture tags

NeOMakinG:
| Questions         | Answers
| Description?      | We want to use picture tags everywhere we have some images and these parts were missing
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | Images on every part should work as expected
| Possible impacts? | Images display

### DIFF
--- a/modules/productcomments/views/templates/hook/post-comment-modal.tpl
+++ b/modules/productcomments/views/templates/hook/post-comment-modal.tpl
@@ -29,10 +29,71 @@
 
                 {block name='product_cover'}
                   <div class="product-cover">
-                    {if $product.cover}
-                      <img class="js-qv-product-cover rounded" src="{$product.cover.bySize.medium_default.url}" alt="{$product.cover.legend}" title="{$product.cover.legend}" style="width:100%;" itemprop="image">
+                    {if !empty($product.cover)}
+                      <picture>
+                        {if isset($product.cover.bySize.default_xs.sources.avif)}
+                          <source 
+                            srcset="
+                              {$product.cover.bySize.default_xs.sources.avif},
+                              {$product.cover.bySize.default_m.sources.avif} 2x"
+                            type="image/avif"
+                          >
+                        {/if}
+
+                        {if isset($product.cover.bySize.default_xs.sources.webp)}
+                          <source 
+                            srcset="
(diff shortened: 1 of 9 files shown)


templates/checkout/_partials/order-confirmation-table.tpl
```diff
@@ -11,9 +11,70 @@
...
+
+              <img
+                class="img-fluid"
+                srcset="
+                  {$product.default_image.bySize.default_xs.url},
+                  {$product.default_image.bySize.default_m.url} 2x"
+                loading="lazy"
+                width="{$product.default_image.bySize.default_xs.width}"
+                height="{$product.default_image.bySize.default_xs.height}"
+                {if !empty($product.default_image.legend)}
```
Hlavtox, line 41: This is handled by the core, there will always be a legend. 👍 

`'legend' => !empty($object->meta_title) ? $object->meta_title : $object->name,`

Hlavtox (approved):
Just check the legends, good otherwise!

SharakPL (changes requested):
Some indentations + legend @Hlavtox mentioned.
Also maybe we could limit the size of the tags eg. put `type` next to `source`:
```
<source type="image/avif"
  srcset="..."
>
```

NeOMakinG:
Thanks a lot for the review @SharakPL @Hlavtox , I'll fix that tomorrow

Hlavtox (approved):

Hlavtox:
@SharakPL Merge it when you check it ;-)

SharakPL (approved):

modules/productcomments/views/templates/hook/post-comment-modal.tpl
```diff
@@ -29,10 +29,67 @@
...
+                        <img
+                          class="js-qv-product-cover rounded"
+                          srcset="
+                            {$product.cover.bySize.default_xs.url},
+                            {$product.cover.bySize.default_m.url} 2x"
+                          loading="lazy"
+                          width="{$product.cover.bySize.default_xs.width}"
+                          height="{$product.cover.bySize.default_xs.height}"
+                          alt="{$product.cover.legend}"
+                          title="{$product.cover.legend}"
```
Hlavtox, line 61: @NeOMakinG maybe > missing here? :D

PR #405, opened 2022-12-14
Password policy result is async and should be handled

NeOMakinG:
| Questions         | Answers
| Description?      | We just introduced a bug on password policy as it's not async
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | Try to register a new user with a password
| Possible impacts? | password policy


kpodemski (approved):

matthieu-rolland (approved):

Hlavtox:
🚀

PR #411, opened 2022-12-19
Update minimum compatibility of the theme

kpodemski:
| Questions         | Answers
| Description?      | Drop support for 1.7
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | n/a
| How to test?      | n/a

### DIFF
--- a/config/theme.yml
+++ b/config/theme.yml
@@ -8,7 +8,7 @@ author:
 
 meta:
   compatibility:
-      from: 8.1.0
+      from: 8.0.0
       to: ~
 
   available_layouts:
--- a/config/theme.yml
+++ b/config/theme.yml
@@ -8,7 +8,7 @@ author:
 
 meta:
   compatibility:
-      from: 1.7.0.0
+      from: 8.1.0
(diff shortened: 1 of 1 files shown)


SharakPL (approved):

NeOMakinG:
@kpodemski which feature?

Hlavtox:
I will go against Krystian here. ;-) I think we should support 8.0 and newer. Because:
- Why not, if it works.
- 8.1 should is backwards compatible, the theme could be also.
- It will allow people to migrate to hummingbird before 8.1 upgrade.
- It will allow people to start building new themes based on this, which will work with 8.0 also.

______

However, if a change comes that will just require 8.1, I will bump the compatibility with no regrets.

kpodemski:
@Hlavtox 

As a theme developer, you don't put your theme for a sell for PrestaShop 8.0 if an 8.1 is available. It's not only about the theme but also about the modules you use inside.

Anyway, we can go for 8.0, important is to drop 1.7 :-)

PR #414, opened 2022-12-21
Fix QA feedbacks

### LINKED ISSUE #409: QA feedbacks - bug list
## Update 22.01.2023

**FILTERING**
- [ ] Using price filter in faceted search removes all other URL paramters other than `q`. This breaks filtering on search page for example, if your module supports it. - https://github.com/PrestaShop/hummingbird/pull/452
- [ ] Price filter slider broken due to weight slider. - https://github.com/PrestaShop/hummingbird/pull/451
- [ ] Resize screen to mobile, open filters, resize screen back to desktop. ➡️ Empty modal is open.
- [ ] Resize screen to mobile, resize it back ➡️ Price slider doesn't work.
- [ ]  Price filter - wrong format of numbers. No space between thousands and . instead of , for decimal places. OK in active filters section. - https://github.com/PrestaShop/hummingbird/pull/451
- [ ] Weight slider formatting not provided, no space between weight and unit in active filters.

**CHECKOUT**
- [ ] Finish checkout up to Payment step. Hit F5, click Shipping method step, see both steps. Active step button on top is missing `active` class on page load. - https://github.com/PrestaShop/hummingbird/pull/450
- [ ]  displayCarrierExtraContent hook is working, but is not appearing and disappearing dynamically, only on page load.

**CART**
- [ ] When you add more items than in stock from product list, they are added and no modal or error is displayed.
- [ ] When you add more items than in stock from product page, it shows `The product is no longer available in this quantity.`, but it gets added to cart.
- [ ] When you delete a product from cart, there are some extra parameters in the URL - `cart?delete=1&id_product=2&id_product_attribute=9&token=4027a89ab545500654847cf1fb45e074`
- [ ] Add more items to cart than is in stock, go to cart.
  - [ ] You see `Mug The best is yet to come is no longer available in this quantity. You cannot proceed with your order until the quantity is adjusted.`, but you don't know how many pieces are in stock.
  - [ ] You change quantity, the error disappears, only a toast is shown in top right.
  - [ ] Even though there is an error `Proceed to checkout` is active.
  - [ ] When you click it, you stay on the same page, but the red error disappears.
  - [ ] When you add another product that is in stock, go to cart, delete this instock product, the red error disappears.

**HEADER**
- [ ] Logo is shifted on homepage VS other pages.

**MOBILE**
- [ ] Question: Return Merchandise, why you didn't change the color of badge according to the status of return (why it always blue) (Response: IDK what you are talking about but colors comes from the backend so it's probably related to the core not the theme?)

**CART RULES**
- [ ] Also, when we already choose a code promo and we click on another the accordion of Promo code closes :x: (Response:  core issue, pretty sure you can reproduce it on classic)

**SPECIFIC PRICES**
- [ ] there's no background and no style to differentiate the table of specific prices

**UNFINISHED THINGS**
 - [ ] Style category left column module.
 - [ ] Style other prices like unit prices and ecotax and others.

_______________________________________

**INSTALL**
- [x] https://github.com/PrestaShop/hummingbird/issues/417

**HEADER:**
- [x] Search results dropdown not working, it's rendered, but not visible for some reason. This worked correcly before.
- [x] Menu & sub-menu: no hover action
- [x] Search: google auto-fill is working (it shouldn't work) + when the list is displayed, there's no hover action on the items (background of the element should be changed)
- [x] Language: the cursor should change as a hand, and there's no css/style for the language list

**BANNER**
- [x] image-sample 2: the text-color is grey, so it's unreadable as the background another grey (the text-color should be white) (Response: Not related the theme but default pictures of the module, we shouldn't really care)

**FOOTER**
- [x] Everything works except the psgdpr checkbow is not displayed and it's known issue [#30348](https://github.com/PrestaShop/PrestaShop/issues/30348) (not related to the theme)

**New Products page + all Products page**
- [x] What's wrong with this page? why it displays only Accessories Products (see the first attached screenshot)
@Hlavtox - There is no issue on theme side, category block displays the category of last category visited in cookie. This is an issue that must be addressed in ps_categorytree module. EDIT: It is already fixed in categorytree 2.0.3 ;-)

- [x] Question: shouldn't all filters be open at first? :sweat_smile: (Response: ISO Behavior of classic is closed, and opened if it's an active filter)
- [x] About category-tree__item like Men and Women under Clothes, shouldn't be with other style? :sweat_smile:
       For ex (see the second attached screenshot), I used (Response: We followed the mockups, nothing to do there):
   font-weight: 400;
   padding-left: 5px;
   
- [x] When I click on wishlist-button-add, nothing happen, it should display a connexion modal (when you're not connected) and it should display the modal Add to my favorites when the user is connected **(Response: Not related to the theme, you need to reset the module and everything will be fine, this is a known bug when you switch from one theme to another or install it the first time)**

**CHECKOUT**
- [x] It's not possible to finish second step in checkout. You click save address, page reloads and nothing happens. This worked correcly before. - https://github.com/PrestaShop/hummingbird/pull/445
- [x] While creating a new account, there's some HTML tags are displayed (known issue [#29940](https://github.com/PrestaShop/PrestaShop/issues/29940)) **(Response: not related to the theme I think)**
- [x] Why I found Mrs. always selected by default, even thought I choose Mr. but I made a mistake in password, the page reloads and display Mrs. checked :exploding_head: (even in Information page, I get the same behavior, I already register the customer as Mr.
- [x] Click on show details, you will see that the product price is written twice as there's a discount on it but there's no discount :open_mouth: (see the third attached screenshot)
- [x] when I click on Customized button, I encounter some weird behavior, when I close the modal, the button status changes to focus (see the attached screen record) (Response: BS5 has the same behavior: https://getbootstrap.com/docs/5.0/components/modal/#live-demo)

**ORDER HISTORY**
- [x] As a customer, I send a message about a product, the message is well displayed :white_check_mark: but there's no space between the Messages and Add a message blocks, you need to add <hr> in the div.box.messages
- [x] The message block is a mess, you need to work on it (see the attached screenshot n°5)

**ADDRESSES**
- [x] It's not possible to create a new address. You save the form, `Please fix the error below.` appears, but there are no errors and the form is correctly filled out. This worked correcly before. - https://github.com/PrestaShop/hummingbird/pull/445
- [x] where are the error messages? (Response: They are not displayed because we use bootstrap validation and most of the time it doesn't work because its out of the viewport so the message disappear, I feel like it's better as it is right now but we might discuss it with the product team, anyway we can't do anything really interesting there :( )

**CART RULES**
- [x] This error is not displayed when we use the hummingbird theme

** MOBILE **
- [x] Except in the checkout, these two blocks (reassurrance & order_summary) should be dispplayed under the checkout process block :wink:
- [x] that banner buttons arrow_left and arrow_right are too big :x:

**CONTACT US**
- [x] the blocks Email us and Address are not displayed :x:

NeOMakinG:
| Questions         | Answers
| Description?      | The QA did some feedbacks that we need to tackle
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #409 .
| How to test?      | See QA feedbacks list of the issue and test it
| Possible impacts? | Global

### DIFF
--- a/templates/_partials/form-fields.tpl
+++ b/templates/_partials/form-fields.tpl
@@ -57,7 +57,6 @@
                 name="{$field.name}"
                 id="{$field.name}_{$value}"
                 value="{$value}"
-                checked
                 {if $field.required}required{/if}
                 {if $value eq $field.value} checked {/if}
               >
--- a/templates/_partials/form-fields.tpl
+++ b/templates/_partials/form-fields.tpl
@@ -183,9 +183,7 @@
 
     {block name='form_field_comment'}
       {if (!$field.required && !in_array($field.type, ['radio-buttons', 'checkbox']))}
-        <div class="form-text">
-          {l s='Optional' d='Shop.Forms.Labels'}
-        </div>
+        <div class="form-text">{l s='Optional' d='Shop.Forms.Labels'}</div>
       {/if}
     {/block}
   </div>
--- a/templates/_partials/form-fields.tpl
+++ b/templates/_partials/form-fields.tpl
@@ -10,9 +10,9 @@
 
 {else}
 
-  <div class="mb-3 {if !empty($field.errors)}has-error{/if}">
+  <div class="mb-3{if !empty($field.errors)} has-error{/if}">
     {if ($field.type !== 'checkbox')}
-      <label class="form-label{if $field.required} required{/if}">
+      <label class="form-label{if $field.required} required{/if}" for="{$field.name}">
         {if $field.type !== 'checkbox'}
           {$field.label}
         {/if}
@@ -59,7 +59,7 @@
                 value="{$value}"
                 {if $field.required}required{/if}
                 {if $value eq $field.value} checked {/if}
-              >
+              />
               <label class="form-check-label" for="{$field.name}_{$value}">
                 {$label}
               </label>
@@ -80,7 +80,7 @@
             id="{$field.name}"
             value="1" {if $field.value}checked="checked"{/if}
             {if $field.required}required{/if}
-          >
+          />
           <label class="form-check-label" for="{$field.name}">
             {$field.label nofilter}
           </label>
@@ -90,7 +90,7 @@
     {elseif $field.type === 'date'}
 
       {block name='form_field_item_date'}
-        <input name="{$field.name}" class="form-control" type="date" value="{$field.value}"{if isset($field.availableValues.placeholder)} placeholder="{$field.availableValues.placeholder}" aria-label="{$field.availableValues.placeholder}"{/if}>
+        <input name="{$field.name}" class="form-control" type="date" value="{$field.value}"{if isset($field.availableValues.placeholder)} placeholder="{$field.availableValues.placeholder}" aria-label="{$field.availableValues.placeholder}"{/if} />
         {if isset($field.availableValues.comment)}
           <span class="form-text">
             {$field.availableValues.comment}
@@ -128,6 +128,7 @@
           <input
             class="form-control js-child-focus js-visible-password"
             name="{$field.name}"
+            id="{$field.name}"
             type="password"
             {if $field.autocomplete}autocomplete="{$field.autocomplete}"{/if}
             value=""
@@ -138,7 +139,7 @@
             {if isset($configuration.password_policy.minimum_score)}data-minscore="{$configuration.password_policy.minimum_score}"{/if}
             data-bs-placement="top"
             data-bs-trigger="manual"
-          >
+          />
 
           <button
             class="btn btn-primary"
@@ -158,14 +159,15 @@
         <input
           class="form-control"
           name="{$field.name}"
+          id="{$field.name}"
           type="{$field.type}"
           value="{$field.value}"
           {if $field.autocomplete}autocomplete="{$field.autocomplete}"{/if}
           {if isset($field.availableValues.placeholder)}placeholder="{$field.availableValues.placeholder}"{/if}
           {if $field.maxLength}maxlength="{$field.maxLength}"{/if}
-          {if $field.required}required{/if}
           aria-label="{$field.name}"
-        >
+          {if $field.required}required{/if}
+        />
         {if isset($field.availableValues.comment)}
           <span class="form-text">
             {$field.availableValues.comment}
@@ -179,13 +181,12 @@
       {include file='_partials/form-errors.tpl' errors=$field.errors}
     {/block}
 
-    <div class="form-text">
-      {block name='form_field_comment'}
-        {if (!$field.required && !in_array($field.type, ['radio-buttons', 'checkbox']))}
-         {l s='Optional' d='Shop.Forms.Labels'}
-        {/if}
-      {/block}
-    </div>
+    {block name='form_field_comment'}
+      {if (!$field.required && !in_array($field.type, ['radio-buttons', 'checkbox']))}
+        <div class="form-text">
+          {l s='Optional' d='Shop.Forms.Labels'}
+        </div>
+      {/if}
+    {/block}
   </div>
-
 {/if}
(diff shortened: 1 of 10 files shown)


SharakPL:
Do we need XHTML compatibility?
I planned to get rid off ` />` from all `img` `input` `br` `hr` etc.

NeOMakinG:
> Do we need XHTML compatibility? I planned to get rid off ` />` from all `img` `input` `br` `hr` etc.

Well it was to check for a browser validation bug but we can just kick them out

SharakPL:
We could use a smarty linter/fixer to enforce consistency through all themes and modules: remove useless `/>`, tags' attributes inline or one per row, etc. But honestly I don't know anything useful for that. I use VSCode and Smarty extension as a formatter, but it's still not perfect.

NeOMakinG:
> We could use a smarty linter/fixer to enforce consistency through all themes and modules: remove useless `/>`, tags' attributes inline or one per row, etc. But honestly I don't know anything useful for that. I use VSCode and Smarty extension as a formatter, but it's still not perfect.

Yeah, I did try to find one years ago but unfortunately, it doesn't seem to exist at all 😢

templates/_partials/form-fields.tpl
```diff
@@ -129,6 +128,7 @@
           <input
             class="form-control js-child-focus js-visible-password"
             name="{$field.name}"
+            id="{$field.name}"
```
SharakPL, line 131: IMO it's too generic. We're talking about something like `id="phone"`. Someone may use that in other context (eg. in contact information) and add special style to `#phone` which will also impact this form field.

`#field-firstname` `#field-phone` `#field-gdpr` etc. should be ok.

templates/_partials/form-fields.tpl
```diff
@@ -180,13 +181,12 @@
...
-        {if (!$field.required && !in_array($field.type, ['radio-buttons', 'checkbox']))}
-         {l s='Optional' d='Shop.Forms.Labels'}
-        {/if}
-      {/block}
-    </div>
+    {block name='form_field_comment'}
+      {if (!$field.required && !in_array($field.type, ['radio-buttons', 'checkbox']))}
+        <div class="form-text">
+          {l s='Optional' d='Shop.Forms.Labels'}
+        </div>
```
SharakPL, line 188:
```suggestion
        <div class="form-text">{l s='Optional' d='Shop.Forms.Labels'}</div>
```
😉

templates/customer/_partials/order-messages.tpl
```diff
@@ -4,17 +4,19 @@
...
-          </div>
+        <div class="customer__message border rounded p-3 my-2">
+          <div class="customer__message__content row">
+            <div class="col-sm-4 mb-2 mb-sm-0">
+              <p class="fw-bold mb-0">{$message.name}</p>
+              <p class="fs-6 mb-0">{$message.message_date}</p>
+            </div>
+            <div class="col-sm-8">
+              {$message.message nofilter}
+            </div>
```
SharakPL, line 18:
```suggestion
            <div class="col-sm-8">{$message.message nofilter}</div>
```
😉

SharakPL (changes requested):

templates/_partials/form-fields.tpl
kpodemski: I agree, and it is like that in `classic-theme`, https://github.com/PrestaShop/classic-theme/blob/develop/templates/_partials/form-fields.tpl#L34

templates/checkout/_partials/cart-voucher.tpl
```diff
@@ -45,7 +45,7 @@
               {/block}
 
               {block name='cart_voucher_notifications'}
-                <div class="alert alert-danger js-error d-none" role="alert">
+                <div class="alert alert-danger js-error mt-2" role="alert" style="display: none;">
```
kpodemski, line 48: Inline style? 🤔

templates/_partials/form-fields.tpl
NeOMakinG: You're right!

templates/checkout/_partials/cart-voucher.tpl
NeOMakinG: @kpodemski this is a core limitation from the `core.js` file toggling this error when the voucher trigger it

SharakPL (approved):

Issue #252, opened 2022-05-02
Optimize bundles sizes, add csso, fix lighthouse issues


NeOMakinG:

Hlavtox:
We need to divide the assets per-module again, so they are not loaded in the bundle if not needed.

NeOMakinG:
> We need to divide the assets per-module again, so they are not loaded in the bundle if not needed.

Seems impossible in the current state of PrestaShop

Hlavtox:
Yes, but at least the module CSS will be loaded only on pages where it's needed, not everywhere.

Hlavtox:
With the way hummingbird is written now, all JS, all CSS, for all modules (even uninstalled) are loaded absolutely everywhere.

NeOMakinG:
> With the way hummingbird is written now, all JS, all CSS, for all modules (even uninstalled) are loaded absolutely everywhere.

Yeah I know, we can improve it and it's almost ready: https://github.com/NeOMakinG/theme-refacto/commit/d570e8b09f0147bfe9fbf35d0b26372fe8373fe2

But we can't do it perfectly for sure :(

NeOMakinG:
@Oksydan wants to take this subject, I'll move it to WIP

Oksydan:
Assign me 👀

PR #416, opened 2022-12-21
Header CLS fixes, reduced styles and classes complexity

Oksydan:
| Questions         | Answers
| Description?      | Reducing complexity and CLS caused by responsive component, changed BEM classes to stay one level deep instead of `BLOCK__ELEMEMT__ELEMENT` to just `BLOCK_EMLEMENT`. 
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/252.
| How to test?      | play around
| Possible impacts? | no

It is first part of CLS fixes for header. Also we are reducing some complexity in styles and I would like to relay more on bootstrap components instead of reinventing the wheel with some extra unneeded styles. 
It will not resolve whole issue but just some part of it. It will be easier to review code, provide feedback and explain changes if we split it into multiple PR-s.

We are right now showing preview of mobile block for the js loading time, there is no longer jumping stuff around on loading on mobile devices. We will be also making preload components in next stage of CLS fixes after this one will be reviewed and merged. 

### Comparison (Throttling 3G FAST):

**BEFORE:**

https://user-images.githubusercontent.com/25306716/210451119-9b552c1e-0243-4073-8d73-bcb98dcd6ffb.mov

**AFTER:**

https://user-images.githubusercontent.com/25306716/210451147-1785fa67-10d2-48a8-a9c3-ab1ac6c3f711.mov

There is multiple changes to header structure and we could go even further with that `header-top__right` element and children inside but I didn't want to make this PR too big. I will be harder to review and make changes according to review.

`.hidden-on-mobile` and `.hidden-on-desktop` classes has been removed and blocks got proper class replacement for that. 

One extra note: we shouldn't make more elements than just one in our BEM classes example:
`.header__top__right` - we got here way to deep, this class has been replaced by `.header-top__right`. 
There is a lot of example when BEM can goes wrong https://scalablecss.com/bem-nesting-grandchild-elements/


modules/ps_searchbar/ps_searchbar.tpl
```diff
@@ -23,7 +23,7 @@
  * International Registered Trademark & Property of PrestaShop SA
  *}
 
-<div id="_desktop_search" class="order-2 ms-auto">
+<div id="_desktop_search" class="order-2 ms-auto col-auto px-1 hidden-on-mobile">
```
NeOMakinG, line 26: We wanted to get rid of these `hidden-on-*` classes but seems like we forgot to remove them, to rely more on bootstrap classes as you specified in the description

NeOMakinG (commented):
Just a small comment but not blocking as it's not really related, as it's a WIP just ping me when it's ready for review 🚀

NeOMakinG:
French-Croissants-SM-2363

And here is a croissant as a gift to make you motivated 🥐 

😄

Oksydan:
@NeOMakinG ready for CR

src/scss/custom/components/misc/_header-block.scss
```diff
@@ -0,0 +1,70 @@
...
+    color: var(--header-top-color);
+    border-radius: $btn-border-radius;
+
+    &:hover,
+    &:focus {
+      color: var(--header-top-color);
+    }
+
+    @include media-breakpoint-up(md) {
+      padding: $btn-padding-y $btn-padding-x;
```
NeOMakinG, line 20: I guess we don't have any `CSS variables` to consume there?

NeOMakinG (dismissed):
Seems ok to me, except a little nitpick comment but 🍬

src/scss/custom/components/misc/_header-block.scss
Oksydan: Thanks @NeOMakinG for your review, changes applied.

NeOMakinG (dismissed):

Hlavtox:
@Oksydan Igor could you please resolve the conflicts? I will merge it after 👍

Oksydan:
@Hlavtox brand rebased

Hlavtox (approved):

Hlavtox:
🚀🚀🚀

Oksydan:
Whoops @Hlavtox. No QA :trollface:

PR #418, opened 2023-01-02
Add an add to cart button to the product list

NeOMakinG:
| Questions         | Answers
| Description?      | We want to add an add to cart button to the product list, if child themes wants to remove it, they can override `product.tpl` and remove that markup
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | Go on product lists and try to add some products from there
| Possible impacts? | Product list

image

### DIFF
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -167,6 +167,11 @@
                 {/if}
               {/block}
             </div>
+            {if $product.attributes|@count}
+              <a href="{$product.url}" class="btn btn-outline-primary mt-3">
+                {l s='Choose variant' d='Shop.Theme.Actions'}
+              </a>
+            {/if}
 
             {if $product.add_to_cart_url && !$product.is_customizable && !$product.attributes|@count}
               <form action="{$urls.pages.cart}" method="post" class="d-flex align-items-center mt-3">
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -182,7 +182,10 @@
                     marginHelper="mb-0"
                   }
                 </div>
-                <button data-button-action="add-to-cart" class="btn btn-primary ms-3"><i class="material-icons">&#xe854;</i></button>
+                <button data-button-action="add-to-cart" class="btn btn-primary ms-3">
+                  <i class="material-icons">&#xe854;</i>
+                  <span class="visually-hidden">{l s='Add to cart' d='Shop.Theme.Actions'}</span>
+                </button>
               </form>
             {/if}
           </div>
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -167,13 +167,14 @@
                 {/if}
               {/block}
             </div>
+
             {if $product.attributes|@count}
               <a href="{$product.url}" class="btn btn-outline-primary mt-3">
-                {l s='Choose variant' d='Shop.Theme.Actions'}
+                {l s='See details' d='Shop.Theme.Actions'}
               </a>
             {/if}
 
-            {if $product.add_to_cart_url && !$product.is_customizable && !$product.attributes|@count}
+            {if $product.add_to_cart_url && !$product.attributes|@count}
               <form action="{$urls.pages.cart}" method="post" class="d-flex align-items-center mt-3">
                 <input type="hidden" value="{$product.id_product}" name="id_product">
                 <input type="hidden" name="token" value="{$static_token}" />
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -168,14 +168,14 @@
               {/block}
             </div>
 
-            {if $product.add_to_cart_url}
+            {if $product.add_to_cart_url && !$product.is_customizable && !$product.attributes|@count}
               <form action="{$urls.pages.cart}" method="post" class="d-flex align-items-center mt-3">
                 <input type="hidden" value="{$product.id_product}" name="id_product">
                 <input type="hidden" name="token" value="{$static_token}" />
                 <div class="quantity-button js-quantity-button">
                   {include file='components/qty-input.tpl'
                     attributes=[
-                      "id"=>"quantity_wanted",
+                      "id"=>"quantity_wanted_{$product.id_product}",
                       "value"=>"1",
                       "min"=>"{if $product.quantity_wanted}{$product.minimal_quantity}{else}1{/if}"
                     ]
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -168,6 +168,23 @@
               {/block}
             </div>
 
+            {if $product.add_to_cart_url}
+              <form action="{$urls.pages.cart}" method="post" class="d-flex align-items-center mt-3">
+                <input type="hidden" value="{$product.id_product}" name="id_product">
+                <input type="hidden" name="token" value="{$static_token}" />
+                <div class="quantity-button js-quantity-button">
+                  {include file='components/qty-input.tpl'
+                    attributes=[
+                      "id"=>"quantity_wanted",
+                      "value"=>"1",
+                      "min"=>"{if $product.quantity_wanted}{$product.minimal_quantity}{else}1{/if}"
+                    ]
+                    marginHelper="mb-0"
+                  }
+                </div>
+                <button data-button-action="add-to-cart" class="btn btn-primary ms-3"><i class="material-icons">&#xe854;</i></button>
+              </form>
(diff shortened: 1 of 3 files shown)


templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -168,6 +168,23 @@
...
+                  {include file='components/qty-input.tpl'
+                    attributes=[
+                      "id"=>"quantity_wanted",
+                      "value"=>"1",
+                      "min"=>"{if $product.quantity_wanted}{$product.minimal_quantity}{else}1{/if}"
+                    ]
+                    marginHelper="mb-0"
+                  }
+                </div>
+                <button data-button-action="add-to-cart" class="btn btn-primary ms-3"><i class="material-icons">&#xe854;</i></button>
```
SharakPL, line 185:
```suggestion
                <button data-button-action="add-to-cart" class="btn btn-primary ms-3">
                  <i class="material-icons">&#xe854;</i>
                  <span class="visually-hidden">{l s='Add to cart' d='Shop.Theme.Actions'}</span>
                </button>
```
Button needs a description for screen readers.

SharakPL (changes requested):

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -168,6 +168,26 @@
...
             </div>
 
+            {if $product.add_to_cart_url}
+              <form action="{$urls.pages.cart}" method="post" class="d-flex align-items-center mt-3">
+                <input type="hidden" value="{$product.id_product}" name="id_product">
+                <input type="hidden" name="token" value="{$static_token}" />
+                <div class="quantity-button js-quantity-button">
+                  {include file='components/qty-input.tpl'
+                    attributes=[
+                      "id"=>"quantity_wanted",
```
kpodemski, line 178: This will cause duplicated ids

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -168,6 +168,26 @@
               {/block}
             </div>
 
+            {if $product.add_to_cart_url}
+              <form action="{$urls.pages.cart}" method="post" class="d-flex align-items-center mt-3">
+                <input type="hidden" value="{$product.id_product}" name="id_product">
+                <input type="hidden" name="token" value="{$static_token}" />
```
kpodemski, line 174: This won't work for the product lists, there's a need to have this done via JS to ensure the token is correct.

CC @Oksydan

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -168,6 +168,26 @@
               {/block}
             </div>
 
+            {if $product.add_to_cart_url}
```
kpodemski, line 171: This button probably shouldn't be displayed if the product requires customization or has attributes.

Hlavtox: Yes, if product has attributes, display "Choose variant". 🙏

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -168,14 +168,14 @@
               {/block}
             </div>
 
-            {if $product.add_to_cart_url}
+            {if $product.add_to_cart_url && !$product.is_customizable && !$product.attributes|@count}
```
SharakPL, line 176:
```suggestion
            {if !$configuration.is_catalog && $product.add_to_cart_url && !$product.is_customizable && !$product.attributes|@count}
```

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -167,6 +167,11 @@
                 {/if}
               {/block}
             </div>
+            {if $product.attributes|@count}
```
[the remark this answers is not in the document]
Hlavtox: Hmmm, yes. I think we can generalize it to `"See detail"`?

templates/catalog/_partials/miniatures/product.tpl
kpodemski: checking for this is already inside add_to_cart_url

SharakPL: then maybe other conditions should be there too to avoid so many conditions in the template

SharakPL: Is this config still used in the template?
obraz
obraz

Hlavtox:
We need to make sure that it works for products with limited stock. If you do this on current Prestashop, have 1 item in stock and you add 2 pieces, you get empty modal.

NeOMakinG:
> We need to make sure that it works for products with limited stock. If you do this on current Prestashop, have 1 item in stock and you add 2 pieces, you get empty modal.

@nicosomb is taking over @mparvazi 's PR related to this

templates/catalog/_partials/miniatures/product.tpl
Oksydan: Yes, @kpodemski is right. 
Block might be cached and we should make sure that token provided to request is correct @NeOMakinG.
We could simply replace `static_token` value with `prestashop.static_token` global variable here https://github.com/PrestaShop/PrestaShop/blob/develop/themes/_core/js/cart.js#L101 with use of https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams or regex replace (might be tricky).

templates/catalog/_partials/miniatures/product.tpl
NeOMakinG: @SharakPL what's the variable exposed with that option? I can't manage to find it by myself

templates/catalog/_partials/miniatures/product.tpl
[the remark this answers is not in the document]
NeOMakinG: `$product.is_customizable` doesn't seems to work as expected on that part btw

templates/catalog/_partials/miniatures/product.tpl
NeOMakinG: @Oksydan do you think it's possible to somehow manage it on the theme side?

templates/catalog/_partials/miniatures/product.tpl
kpodemski: @NeOMakinG I don't think it is available, I think we should add it inside `FrontController::getTemplateVarConfiguration()`

SharakPL: Actually I think this is already handled too. Name is `allow_add_variant_to_cart_from_listing`

https://github.com/PrestaShop/PrestaShop/blob/c15bdae80e963714c55c4438aeb5a5adaec1e2d7/src/Adapter/Presenter/Product/ProductListingLazyArray.php#L38-L49

templates/catalog/_partials/miniatures/product.tpl
Oksydan: @NeOMakinG you can make some dirty tricks like:

```js
document.querySelectorAll('[name="token"]').forEach(el => {
  el.value = prestashop.static_token
})
```
That's major disadvantage of using core.js 😕

NeOMakinG: > @NeOMakinG you can make some dirty tricks like:
> 
> ```js
> document.querySelectorAll('[name="token"]').forEach(el => {
>   el.value = prestashop.static_token
> })
> ```
> 
> That's major disadvantage of using core.js 😕
> 
> EDIT: I just realized that you can ofc create you own function that will handle add to cart process but that's some extra code to maintain with core.js changes. Still no recommended.

I would probably prefer that dirty tricks over changing the corejs behavior on a minor version

templates/catalog/_partials/miniatures/product.tpl
NeOMakinG: Well it's probably doing the tricks, seems simpler now

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -167,13 +167,14 @@
...
+
             {if $product.attributes|@count}
               <a href="{$product.url}" class="btn btn-outline-primary mt-3">
-                {l s='Choose variant' d='Shop.Theme.Actions'}
+                {l s='See details' d='Shop.Theme.Actions'}
               </a>
             {/if}
 
-            {if $product.add_to_cart_url && !$product.is_customizable && !$product.attributes|@count}
+            {if $product.add_to_cart_url && !$product.attributes|@count}
```
SharakPL: Let's make it simpler :)
```
            {if $product.attributes|@count}
              <a href="{$product.url}" class="btn btn-outline-primary mt-3">{l s='See details' d='Shop.Theme.Actions'}</a>
            {elseif $product.add_to_cart_url}
```

kpodemski: yes, `{else}` would be enough :)

NeOMakinG: done

SharakPL (approved):

PR #423, opened 2023-01-06
Add blockwishlist to theme.yml displayFooter

jirbel:
| Questions         | Answers
| Description?      | Blockwishlist missing in displayFooter hook cause missing EventBus, so there is no Blockwishlist listener on fresh Hummingbird installation
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no

### DIFF
--- a/config/theme.yml
+++ b/config/theme.yml
@@ -85,7 +85,7 @@ global_settings:
         - ps_linklist
         - ps_customeraccountlinks
         - ps_contactinfo
-        - ~
+        - blockwishlist
       displayLeftColumn:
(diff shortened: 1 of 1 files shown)


Hlavtox (approved):

SharakPL:
About that! IMO it shouldn't be hooked in the footer at all. Module loads its modals here and footer styles may mess theme up (wrong colors on buttons and links, missing backdrop for modal, etc.).

It should be hooked just before body closing tag and outside footer, but it has to be done in the blockwishlist module first...

Issue #251, opened 2022-05-02
Think and improve accessibility


NeOMakinG:

mparvazi:
I'm ready to work on it.

mparvazi:
## Header

- [x] Text labels
- [x] Contrast
- [ ] Keyboard:
- - [x] Links & Search widget #259
- - [ ] Main menu

NeOMakinG:
@Hlavtox why did we put this issue to blocked? What are we needing for?

yannicka:
Is there an audit of the current accessibility of the theme? That would be useful to know exactly what needs to be done.

yannicka:
I'm not an accessibility expert by any means, but there still seem to be some major flaws in terms of accessibility.

For example, the desktop menu does not open from the keyboard. The same goes for the mobile menu, which you can't open from the keyboard, and you can't scroll through the sub-menus from the keyboard (no focus possible).

I can also imagine that there are a few "aria" missing here and there.

yannicka:
I'm not an accessibility expert by any means, but there still seem to be quite a few accessibility problems.

For example, in the mobile menu, it's not possible to focus on the arrows to open the sub-menus.

Also, on the desktop menu, I'm wondering about the accessibility of opening sub-menus with the arrows, but I don't know much about that.

I'd also imagine that there are a few "aria" missing here and there.

PR #433, opened 2023-01-12
Improved a11y for search form

nicosomb:
| Questions         | Answers
| Description?      | Improved a11y for search form
| Type?             | bug fix 
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 
| Possible impacts? | 

Related to https://github.com/PrestaShop/hummingbird/issues/251

### DIFF
--- a/modules/ps_searchbar/ps_searchbar.tpl
+++ b/modules/ps_searchbar/ps_searchbar.tpl
@@ -28,7 +28,7 @@
     <form method="get" action="{$search_controller_url}">
       <input type="hidden" name="controller" value="search">
       <i class="material-icons search" aria-hidden="true">search</i>
-      <input class="js-search-input" type="text" name="s" value="{$search_string}" placeholder="{l s='Search our catalog' d='Shop.Theme.Catalog'}" aria-label="{l s='Search' d='Shop.Theme.Catalog'}">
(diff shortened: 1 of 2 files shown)


0x346e3730 (approved):

kpodemski:
I think we'll need to have `$hookName` available to avoid having two elements with the same id.

Hlavtox:
We can't have that ID there if we want to use multiple search forms, we must find another solution 👍

mflasquin (commented):
If I am not mistaken, update a html element id is considered as a breaking change

Hlavtox:
@mflasquin This is not core bro, we can do whatever we want here

NeOMakinG:
> We can't have that ID there if we want to use multiple search forms, we must find another solution 👍

The theme is currently containing only one, it might not be a blocker for this PR but should be an improvement for the future

NeOMakinG (approved):
Give it a 🍬 because I feel like the ID not a theme blocker but more a core/ps_searchbar related problem or ever add `{$hookName}` to the ID to make it more unique

Hlavtox (approved):
Good for me, we can merge

kpodemski (changes requested):
@nicosomb did you test it? is `$hookName` available and it has a correct value? You have to add a `hookName` to the view.

nicosomb:
Hmm, in this module, it doesn't work.

kpodemski:
@Hlavtox I would keep it simple and add a hookName variable to the search module

PR #442, opened 2023-01-27
Update data-refresh-url in address-form.tpl 

### LINKED ISSUE #441: Can't add new adress with hummingbird theme
Problem is when write address information and continue to shipping carrier selection, it just reloads and asking the adress information again.
I tried on fresh prestashop with current develop branch.

https://user-images.githubusercontent.com/104589134/215178215-1ea1312e-b2b4-4e22-94a7-3b281960ab4d.mp4

||Red text in end saying "No suitable carrier for your location", its ok||

nizelg:
| Questions         | Answers
| Description?      | Please be specific when describing this change. What did you change? Why?
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | Fixes #441 
| Sponsor company   | 
| How to test?      |

### DIFF
--- a/templates/checkout/_partials/address-form.tpl
+++ b/templates/checkout/_partials/address-form.tpl
@@ -13,7 +13,7 @@
       method="POST"
       action="{url entity='order' params=['id_address' => $id_address]}"
       data-id-address="{$id_address}"
-      data-refresh-url="{url entity='order' params=['ajax' => 1, 'action' => 'addressForm', 'id_address' => $id_address]}"
+      data-refresh-url="{url entity='order' params=['ajax' => 1, 'action' => 'addressForm']}"
    >
 {/block}
 


nizelg:
It works in order page but still not works in account > add new address

Hlavtox:
The TPL should not be the case, it's the same on classic - https://github.com/PrestaShop/classic-theme/blob/develop/templates/checkout/_partials/address-form.tpl

nizelg:
I understand, Something is wrong about how the form response handled, I compared the response of add new adress form and on classic theme it returns 302, but while hummingbird is running response 200 and content is the error page we see...
image

nizelg:
@Hlavtox Deleted assets/js folder and yayy
I know this is not a fix but i'm sure we found where the problem is
image

nizelg:
@Hlavtox I found where the issue is! I deleted assets/js and it works now. What do you think about this?
I'm not sure i can do anything about this I don't know javascipt.
image

nizelg:
latest working release is [v0.1.1](https://github.com/PrestaShop/hummingbird/releases/tag/v0.1.1)

PR #462, opened 2023-02-21
Language selection not displaying correctly because of the currency b…

### LINKED ISSUE #459: Language selection not displaying correctly.
Left padding or margin It's too big

Screenshot_20230219_134753_Samsung Internet

JBLach:
| Questions         | Answers
| Description?      | Language selection not displaying correctly because of the currency block when it's empty
| Type?             | bug fix 
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #459
| How to test?      | Enable only one currency and few languages ​​for your shop. Screenshots below

## Before
image

## After
image

### DIFF
--- a/src/js/modules/ps_currencyselector.ts
+++ b/src/js/modules/ps_currencyselector.ts
@@ -7,6 +7,11 @@ const initCurrencySelector = () => {
   const {Theme} = window;
   const {currencySelector: CurrencySelectorMap} = Theme.selectors;
   const currencySelector = document.querySelector<HTMLElement>(CurrencySelectorMap.currencySelector);
+  const currencyMobileSelector = document.querySelector<HTMLElement>(CurrencySelectorMap.currencyMobileSelector)
+
+  if(!currencySelector) {
+    currencyMobileSelector?.remove();
+  }
 
   currencySelector?.addEventListener('change', (event) => {
     const option = event.target as HTMLOptionElement;
--- a/src/js/modules/ps_currencyselector.ts
+++ b/src/js/modules/ps_currencyselector.ts
@@ -7,9 +7,9 @@ const initCurrencySelector = () => {
   const {Theme} = window;
   const {currencySelector: CurrencySelectorMap} = Theme.selectors;
   const currencySelector = document.querySelector<HTMLElement>(CurrencySelectorMap.currencySelector);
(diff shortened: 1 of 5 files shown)


src/js/modules/ps_currencyselector.ts
```diff
@@ -7,6 +7,11 @@ const initCurrencySelector = () => {
   const {Theme} = window;
   const {currencySelector: CurrencySelectorMap} = Theme.selectors;
   const currencySelector = document.querySelector<HTMLElement>(CurrencySelectorMap.currencySelector);
+  const currencyMobileSelector = document.querySelector<HTMLElement>(CurrencySelectorMap.currencyMobileSelector);
+
+  if (!currencySelector) {
+    currencyMobileSelector?.remove();
```
matthieu-rolland, line 13: can you please explain the logic and what's happening here exactly ?

matthieu-rolland (commented):
thank you for your contribution @JBLach , I just have one question :+1:

JBLach:
> I don't think this is a right solution. `_mobile_currency_selector` and `_desktop_currency_selector` ids provide native moving of elements inside the theme (see other IDs with this prefix.
> 
> If there are some styling issues, we need to adjust the CSS targetting.

Hi @Hlavtox,
You are right. I forgot we have pseudo-class :empty that i can use in this case :)

Hlavtox:
@JBLach I will check it later, but there should be no need for this css, we shouldnt target these “switching divs” at all I think 🤔

JBLach:
@Hlavtox 
Sure, maybe it would be better to target a class col-auto?

src/js/modules/ps_currencyselector.ts
nizelg: This will surely cause content jump

kpodemski: @nizelg it's hidden under off-canvas if I'm not mistaken, shouldn't cause CLS

kpodemski:
ping @Hlavtox

kpodemski:
ping @Hlavtox

NeOMakinG (approved):
lgtm

kpodemski (approved):

kpodemski:
thanks @JBLach

PR #468, opened 2023-03-03
Fix subcategories images

### LINKED ISSUE #467: Wrong images used for subcategories
That's how it looks now:

obraz

subcategories.tpl uses `large` image which is main `category_cover` (2000x400 px). The category's thumbnail

obraz
is generated only as `small` so that's what should be used.

There are several problems with category images, but they need to be fixed in PrestaShop repo. For example:
- if only thumbnail is used then none of the images (including thumbnail) is available ($subcategory.image is null) so it can't be used even though it's set
- thumbnail image can't be deleted

SharakPL:
| Questions         | Answers
| Description?      | Use correct image as category thumbnail
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #467
| Sponsor company   | ~
| How to test?      | Open category with subcategories available and category thumbnail set.

### Before
obraz

### After 
obraz

### DIFF
--- a/templates/catalog/_partials/subcategories.tpl
+++ b/templates/catalog/_partials/subcategories.tpl
@@ -8,11 +8,18 @@
   {if (isset($display_subcategories) && $display_subcategories eq 1) || !isset($display_subcategories) }
     <div id="subcategories" class="{$componentName} row gx-3">
       {foreach from=$subcategories item=subcategory}
-        <div class="subcategory__wrapper col-6 col-lg-4 col-xl-3"> 
+        <div class="subcategory__wrapper col-6 col-lg-4 col-xl-3">
           <a class="subcategory" href="{$subcategory.url}" title="{$subcategory.name|escape:'html':'UTF-8'}">
             <div class="subcategory__image">
-              {if !empty($subcategory.image.large.url)}
-                <img class="img-fluid" src="{$subcategory.image.large.url}" alt="{$subcategory.name|escape:'html':'UTF-8'}" loading="lazy" width="{$subcategory.image.large.width}" height="{$subcategory.image.large.height}" />
+              {if !empty($subcategory.image.small.url)}
+                <img
+                  class="img-fluid"
+                  src="{$subcategory.image.small.url}"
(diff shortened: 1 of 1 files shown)


matthieu-rolland (approved):
LGTM 👍

Hlavtox:
Thanks Tomek :-)

pawelkierklo:
Where is option to disable images, when categories haven't image? Only injecting .subcategory__image { display: none; } in custom.css ??

Hlavtox:
@pawelkierklo I was thinking that we could loop the subcategories twice. Find out if all of them have an image, if not, do not render them.

templates/catalog/_partials/subcategories.tpl
```diff
@@ -8,11 +8,18 @@
...
+                <img
+                  class="img-fluid"
+                  src="{$subcategory.image.small.url}"
+                  width="{$subcategory.image.small.width}"
+                  height="{$subcategory.image.small.height}"
+                  alt="{$subcategory.name|escape:'html':'UTF-8'}"
+                  loading="lazy"
+                >
               {/if}
             </div>
```
SharakPL, line 24:
```smarty
            {if !empty($subcategory.image.small.url)}
              <div class="subcategory__image">
                <img
                  class="img-fluid"
                  src="{$subcategory.image.small.url}"
                  width="{$subcategory.image.small.width}"
                  height="{$subcategory.image.small.height}"
                  alt="{$subcategory.name|escape:'html':'UTF-8'}"
                  loading="lazy"
                >
              </div>
            {/if}
```
@Hlavtox no need for any extra loops 😉

But there's another issue. You can't access `$subcategory.image.small` when you remove main category image and use only the miniature, because whole ``$subcategory.image` is null then.

Hlavtox: This will be fixed for v9, I will do it with you. ;-)

SharakPL:
Showing just the name if there's no image should be enough in this layout. If not, then I suggest loading the logo with lowered opacity as a placeholder.

Issue #476, opened 2023-03-27
If category image don't exist, this will throw a null on php logs


rodriciru:
https://github.com/PrestaShop/hummingbird/blob/e1429f34253756117e4f669d48e569748fec9cd1/templates/catalog/_partials/category-header.tpl#L12

I think this only needs  {if $category.image} because if  {if $category.image} don't exist, the values of large and large.url it will be null or undefined
https://github.com/PrestaShop/hummingbird/pull/477

rodriciru:
https://github.com/PrestaShop/hummingbird/pull/477

tswfi:
I can't replicate this on ps 8.1.2 and current develop branch of hummingbird.

PR #477, opened 2023-03-27
Update category-header.tpl

rodriciru:
close https://github.com/PrestaShop/hummingbird/issues/476

### DIFF
--- a/templates/catalog/_partials/category-header.tpl
+++ b/templates/catalog/_partials/category-header.tpl
@@ -9,7 +9,7 @@
             {if $category.description}
               <div id="category-description" class="rich-text mb-4">{$category.description nofilter}</div>
             {/if}
-            {if !empty($category.image}
+            {if !empty($category.image)}
               <div class="category-cover mb-4">
                 <img src="{$category.image.large.url}" 
                   alt="{if !empty($category.image.legend)}{$category.image.legend}{else}{$category.name}{/if}" 
--- a/templates/catalog/_partials/category-header.tpl
+++ b/templates/catalog/_partials/category-header.tpl
@@ -9,7 +9,7 @@
             {if $category.description}
               <div id="category-description" class="rich-text mb-4">{$category.description nofilter}</div>
             {/if}
-            {if !empty($category.image.large.url)}
+            {if !empty($category.image}
               <div class="category-cover mb-4">
                 <img src="{$category.image.large.url}" 
                   alt="{if !empty($category.image.legend)}{$category.image.legend}{else}{$category.name}{/if}" 


sudo-do (dismissed):
Since I don't see a reason to get invalid index in line 14.

templates/catalog/_partials/category-header.tpl
```diff
@@ -9,7 +9,7 @@
             {if $category.description}
               <div id="category-description" class="rich-text mb-4">{$category.description nofilter}</div>
             {/if}
-            {if !empty($category.image.large.url)}
+            {if !empty($category.image}
```
mflasquin, line 12:
```suggestion
            {if !empty($category.image) && !empty($category.image.large.url)}
```

mflasquin (changes requested):

templates/catalog/_partials/category-header.tpl
sudo-do: isn't large url exist if image exists?

rodriciru (commented):
Not necessary IMO

rodriciru:
I dont know how to reject this changes, so i hope i do it ok

sudo-do:
yeah, the change is not nessesary as far as i know prestashop environment

Hlavtox (approved):

Hlavtox:
@rodriciru Can you show what error you are getting? I deleted the images from the category but I get no error 🤔

kpodemski:
@rodriciru that's odd, `empty` for this key should be safe 🤔  have you checked this file 078ccc9839dcacadc15c0f4aa16f402af5079aa_2.file.category.tpl.php on line 57 to see if it is really about this line?

rodriciru:
> @rodriciru that's odd, `empty` for this key should be safe 🤔 have you checked this file 078ccc9839dcacadc15c0f4aa16f402af5079aa_2.file.category.tpl.php on line 57 to see if it is really about this line?

Yep, it's weird as hell, but I have some answers.
Looking at compiled PHP, I see that some empty functions translate to simple `if`'s and others don't.
Concretely, this image `empty `translates to `if `and the `empty `2 lines above translate to an `empty `instead of an `if`:
**PLEASE KEEP IN MIND It's NOT EXACTLY THIS FILE, BUT IT'S ALMOST THE SAME**
```
            <?php if ($_smarty_tpl->tpl_vars['category']->value['image']) { ?>
                <div class="category-cover">
                  <img src="<?php echo htmlspecialchars((string) $_smarty_tpl->tpl_vars['category']->value['image']['large']['url'], ENT_QUOTES, 'UTF-8'); ?>
" alt="<?php if (!empty($_smarty_tpl->tpl_vars['category']->value['image']['legend'])) {
                  echo htmlspecialchars((string) $_smarty_tpl->tpl_vars['category']->value['image']['legend'], ENT_QUOTES, 'UTF-8');
                } else {
                  echo htmlspecialchars((string) $_smarty_tpl->tpl_vars['category']->value['name'], ENT_QUOTES, 'UTF-8');
                } ?>">
                </div>
            <?php } ?>
```
Effectively, `empty` not raises any warnings, but `if` does, so this is why we get the warning.
Still, I think this is an easy and no harmful at all fix, even if the blame of the issue is smarty

leemyongpakvn:
@florine2623 FYI, hummingbird is different from classic-theme. There is no `_dev` directory in hummingbird, you need to create a .env file inside its webpack directory, and run `npm run build` from hummingbird directory following this _cookbook_ https://github.com/PrestaShop/hummingbird#how-to-build-assets ;)

florine2623:
Hello @rodriciru ,

It seems like no one can reproduce your issue, neither @Hlavtox , @kpodemski nor @leemyongpakvn .

Could you explain further so it can be reproduced and tested ?

leemyongpakvn:
As usual, QA will close the issue if they can not reproduce it :ok_hand:

Hlavtox:
@rodriciru I will try to reproduce once again, if you really say it happens. There is no reason for smarty to convert empty to if. If it does so, it's stupid. 🤔

On what core version you are getting this behavior?

Prestaworks:
{if !empty($category.image.large.url)} is also present in classic theme, I tried deleting the category images, "image" => null shows on the category dump, tested on PS 8.1.2 and php 7.2, 7.4 and 8.1 still nothing in the php logs or Prestashop log files.

so if this is caused by "image" not existing it's maybe an older PS version?
To make the check safe it should maybe be something like
{if !empty($category.image) && !empty($category.image.large)} 

But I can't recreate the error on PS 8.1.2.

rodriciru:
I don't remember exactly the version, but I try to keep the latest version
of ps on my stores. So by the time the issue is created, the most recent
version at that time.

Anyway,seeing I'm the only,and can't reproduce by any of yours, let's close
and will open a new one if happens again

kpodemski:
@Hlavtox @Prestaworks maybe it was a bug in some Smarty version because I noticed that problem as well, I've had `{if !empty($array.key)}` it started flooding logs with a warnings/notice.

kpodemski:
@rodriciru 1.7.6 something

florine2623:
I suggest closing this issue as it cannot be reproduced https://github.com/PrestaShop/hummingbird/issues/476#issuecomment-1864780014

PR #492, opened 2023-04-23
refactor(product): simplify code

davidglezz:
| Questions         | Answers
| Description?      | Make the code less verbose:
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | no
| Sponsor company   | no
| How to test?      | jest

### DIFF
--- a/src/js/product.ts
+++ b/src/js/product.ts
@@ -11,34 +11,15 @@ export default () => {
   const {prestashop, Theme: {events}} = window;
 
   const initProductSlide = () => {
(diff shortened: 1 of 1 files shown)


boherm (approved):
Thanks for your contribution @davidglezz, LGTM.
It can be good to simplify some TS scripts for 9.0.0 like you've done, but we must be careful that the code does not become unreadable in the future!

Hlavtox (approved):

Issue #485, opened 2023-04-03
[FO] CART-  Wrong behavior when there are more items than there are in stock


Zoefinders:
### Describe the bug

In cart page, when you add more items than in stock, they are added and only warning popup is displayed.

### Expected behavior

The warning popup should be displayed as an error popup with red color and same content, and the quantity shound't be updated. 

### Steps to reproduce 

1. BO>create a product with qty=5
2. FO> add 5 of this product to the cart
3. Check the cart and try to add another one
4. A warning popup is displayed => Error (should be an error popup)
5. See error => The qty of product is updated and is displayed as 6

**Screen record**

[hb_cart.webm](https://github.com/PrestaShop/hummingbird/assets/23308427/ecf755b5-54fa-4d68-a647-68a98dc1233a)

### Additional infos

- HummingBird version: develop
- PS version: develop
- PHP version: 7.4

ga-devfront:
@Zoefinders is it possible to know what we need to display (mockup or more detailed description, maybe) ?

Hlavtox:
@ga-devfront I would use a toast with some error message. https://build.prestashop-project.org/hummingbird/?path=/story/ui-toasts--extended-toasts

Check if some nice error is coming with the response from the core. If not, we will find some string to display on theme side.

I am not sure how well big role https://github.com/PrestaShop/ps_shoppingcart/ plays in this behavior. The module may need to be updated. 👍

Hlavtox:
@Zoefinders The confirmation popup is handled by the https://github.com/PrestaShop/ps_shoppingcart/ module, so if TPL modification is not enough, maybe the module will need to be modified to handle the response from the core better.

tajriba:
hi,
we can also add more products to the cart from the product detail page (hummingbird+prestashop 801).
I see no fixes for this in prestashop 802.

ga-devfront:
Hello everyone !
I may have found a solution to our problem by changing the TPL and JS a bit.
However, I would like to have more precision on what is expected.

**Currently on Hummingbird** => It is possible to go above the quantity in stock (even if the merchant does not allow it in the BO), this generates a toast (warning) and the checkout button is always active.

**Currently on the classic** =>  It is possible to go above the quantity in stock (even if the merchant does not allow it in the BO), this generates an alert in top of cart (danger) and the checkout button is disabled.

[...]

PR #497, opened 2023-05-24
Fix checkout with outstock product

ga-devfront:
| Questions         | Answers
| Description?      | This fix is not the best fix we can do, but it works without touching any module or Core. We're disabling the checkout button on layout if any product is no more available, and we put the max limit of the product if you try to take more than in stock.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes [#485](https://github.com/PrestaShop/hummingbird/issues/485)
| How to test?      | Try to go to the cart page with limited product (without oops active) and put more than that is in stock.

### DIFF
--- a/src/js/components/useQuantityInput.ts
+++ b/src/js/components/useQuantityInput.ts
@@ -134,8 +134,25 @@ const updateQuantity = async (qtyInputGroup: Theme.QuantityInput.InputGroup, cha
               const errors = data.errors as string;
 
               if (errors) {
-                useToast(errors, {type: 'warning', autohide: false}).show();
+                useToast(errors, {type: 'danger', autohide: false}).show();
               }
+              // we check if the quantity is more than wanted and if oosp is disabled
+              // to recall with the correct number we can have
+              if (data.cart?.products?.length > 0) {
+                const productData = data.cart.products.find(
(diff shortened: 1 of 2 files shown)


guizme555 (approved):

EmnaChou:
I m going to test this PR

EmnaChou (commented):
Hello @ga-devfront 
1-There is no alert message when i add more items than available in stock 
2- i can add items as i want to the cart
Capture d’écran 2023-05-29 à 12 50 11
Capture d’écran 2023-05-29 à 12 49 58

EmnaChou (commented):
Hello @ga-devfront 
I ve tested the issue again  after switching to the branch to ticket_485/fix_checkout_with_bad_quantity' ( i was before on pr-497) , the obtained results are the following ones : 
1- on Prestashop version 8.1.0 : **the quantity of product is not updated** but the pop up is n't on her good location ( on the right corner of the page ) --> just need to check about it
2- on the version 8.0.2 : it doesn't work as it s expected

kpodemski:
Hummingbird should be tested only on PrestaShop 8.1+.

PR #506, opened 2023-06-11
Accessibility: add aria-hidden (and some aria-label)

yannicka:
| Questions         | Answers
| Description?      | The old PrestaShop theme had a number of accessibility problems. Let's try to do better with this new theme! This PR adds "aria-hidden" or "aria-label" to some elements. There's still a lot to do, but it's at least a start.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| Sponsor company   | N/A
| How to test?      | Install the theme and test all pages

### DIFF
--- a/templates/customer/password-new.tpl
+++ b/templates/customer/password-new.tpl
@@ -32,14 +32,14 @@
       <div class="mb-3">
         <label class="form-label">{l s='New password' d='Shop.Forms.Labels'}</label>
         <div class="input-group password-field js-parent-focus">
-            <input class="form-control js-child-focus js-visible-password" type="password" data-validate="isPasswd" name="passwd" value="">
-            <button
-              class="btn btn-primary"
-              type="button"
-              data-action="show-password"
-              data-text-show="{l s='Show' d='Shop.Theme.Actions'}"
-              data-text-hide="{l s='Hide' d='Shop.Theme.Actions'}"
-            >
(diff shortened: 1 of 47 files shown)


templates/customer/password-new.tpl
```diff
@@ -48,13 +48,13 @@
...
-              data-text-hide="{l s='Hide' d='Shop.Theme.Actions'}"
-            >
+          <button
+            class="btn btn-primary"
+            type="button"
+            data-action="show-password"
+            data-text-show="{l s='Show' d='Shop.Theme.Actions'}"
+            data-text-hide="{l s='Hide' d='Shop.Theme.Actions'}"
+          >
             <i class="material-icons">visibility</i>
```
matthieu-rolland, line 58: Is there a reason this one for not adding `aria-hidden` on this one ?

matthieu-rolland (approved):
thank you for your contribution @yannicka :+1: 

I'm not so familiar with `aria-hidden`, very nice idea indeed !

templates/customer/password-new.tpl
Hlavtox: @matthieu-rolland It's because the button does not have any text, so the icon actually serves a purpose.

Hlavtox (approved):

templates/customer/password-new.tpl
yannicka: @matthieu-rolland Fixed on PR #508

PR #524, opened 2023-06-27
Improvement: Rework Hn Structure

tblivet:
| Questions         | Answers
| Description?      | This PR has for objective to improve the **Hn** structure of hummingbird and ensure a harmonization of the different titles styles by adding specific class on each title instead of using Bootstrap class like `.h4` on a `<h2>` tag for example that can look confusing. In another hand, those modifications will help theme devs to edit global titles styles in an easier way.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | You can check for the Hn stucture of the main FO pages and ensure we have a good structure like H1 → H2 → H3 instead of H3 → H1 → H4 for example.


Hlavtox (commented):
Hi @tblivet, I love fixing the semantic strucutre. ❤️

What I don't agree with is adding the new `first-title` to `sixth-title`. In my opinion, it only brings extra data to transfer and `h1` to `h6` classes are understandable enough.

tblivet:
Hi @Hlavtox 🙂

I agree over the fact that it brings extra data, the initial approach was to separate the semantic Hn naming and the CSS class naming to avoid this type of code : `<h1 class="h3">` who in my opinion can be a bit strange.
But I think, it's more related to a habit than a good practice, so do we agree that I have to do the same but using the Hn classes ?

ga-devfront:
Hello, I understand your concerns, it is true that a lot of CSS variables have been added in this PR, however from my point of view this makes it easier to modify the theme settings by any dev without having to recompile what is really an asset and a good practice!

ga-devfront (dismissed):

Oksydan:
Hi,

I also don't like the idea to have more unneeded classes for headings. 
We are using BS5 and we should stick to it's typography classes. It should be easier for module developers to transition their solutions to hummingbird and style modules in general. We should follow one guidelines and don't reinvent the wheel. For new developers that are familiar with Bootstrap it will be also easier to add new functionalities w/o reading that whole css files to grab right class for heading.

templates/checkout/_partials/steps/checkout-step.tpl
```diff
@@ -18,9 +18,18 @@
             role = "tabpanel"
   >
     <div class="step__title js-step-title">
-      <h1 class="step__title-left h3">
+      {if isset($step_is_current) && $step_is_current}
```
kpodemski, line 21: I think it always exists so you can do this:
```suggestion
      {if $step_is_current eq true}
```

kpodemski (commented):
Great job @tblivet and thanks for listening to the feedback from the community :) I added one suggestion

Hlavtox (approved):

kpodemski (approved):

templates/checkout/_partials/steps/checkout-step.tpl
```diff
@@ -18,9 +18,18 @@
...
     <div class="step__title js-step-title">
-      <h1 class="step__title-left h3">
+      {if $step_is_current eq true}
+        <h1 class="step__title-left h3">
+      {else}
+        <p class="step__title-left h3">
+      {/if}
         {$title}
-      </h1>
+      {if $step_is_current eq true}
```
SharakPL, line 27: `{if $step_is_current}` should be enough.

templates/components/page-title-section.tpl
```diff
@@ -5,5 +5,5 @@
 {$componentName = 'page-title-section'}
 
 {block name='page_title_section'}
-  <h1 class="{$componentName} mb-4">{$title}</h1>
+  <h1 class="{$componentName} h2 mb-4">{$title}</h1>
```
SharakPL, line 8: If even the page title doesn't deserve class `h1` then maybe we should adjust the styling of `.h1`

SharakPL (approved):

templates/components/page-title-section.tpl
kpodemski: the idea is to keep the hierarchy and styling apart, it looks good

aniszr (approved):
Hello @tblivet 

Thanks for your PR, I tested it and all LGTM :heavy_check_mark: 

Thanks!

kpodemski:
thanks @tblivet 👏🏻

PR #526, opened 2023-06-30
Improvements : small theme improvements

tblivet:
| Questions         | Answers
| Description?      | This PR has for objective to add some improvements to the theme.
| Type?             | bug fix / improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | --

1. The breadcrumb has been removed in smarty from the homepage instead of making him invisible with CSS.
2. A condition has been added on jsonld data to adapt the "Type" if a product has a brand associated or not.
3. The loading attribute of ps_imageslider images has been optimized, the first image is charge with the loading parameter value `eager` which is better in terms of performance because the first image is identified as the LCP.
4. Improve the responsiveness of the pagination on category pages which have a lot of pages, like in the capture below, which show a content overflow on small devices:

image

### DIFF
--- a/modules/ps_imageslider/views/templates/hook/slider.tpl
+++ b/modules/ps_imageslider/views/templates/hook/slider.tpl
@@ -20,7 +20,7 @@
             aria-hidden="{if $smarty.foreach.homeslider.first}false{else}true{/if}">
             {if !empty($slide.url)}<a class="carousel-link" href="{$slide.url}">{/if}
               <figure class="carousel-content">
-                <img src="{$slide.image_url}" alt="{$slide.legend|escape}" loading="lazy" {$slide.size|replace: '"':''}>
+                <img src="{$slide.image_url}" alt="{$slide.legend|escape}" {if $slide@iteration == 1}loading="eager"{else}loading="lazy"{/if} {$slide.size|replace: '"':''}>
                 {if $slide.title || $slide.description}
                   <figcaption class="carousel-caption caption">
                     <h2 class="display-1 text-uppercase">{$slide.title}</h2>
--- a/src/scss/core/components/_pagination.scss
+++ b/src/scss/core/components/_pagination.scss
@@ -9,6 +9,7 @@
 
   &-container {
     align-items: center;
-    margin: 1rem 0;
+    margin-top: 1rem;
+    margin-bottom: 1rem;
   }
 }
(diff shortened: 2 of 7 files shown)


kpodemski (approved):
Looks good!

Hlavtox (approved):

Hlavtox:
@tblivet Good job dude, I love the level of attention to details! 🔥

modules/ps_imageslider/views/templates/hook/slider.tpl
```diff
@@ -20,7 +20,7 @@
             aria-hidden="{if $smarty.foreach.homeslider.first}false{else}true{/if}">
             {if !empty($slide.url)}<a class="carousel-link" href="{$slide.url}">{/if}
               <figure class="carousel-content">
-                <img src="{$slide.image_url}" alt="{$slide.legend|escape}" loading="lazy" {$slide.size|replace: '"':''}>
+                <img src="{$slide.image_url}" alt="{$slide.legend|escape}" {if $slide@iteration == 1}loading="eager"{else}loading="lazy"{/if} {$slide.size|replace: '"':''}>
```
SharakPL, line 23:
```suggestion
                <img src="{$slide.image_url}" alt="{$slide.legend|escape}" {if $slide@first}loading="eager"{else}loading="lazy"{/if} {$slide.size|replace: '"':''}>
```
But since `loading="eager"` is the default then it would be best to only add `loading="lazy"` for every item except the first one.

src/scss/core/components/_pagination.scss
```diff
@@ -9,6 +9,7 @@
 
   &-container {
     align-items: center;
-    margin: 1rem 0;
+    margin-top: 1rem;
+    margin-bottom: 1rem;
   }
```
SharakPL, line 14: I'd rather add proper classes to the template (`my-3 align-items-center`) instead of duplicating CSS that is already in Bootstrap. The theme should be as light as possible. Custom CSS should be added only to custom components that can't be styled directly with BS or in loops to limit the size of compiled php file

SharakPL (approved):
All good. Just small suggestions.

hibatallahAouadni (approved):
Hello @tblivet 

- The breadcrumb has been removed in smarty from the homepage :heavy_check_mark: 
- The loading attribute of ps_imageslider images has been optimized :heavy_check_mark: 
- Improve the responsiveness of the pagination on category pages :heavy_check_mark: 

LGTM, QA :heavy_check_mark:  

Thanks!

kpodemski:
thank you @tblivet 🎉

PR #533, opened 2023-07-28
Edit payment option to work with ps_checkout

tblivet:
| Questions         | Answers
| Description?      | Payment option display has been edited to make **ps_checkout** display properly
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | Check if payment methods are correctly displayed. ~Not easy to test because the shop must have **ps_checkout** configure, but this should render the payment option like in the screen below : top -> before & bottom -> after ![image]~(https://github.com/PrestaShop/hummingbird/assets/110676325/519e3161-c91a-465c-9f9b-9e6a7bbeefe4)


src/scss/custom/modules/_checkout.scss
```diff
@@ -0,0 +1,10 @@
+.payment__option {
```
kpodemski, line 1: This shouldn't be inside core of Hummingbird but rather inside your theme or ps_checkout module :)

kpodemski (changes requested):
ps_checkout is not a native module, this is why the fix shouldn't be there

if changes in .tpl are valid for other payment methods, it's ok, but CSS rules shouldn't be part of Hummingbird

you can easily add it to your version of the theme, or contribute to ps_checkout to display correctly if the theme is Hummingbird

tblivet:
Hello @kpodemski !
I have removed the CSS related to "ps_checkout".
I think the template is still relevant because it improves the display of payment option in case we have an image.

kpodemski (approved):

aniszr (approved):
Hello @tblivet 
I tested your PR, the checkout options are well displayed, LGTM :heavy_check_mark: 
QA :heavy_check_mark: 

See attached screenshot :
image

Thanks!

nicosomb:
Thank you @tblivet !

PR #542, opened 2023-08-10
Fix: when you edit the product qty to zero, the product won't be removed from the cart

### LINKED ISSUE #536: [Shopping Cart] when you edit the product qty to zero, the product won't be removed from the cart
when you change the product quantity to zero, usually the product will be deleted, but here the page shows me an error, and the product stay still in the cart, see the attached screen record below:

[issue_536.webm](https://github.com/PrestaShop/hummingbird/assets/23308427/98430456-f6b1-4f82-ba9b-eeed7a4f64e4)

### Expected behavior
The product should be deleted from the cart (check the Classic theme)

### Steps to reproduce

1. FO>add a product in the cart
2. Click on **Proceed to Checkout**
3. Try to minimize the product quantity to zero
4. An error is displayed and the product stay still in the cart

### PrestaShop version(s) where the bug happened
8.1.0

7.4

tblivet:
| Questions         | Answers
| Description?      | Correct the issue 536
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/536
| How to test?      | You can refer to the issue

### DIFF
--- a/src/js/pages/cart.ts
+++ b/src/js/pages/cart.ts
@@ -6,7 +6,6 @@
 import {Collapse} from 'bootstrap';
 import {isHTMLElement} from '@helpers/typeguards';
 import handleCartAction from '../components/UseHandleCartAction';
-import { qtyInput } from '@constants/selectors-map';
 
 export default () => {
   const {Theme} = window;
@@ -47,7 +46,7 @@ export default () => {
         if (eventTarget.classList.contains('js-increment-button')) {
           if (targetValue.dataset.mode === 'confirmation' && Number(targetValue.value) < 1) {
             if (removeButton) {
-             removeButton.click();
+              removeButton.click();
             }
           }
         }
--- a/src/js/pages/cart.ts
+++ b/src/js/pages/cart.ts
@@ -6,32 +6,12 @@
 import {Collapse} from 'bootstrap';
 import {isHTMLElement} from '@helpers/typeguards';
 import handleCartAction from '../components/UseHandleCartAction';
+import { qtyInput } from '@constants/selectors-map';
 
 export default () => {
   const {Theme} = window;
   const voucherCodes = document.querySelectorAll(Theme.selectors.cart.discountCode);
   const cartContainer = document.querySelector<HTMLElement>(Theme.selectors.cart.container);
-
-  if (cartContainer) {
-    cartContainer.addEventListener('click', (event: Event) => {
-      const eventTarget = event.target as HTMLElement;
-
-      if (eventTarget.classList.contains('js-decrement-button')) {
-        const targetItem = eventTarget.closest('.cart__item');
-        const targetValue = targetItem?.querySelector('.js-cart-line-product-quantity') as HTMLElement | null;
-
-        if (targetValue && targetValue.getAttribute('value') === '1' && targetValue.getAttribute('min') === '1') {
-          if (targetItem) {
-            const removeButton = targetItem.querySelector('.remove-from-cart') as HTMLElement | null;
-
-            if (removeButton) {
-              removeButton.click();
-            }
-          }
-        }
-      }
-    });
-  }
 
   voucherCodes.forEach((voucher) => {
     voucher.addEventListener('click', (event: Event) => {
@@ -56,8 +36,31 @@ export default () => {
   });
 
   if (cartContainer) {
+    console.log(cartContainer);
     cartContainer.addEventListener('click', (event: Event) => {
       const eventTarget = event.target as HTMLElement;
+
+      const targetItem = eventTarget.closest('.cart__item');
+      const targetValue = targetItem?.querySelector('.js-cart-line-product-quantity') as HTMLInputElement | null;
+      const removeButton = targetItem?.querySelector('.remove-from-cart') as HTMLElement | null;
+
+      if (targetValue) {
+        if (eventTarget.classList.contains('js-increment-button')) {
+          if (targetValue.dataset.mode === 'confirmation' && Number(targetValue.value) < 1) {
+            if (removeButton) {
+             removeButton.click();
+            }
+          }
+        }
+
+        if (eventTarget.classList.contains('js-decrement-button')) {
+          if (targetValue.getAttribute('value') === '1' && targetValue.getAttribute('min') === '1') {
+            if (removeButton) {
+              removeButton.click();
+            }
+          }
+        }
+      }
 
       if (eventTarget.dataset.linkAction === Theme.selectors.cart.deleteLinkAction) {
         handleCartAction(event);
--- a/src/js/pages/cart.ts
+++ b/src/js/pages/cart.ts
@@ -36,7 +36,6 @@ export default () => {
   });
 
   if (cartContainer) {
-    console.log(cartContainer);
     cartContainer.addEventListener('click', (event: Event) => {
       const eventTarget = event.target as HTMLElement;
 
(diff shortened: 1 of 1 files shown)


matks (dismissed):
Looks good to me 👍 

ping @ga-devfront for 2nd opinion

SharakPL:
IMO it would be better to simply disallow changing quantity to 0. This can be done by mistake and then it would have to be searched again. If a user wants to remove the product from the cart then the Remove button should be clicked.

Hlavtox:
@SharakPL That was how it was before. It should throw an error message "Minimal quantity for sale is 1".

Also, it's a bit incosistent. What does it do if minimum quantity for sale is higher than 1? 🤔

I am a bit torn apart for the expected behavior here.

sallemiines (approved):
Hello @tblivet 

Thanx for the PR ! LGTM QA :heavy_check_mark: 

https://github.com/PrestaShop/hummingbird/assets/31663369/7140700b-7cf9-4281-9bfc-aaa01023d520


Thank you :rocket:

sallemiines (dismissed):
Hello @tblivet 

Thanx for the PR ! LGTM QA :heavy_check_mark: 


https://github.com/PrestaShop/hummingbird/assets/31663369/0e0b481e-c8a6-4014-8163-64b8542ee07c




Thank you :rocket:

SharakPL:
@sallemiines could you please check again with product that has minimum order quantity higher than 1? It doesn't look like it's covered.

FabienPapet (dismissed):
blocking as we need to check before merge

MhiriFaten (commented):
Merging is blocked, I will remove the **`waiting for QA`** label until the changes requests are resolved.

kpodemski (dismissed):
I wanted to agree with @Hlavtox and @SharakPL, but I've just tested it on PS 8.1 with Classic, and it works like this:

The behavior should be as follows:
1. If you change from 1 to 0 using arrows, the product is deleted.
2. If you change from 1 to 0 using the input value, the product is deleted.
3. If you have a product with a minimal quantity set to 2 and you change it to 1, you have an error.

Why does it work in Classic, and it doesn't work with Hummingbird?
What's the difference between Classic and Humminbird?

If you have 1 unit in your shopping cart and you change it to 0 (either by using arrows or input), the request goes to `delete` action. In Hummingbird, it always goes to the `update` action, which is wrong.

This PR almost fixes the issue but there's a problem with getting a minimal quantity which could be different than 1.

kpodemski:
@tblivet Related function in Classic is inside:

themes/classic/_dev/js/cart.js in `updateProductQuantityInCart` function.

```
if (targetValue === '0') {
      $target.closest('.product-line-actions').find('[data-link-action="delete-from-cart"]').click();
    } else {
      $target.attr('value', targetValue);
      sendUpdateQuantityInCartRequest(updateQuantityInCartUrl, getRequestData(qty), $target);
    }
```

It looks that instead of checking minimal quantity you should only check whether the target value is equal to 0.

SharakPL:
Is there a place for improvements or does it have to be done like in Classic?

kpodemski:
@SharakPL what would be an improvement in your opinion?

SharakPL:
To be more clear on what each action does. Decreasing the amount shouldn't delete the product from the cart.
If the requested amount is lower than the minimum required for order then it should throw an error:

> The minimum amount required for order of this product is X. If you want to remove the product completely from your cart then use the Remove button

Or at least change [ - ] to [ X ] to indicate  that next use of this button will remove it completely instead of just decreasing the amount. Especially useful if the minimum required amount is higher than 1.

kpodemski:
The one improvement I see in this change is that if one makes a mistake and changes the value below 1, it could save the cart. I've checked a few stores, and when there are arrows, changing below 0 is forbidden. There's no request, which could also be a good thing.

Something to be decided by @PrestaShop/product-council :)

To summarize:
- If you change the value below 1, it should inform about minimal quantity or doesn't do anything.

SharakPL:
> I've checked a few stores, and when there are arrows, changing below 0 is forbidden.

Did you mean "below 1"?

kpodemski:
@SharakPL yes, I edited the message

Hlavtox:
@RosaBenouamer There needs to be a number input, dropdown for selecting count is not good UX. :-)

src/js/pages/cart.ts
```diff
@@ -10,6 +10,28 @@ import handleCartAction from '../components/UseHandleCartAction';
...
+
+  if (cartContainer) {
+    cartContainer.addEventListener('click', (event: Event) => {
+      const eventTarget = event.target as HTMLElement;
+
+      if (eventTarget.classList.contains('js-decrement-button')) {
+        const targetItem = eventTarget.closest('.cart__item');
+        const targetValue = targetItem?.querySelector('.js-cart-line-product-quantity') as HTMLElement | null;
+
+        if (targetValue && targetValue.getAttribute('value') === '1' && targetValue.getAttribute('min') === '1') {
```
jolelievre, line 23: Be careful the min quantity can also be 0 which should be handled like 1, meaning the feature is disabled there is no minimum
Except that I think the original behaviour from classic is consistent and should be used as a base as it's a common way to handle this in many e-commerce shops

tldr; let's stay iso functional

jolelievre: Ah and one other comment you seem to be listening to the click event, so only the arrows clicking but the value can also be edited via keyboard, so you should either listen to the `keydown` event as well or maybe easier simply the `change` event?

Unless you need to listen to the specific event to cancel it when the value is not the minimum one?

tblivet:
Hello 👋 With @ga-devfront we have update the PR we now have this behavior :
- When we click on decrement to go from 1 to 0 -> product is deleted
- When we click on decrement to go from 5 to 4 when product has minimum qty of 5 -> product is not deleted error message appear and qty remain the same.
- When we change the quantity manually (without + or - ) to 0 -> product is deleted
- When we change the quantity manually (without + or - ) to 0 and the product have minimum qty -> product is not deleted error message appear and qty remain the same.

ga-devfront (approved):

src/js/pages/cart.ts
```diff
@@ -33,12 +34,32 @@ export default () => {
...
+        }
+
+        if (eventTarget.classList.contains('js-decrement-button')) {
+          if (targetValue.getAttribute('value') === '1' && targetValue.getAttribute('min') === '1') {
+            if (removeButton) {
+              removeButton.click();
+            }
+          }
+        }
+      }
```
davidglezz, line 61:
```suggestion
      if (!removeButton || !targetValue) {
        return;
      }
      
      if (eventTarget.classList.contains('js-increment-button')) {
        if (targetValue.dataset.mode === 'confirmation' && Number(targetValue.value) < 1) {
            removeButton.click();
        }
      }

      if (eventTarget.classList.contains('js-decrement-button')) {
        if (targetValue.getAttribute('value') === '1' && targetValue.getAttribute('min') === '1') {
            removeButton.click();
        }
      }
```

The conde can be simplified, the firs `if` avoid `3 ifs`

djoelleuch (approved):
Hello @tblivet  , 
LGTM , 
If you change from 1 to 0 using arrows, the product is deleted. :heavy_check_mark: 
If you change from 1 to 0 using the input value, the product is deleted. :heavy_check_mark: 
If you have a product with a minimal quantity set to 2 and you change it to 1, you have an error.
:heavy_check_mark: 
**[but i think that the error message disappear too fast  ]** 


 

https://github.com/PrestaShop/hummingbird/assets/115152460/1bd6c18e-33ad-4920-9d1e-ec41959416bf

It's QA approved ✔️
thank you !

PR #547, opened 2023-09-05
Add classes to make menu elements tall as parent

### LINKED ISSUE #546: Unusable menu dropdown when store logo is tall
Hello,
When you have a tall logo on the menu, it is almost impossible to use the menu dropdowns.
This is because `<a>` element is not 100% percent tall of the available space (generate by the logo).
So when the dropdown is visible, and you move the mouse, it passes through a section which is not from the `<a>` element, and the dropdown menu closes.

https://github.com/PrestaShop/hummingbird/assets/10741347/7088c3a5-dfa5-42f5-9649-779520aa31ef

rodriciru:
Fixes #546 

I made the changes using bootstrap classes.
I don't find any drawbacks. If anyone find something wrong, tell me

### DIFF
--- a/templates/_partials/header.tpl
+++ b/templates/_partials/header.tpl
@@ -29,8 +29,8 @@
 {block name='header_bottom'}
   <div class="{$headerBottomName}">
     <div class="container-md {$headerBottomName}__container">
-      <div class="row gx-2 {$headerBottomName}__row">
-        <div class="col-auto logo order-xl-1 ms-lg-0 ms-2 me-lg-0 me-auto">
+      <div class="row gx-2 align-items-stretch {$headerBottomName}__row">
+        <div class="d-flex align-items-center col-auto logo order-xl-1 ms-lg-0 ms-2 me-lg-0 me-auto">
           {if $shop.logo_details}
             {if $page.page_name == 'index'}<h1 class="{$headerBottomName}__h1 mb-0">{/if}
               {renderLogo}
--- a/templates/_partials/header.tpl
+++ b/templates/_partials/header.tpl
@@ -29,7 +29,7 @@
 {block name='header_bottom'}
   <div class="{$headerBottomName}">
     <div class="container-md {$headerBottomName}__container">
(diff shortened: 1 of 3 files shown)


tblivet (changes requested):
Hello 👋
I think you want to do this modification, because if the logo is higher than the menu, we cannot access the submenu ? Is it right ?
Your modification looks good, but with those, the logo is no longer vertically centered.
So I suggest adding some classes to fix it.

templates/_partials/header.tpl
```diff
@@ -29,7 +29,7 @@
 {block name='header_bottom'}
   <div class="{$headerBottomName}">
     <div class="container-md {$headerBottomName}__container">
-      <div class="row align-items-center gx-2 {$headerBottomName}__row">
+      <div class="row gx-2 {$headerBottomName}__row">
```
tblivet, line 32:
```suggestion
      <div class="row gx-2 align-items-stretch {$headerBottomName}__row">
```

templates/_partials/header.tpl
```diff
@@ -29,7 +29,7 @@
...
         <div class="col-auto logo order-xl-1 ms-lg-0 ms-2 me-lg-0 me-auto">
```
tblivet, line 33:
```suggestion
        <div class="d-flex align-items-center col-auto logo order-xl-1 ms-lg-0 ms-2 me-lg-0 me-auto">
```

tblivet (changes requested):

rodriciru:
Thanks tblivet. You were right in your suggestions. Align-items-stretch it's not necessary, because in flex normal it's the same as stretch, but it doesn't matter.
https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#values

tblivet (approved):

matks (approved):

paulnoelcholot:
Hello @rodriciru,

I tested your PR and it's good for me! :tada: 

Thanks!

paulnoelcholot (approved):

PR #551, opened 2023-09-28
POC: theme icon map

Oksydan:
| Questions         | Answers
| Description?      | Simple POC replacing plain icon fonts with icons map and simple renderIcon smarty function. It should allow developers to adjust their icons in themes and allow module developers to use icons w/o being aware of what icons are being used by theme. Icon map should be standardized some way 🥶 
| Type?             | POC
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | https://github.com/PrestaShop/hummingbird/discussions/549
| Sponsor company   | Waynet
| How to test?      | N/A

### DIFF
--- a/templates/_partials/helpers.tpl
+++ b/templates/_partials/helpers.tpl
@@ -14,3 +14,66 @@
     >
   </a>
 {/function}
+
+{assign
+  var=iconsMap
+  value=[
+    "cart" => "shopping_cart"
+  ]
+  scope="global"
+}
+
+
+{** SVG SPRITE **}
+
+<svg class="d-none">
+  <defs>
+    <g id="shopping_cart">
+      <path
+        d="M280-80q-33 0-56.5-23.5T200-160q0-33 23.5-56.5T280-240q33 0 56.5 23.5T360-160q0 33-23.5 56.5T280-80Zm400 0q-33 0-56.5-23.5T600-160q0-33 23.5-56.5T680-240q33 0 56.5 23.5T760-160q0 33-23.5 56.5T680-80ZM208-800h590q23 0 35 20.5t1 41.5L692-482q-11 20-29.5 31T622-440H324l-44 80h480v80H280q-45 0-68-39.5t-2-78.5l54-98-144-304H40v-80h130l38 80Z"/>
+    </g>
+  </defs>
+</svg>
+
+
+{**
+ * @param string iconName
+ * @param array extraAttributes
+ * @return string
+ * @example
+ * {renderIcon name='cart' extraAttributes=['class' => 'my-extra-class']}
+ *}
+{function renderThemeIcon iconName="" extraAttributes=[]}
+    {if $iconName && !empty($iconsMap[$iconName])}
+
+      <span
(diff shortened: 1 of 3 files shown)


templates/_partials/helpers.tpl
```diff
@@ -14,3 +14,128 @@
...
+ * @param array extraAttributes
+ * @return string
+ * @example
+ * {renderIcon name='cart' extraAttributes=['class' => 'my-extra-class']}
+ *}
+{function renderThemeIcon iconName="" extraAttributes=[]}
+    {if $iconName && !empty($iconsMap[$iconName])}
+
+      <span
+        class="material-icons {if !empty($extraAttributes['class'])}{$extraAttributes['class']}{/if}"
```
matks, line 50: Here we have `material-icons` hardcoded so what would happen if someone wanted to use another icon lib? They override this smarty template?

matks (commented):
>  Icons map should be standardized in some way

I think indeed it's important to make it very clear how to build / provide this icon map else the feature can create confusion or bad usage

templates/_partials/helpers.tpl
Oksydan: @matks thank you for your response. 
Tbh modules aren't aware what icon fonts theme use. It's theme responsibility do display proper icon to given icon key. If you would like to change icon fonts to for example font-awsome, all you have to do is changing material icons to fa and map keys to your new icon-fonts.
Instead of:
```tpl
{assign
  var=iconsMap
  value=[
    "cart" => "shopping_cart"
  ]
  scope="global"
}
```

for example:

```tpl
{assign
  var=iconsMap
  value=[
    "cart" => "fa-shopping-cart" // not sure it's valid icon class
  ]
  scope="global"
}
```

SharakPL:
Looks like svgs are defined in 2 places: icons-sprite.svg and _helpers.tpl. I'm not sure this is a good idea. Also why don't you let the SVG be compiled and minified? IMO /assets should contain only compiled resources.

Oksydan:
Hi @SharakPL 
It's only for POC. It's not ready solution.

tblivet:
Thank you, @Oksydan. I think this proof of concept is a good idea. 👍

From my point of view, I'm not sure that using an icon map is a good idea. If we want to standardize this, we should not test if the `iconName` is part of the `iconsMap`, because this implies adding new icons that don’t appear on the map is impossible (for example, if a module dev wants to use a new icon).

For the SVG sprite, the idea is very interesting. This would make the use of SVG sprites much easier.

Finally, I want to point out a problem with this POC that I think can't be solved within the theme. In your example, you change the `shopping_cart` icon, and this icon is called inside a template which can be refreshed if we add a product to the cart. So, when we add a product to the cart, the "add to cart" modal doesn't show up because the function defined in Smarty .tpl is not registered globally, I suppose. This triggers an error and breaks the JS. Therefore, we can't use the stuff you propose inside `.tpl` files that are refreshed by JS (I think about multiple .tpl files call inside the product page for example). I think this can only be solved by adding a module in the theme that registers the Smarty function, or by adding this to the core in `php`. What do you think about that?

tblivet (changes requested):
PR need to work on refreshed elements

PR #553, opened 2023-10-04
Add demo link

MatShir:
@pawelkierklo is proving us a demo. Big thanks !

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 This is a PrestaShop's theme we are working on. Please, if you work on this theme, use the `8.0.x` branch of PrestaShop to make sure this theme is compatible with the latest `8.x` version.
 
 [Read more](https://build.prestashop.com/news/new-theme-announce/) about this theme on the blog.
+[Demo: https://hb.herf.pl/](https://hb.herf.pl/
+) provided by @pawelkierklo 
 
 ![image](https://user-images.githubusercontent.com/16455155/199937084-3d2eab3f-dc3e-488f-8b87-e8d4565219b3.png)
 


matks (changes requested):
Hi @MatShir I am sorry I am against this change.

I explain: it's really nice from @pawelkierklo but if we put this link inside the README, then we also set very high expectations.

If you are a github user and you go on any GitHub project, you expect that if there is a demo link in the README it is up to date.

Does @pawelkierklo commit to update his demo _every time we merge something in the `dev` branch_? Does he commit to keep it running 24H/24H, 7 days on 7? Does he commit to troubleshoot it when it has bugs?

If he does it's really cool 😄 

If he can't (and it's totally acceptable because it's very high expectations) then we should not put this link in the README. It's better to have NO demo instead of an OUTDATED demo.

Yes :/ demos are expensive and time-consuming to maintain

MatShir:
We can share the release version and a comment to report if there is an issue. Let's be simple and not overthink.

MatShir:
If the demo is down, it's not the end of the world. We remove the link when it is reported.

pawelkierklo:
I will update, I can also prepare a cron.php that will do it for me. I want to develop HB. Let's go!

MatShir:
Hi @pawelkierklo, 
As a new member of the community, we are 100% confident to have a demo which we don't have full control of. But we appreciate your involvement in the project, keep going 💪

pawelkierklo:
Ready! CRON will update "develop" theme on demo every night

MatShir:
Hi @pawelkierklo, 
Sorry for the back and forth, could you share the demo in the [discussion on Showcase category](https://github.com/PrestaShop/PrestaShop/discussions/new?category=showcase). Then, we will link the issue here, so later on if users can share report and feedback on the thread directly ;)

PR #558, opened 2023-10-25
Different spacing in hyperlinks on footer links.

pawelkierklo:
I added spaces in hyperlinks, which improved the footer's style.

user-footer

Used "Beautify HTML Formatter" in VSCode

### DIFF
--- a/modules/ps_customeraccountlinks/ps_customeraccountlinks.tpl
+++ b/modules/ps_customeraccountlinks/ps_customeraccountlinks.tpl
@@ -16,31 +16,65 @@
   </div>
   <ul class="footer__block__content footer__block__content-list collapse" id="footer_account_list">
     {if $customer.is_logged}
-      <li><a href="{$urls.pages.identity}" title="{l s='Information' d='Shop.Theme.Customeraccount'}" rel="nofollow">{l s='Information' d='Shop.Theme.Customeraccount'}</a></li>
+      <li><a href="{$urls.pages.identity}" title="{l s='Information' d='Shop.Theme.Customeraccount'}"
+          rel="nofollow">
+          {l s='Information' d='Shop.Theme.Customeraccount'}
+          </a></li>
       {if $customer.addresses|count}
-        <li><a href="{$urls.pages.addresses}" title="{l s='Addresses' d='Shop.Theme.Customeraccount'}" rel="nofollow">{l s='Addresses' d='Shop.Theme.Customeraccount'}</a></li>
+        <li><a href="{$urls.pages.addresses}" title="{l s='Addresses' d='Shop.Theme.Customeraccount'}"
+            rel="nofollow">
+            {l s='Addresses' d='Shop.Theme.Customeraccount'}
+            </a></li>
       {else}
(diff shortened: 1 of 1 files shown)


Hlavtox (commented):
That's not how to do spacing, you should increase the margin or padding on the list item. ;)

Moreover, I think the links without modifications look pretty good no?

pawelkierklo:
No, where you using ::before

Answer from ChatGPT:

`The difference you observed is due to the addition of a `::before` pseudo-element to the `a` element in your CSS. When you use `a::before { content: "> "; }`, you are actually adding a small character (in this case, the ">" character, but it can be anything) before the content of the hyperlink visually. This means that "Jakiś tekst" will be displayed after the ">" character, creating a visual indication effect.

However, the actual difference between `<a href="#">Jakiś tekst</a>` and `<a href="#"> Jakiś tekst </a>` without using CSS is not visible because any additional white spaces (spaces or new lines) inside the `<a>` tag are treated as white spaces and are removed in the rendering process.

If you use `a::before` with `content`, it will be inserted before the content of the hyperlink, resulting in a visible difference in the displayed text, but this is the result of additional CSS styling and not the HTML tag formatting itself.`

le-chatgpt:
> Answer from ChatGPT:

Oh you guys are taling about me!

SharakPL:
Sorry @pawelkierklo but it doesn't look like a good change. Spacing should be handled by CSS.

Closing for now, but feel free to add new commits if you come up with a better solution.

pawelkierklo:
@SharakPL Let me give you an example for technical understanding.

image

This is an example where the use of ::after and ::before is important because these elements will be offset from each other by the space given in the code.

pawelkierklo:
Please add this code to CSS

> .footer .footer__block__content-list a::before {
>     content: '|';
> }

and watch effect, only one file (ps_customeraccountlinks.tpl) haven't spaces in hyperlinks

image

SharakPL:
These spaces are **not intended**. Actually account links are the correct ones! It's the ps_linklist that needs changes. I hope this video will explain it:

https://github.com/PrestaShop/hummingbird/assets/5007456/6dd64525-f602-46cb-bb1d-14e02d125724

If you want to add a structure to the code then this is the correct way of wrapping strings, because it generates only the string without any extra whitespaces:
obraz

And this may look a little cleaner but it's just **BAD!** because it adds extra spaces and makes the generated code inconsistent:
obraz

___
Besides it looks like it's the problem with your custom theme since Hummingbird doesn't use any extra characters in footer links. A simple fix for you is to always use spaces inside _content_:
```css
.footer li a {
  content: "> "; // instead of ">"
```
This way it doesn't matter how many whitespaces are there wrapping your strings. The browser should always generate **a single whitespace**

pawelkierklo:
Following this pattern, you can see that hummingbird mostly has this bug

SharakPL:
I noticed. I'm preparing a fix for that - a general clean-up since I also found a few other problems (useless conditions, unclosed tag, duplicated attributes, etc.). It should be ready later today...

Hlavtox:
Misclick, we should remove the spaces from account links instead. :-)

PR #563, opened 2023-11-20
Use scrollIntoView when paging, fixes #562

### LINKED ISSUE #562: Really strange scrolling issue when paging with firefox
Sometimes `scrollTo(0,0)` does not work in firefox. This is really strange and I cannot even reproduce this all the time, but enough to be really annoying :D

Steps to reproduce
1) with "default demo products" change the pagesize to 3 (easier to get pager)
2) go to "art" category (it should have three pages as the pagesize is small)
3) scroll a bit to see the pager (important, also it seems that the viewport size also affects this a bit)
4) change page
=> https://github.com/PrestaShop/hummingbird/blob/986d1a2f7f67bc865774b9500b18051160932ffe/src/js/modules/facetedsearch/update.ts#L102 tries to scroll to top but it doesn't always happen (no errors, nothing).

It looks like firefox gets confused when the content changes and it doesn't always understand that it should scroll.. (also tried adding a small delay with `setTimeout` but that didn't seem to work either)

I can force it to work consistently by adding `window.scrollTo(0,1)` before the `window.scrollTo(0,0)` but that seems wrong.

This does not seem to ever happen with chrome.

As a side not the scrolling should probably happen to be at the top of `<div id="products">` so that for mobile users the products are shown even if the header is tall.. In place of window.scrollTo(0,0) should be something like

```js
document.getElementById('products').scrollIntoView(true); 
```

and as we have a fixed header give it some scroll margin

```css
section#products {
  scroll-margin-top: 150px; /* really the height of header */
}
```

In my opinion the scroll-margin-top should be set so that in desktop view the text "There are 8 products" is still visible and in mobile view filters and sort order are still visible

tswfi:
| Questions         | Answers
| Description?      | Instead of doing scrollTo(0,0) use scrollIntoView to get the `section#products` start to be visible.
| Type?             | bug fix && improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #562
| Sponsor company   | 
| How to test?      | See issue for steps for trying to replicate the original issue with firefox

WIP, still requires a css rule

### DIFF
--- a/src/js/modules/facetedsearch/update.ts
+++ b/src/js/modules/facetedsearch/update.ts
@@ -99,6 +99,7 @@ export default () => {
   prestashop.on(events.updateProductList, (data: Record<string, never>) => {
     updateProductListDOM(data);
     useQuantityInput();
-    window.scrollTo(0, 0);
+    // list was updated, scroll back up to to start of the list
+    document.getElementById('products')?.scrollIntoView(true);
   });
 };


matthieu-rolland (approved):

tswfi:
as noted by @Hlavtox here https://github.com/PrestaShop/hummingbird/issues/562#issuecomment-1819181912 the scrolling is not nice when choosing facets and after testing I agree.

Closing this as this requires a bit more thought on how to make this really nice.

PR #565, opened 2023-11-28
Change pager scrolling behaviour, Fixes #562

### LINKED ISSUE #562: Really strange scrolling issue when paging with firefox

tswfi:
| Questions         | Answers
| Description?      | Change how pager links scroll back to top of the product list. Also remove scrolling when changing facets
| BC breaks?        | no (maybe? one class changed?)
| Deprecations?     | no
| Fixed ticket?     | Fixes #562
| Sponsor company   | 
| How to test?      | Change facet filters => should not scroll so you can easily select other facets too. Use pager in product listing => should scroll to top (pretty consistently even with firefox)

### DIFF
--- a/src/js/modules/facetedsearch/update.ts
+++ b/src/js/modules/facetedsearch/update.ts
@@ -85,13 +85,12 @@ export default () => {
    */
   $('body').on('click', Theme.selectors.listing.pagerLink, (event) => {
     event.preventDefault();
-    document.getElementById('js-product-list-top')?.scrollIntoView({block: "start", behavior: "auto"});
+    document.getElementById('js-product-list-top')?.scrollIntoView({block: 'start', behavior: 'auto'});
     prestashop.emit(
       events.updateFacets,
       $(event.target)?.closest('a')?.get(0)?.getAttribute('href'),
     );
   });
-
 
   if ($(Theme.selectors.listing.list).length) {
     window.addEventListener('popstate', (e) => {
--- a/src/js/modules/facetedsearch/update.ts
+++ b/src/js/modules/facetedsearch/update.ts
@@ -85,7 +85,7 @@ export default () => {
    */
   $('body').on('click', Theme.selectors.listing.pagerLink, (event) => {
     event.preventDefault();
-    document.getElementById('js-product-list-top')?.scrollIntoView({block: 'start', behavior: 'auto'});
+    document.querySelector(Theme.selectors.listing.listTop)?.scrollIntoView({block: 'start', behavior: 'auto'});
     prestashop.emit(
       events.updateFacets,
       $(event.target)?.closest('a')?.get(0)?.getAttribute('href'),
--- a/src/js/modules/facetedsearch/update.ts
+++ b/src/js/modules/facetedsearch/update.ts
@@ -80,6 +80,19 @@ export default () => {
     );
   });
 
+  /**
+   * Pager links also scroll up
+   */
+  $('body').on('click', Theme.selectors.listing.pagerLink, (event) => {
+    event.preventDefault();
+    document.getElementById('js-product-list-top')?.scrollIntoView({block: "start", behavior: "auto"});
+    prestashop.emit(
+      events.updateFacets,
+      $(event.target)?.closest('a')?.get(0)?.getAttribute('href'),
+    );
+  });
+
+
   if ($(Theme.selectors.listing.list).length) {
     window.addEventListener('popstate', (e) => {
       const {state} = e;
@@ -99,6 +112,5 @@ export default () => {
   prestashop.on(events.updateProductList, (data: Record<string, never>) => {
     updateProductListDOM(data);
     useQuantityInput();
-    window.scrollTo(0, 0);
   });
 };
--- a/src/js/constants/selectors-map.ts
+++ b/src/js/constants/selectors-map.ts
@@ -25,6 +25,7 @@ export const listing = {
   listHeader: '#js-product-list-header',
   searchFiltersClearAll: '.js-search-filters-clear-all',
   searchLink: '.js-search-link',
+  pagerLink: '.js-pager-link',
 };
 
 export const cart = {
--- a/src/scss/custom/components/category/_product-list.scss
+++ b/src/scss/custom/components/category/_product-list.scss
@@ -6,6 +6,9 @@
 
 .layout-left-column,
 .layout-full-width {
+  #js-product-list-top {
+    scroll-margin: 120px;
+  }
   .products-selection {
     margin-bottom: 1.25rem;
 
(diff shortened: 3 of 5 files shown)


tleon (dismissed):

matthieu-rolland (dismissed):

src/js/modules/facetedsearch/update.ts
```diff
@@ -80,6 +80,18 @@ export default () => {
     );
   });
 
+  /**
+   * Pager links also scroll up
+   */
+  $('body').on('click', Theme.selectors.listing.pagerLink, (event) => {
+    event.preventDefault();
+    document.getElementById('js-product-list-top')?.scrollIntoView({block: 'start', behavior: 'auto'});
```
SharakPL, line 88: Maybe use here `Theme.selectors.listing.listTop`

tswfi:
```
src/js/constants/selectors-map.ts:  listTop: '#js-product-list-top',
```
Those are in jQuery format.. should there be a `listTopId: 'js-product-list-top'`

tswfi: Or 

```diff
--- a/src/js/modules/facetedsearch/update.ts
+++ b/src/js/modules/facetedsearch/update.ts
@@ -85,7 +85,8 @@ export default () => {
    */
   $('body').on('click', Theme.selectors.listing.pagerLink, (event) => {
     event.preventDefault();
-    document.getElementById('js-product-list-top')?.scrollIntoView({block: 'start', behavior: 'auto'});
+    /* selectors are in jQuery format, slice the leading # character out */
+    document.getElementById(Theme.selectors.listing.listTop.slice(1))?.scrollIntoView({block: 'start', behavior: 'auto'});
     prestashop.emit(
       events.updateFacets,
       $(event.target)?.closest('a')?.get(0)?.getAttribute('href'),
```

but this feels a bit dirty..

SharakPL: > ```
> src/js/constants/selectors-map.ts:  listTop: '#js-product-list-top',
> ```
> 
> Those are in jQuery format.. should there be a `listTopId: 'js-product-list-top'`

Well, first of all I'm not sure it should even be an ID. Any `js-...` selector should be a class.

And what do you mean by _jQuery format_? They are simple selectors :)
Just use `document.querySelector(Theme.selectors.listing.listTop)` instead of `document.getElementById('js-product-list-top')`

tswfi: well, css selector (where `#` denotes id) which jQuery uses everywhere..

vanilla js getElementById takes only the id not the css selector. 

getElementById is faster than querySelector (not by much but it is by definition).

I'll change it to use querySelector

SharakPL: Some say [querySelector is faster](https://www.reddit.com/r/learnjavascript/comments/i0f5o8/performance_of_getelementbyid_vs_queryselector/?rdt=42342) 😉 Doesn't matter nowadays really, but just for consistency we should use `querySelector` and `querySelectorAll`. Especially since we're using selectors-map.

tswfi: yup, it was changed here [4a29900](https://github.com/PrestaShop/hummingbird/pull/565/commits/4a299009dbe9758aa6a5da22a65f2921bc252f14)

SharakPL (approved):

florine2623 (commented):
Hello @tswfi ,

I have 2 different behaviors with your PR.

With Firefox, when changing/using filter, _**it doesn't scroll back to the top**_. So the customer can filter multiple attributes at the same time.

https://github.com/PrestaShop/hummingbird/assets/16019289/12716e64-5f7b-4abc-a573-d5ea8b373558

On Chrome, _**it always scrolls to the top**_. 

https://github.com/PrestaShop/hummingbird/assets/16019289/5caafa33-c62c-46aa-b86a-dfeedc5241fa

The behavior should be consistent of both browsers. It should be like Firefox.
I have `npm run build` before testing on both browsers.

Could you check ? ^^
Thanks!

tswfi:
Cannot replicate on OSX chrome, did you also clear your browser cache (or force refresh)?

image

https://github.com/PrestaShop/hummingbird/assets/1352132/ede217de-d3c5-4563-ae0a-6ccf456990e2

There are some minor "jumps" as new things are added to the page (like the "remove this filtering facet" buttons above the products and "clear all" button above the facets, but those are outside of the scope).

florine2623:
Yes, i have updated my chrome to version 120.0.6099.129, empty cache and hard reload 🤔 

Maybe another @PrestaShop/qa-functional could test it. In any case, it works well with Firefox. It just needs to be checked one more time on Chrome.

SharakPL:
TBH I never liked filters in the left column. It's too messed up. Would be much better to move it on top of the products list in the center column and categories block to the top.

tswfi:
> TBH I've never liked filters in the left column. It's too messed up. Would be much better to move it on top of the products list in the center column and categories block to the top.

A setting would be the best.. Some merchants have lots of filters and having them on top of the products is not that easy to style nicely. Preferably there should be at least three different ways to show the filters. 1) like now 2) above the products 3) as a "sliding panel" from a button

Anyway, changing the logic is out of scope for this pull request.

AureRita:
Hi @tswfi 

Thank you for your PR, I tested it and it seems to works in my case as you can see : 

[recording(80).webm](https://github.com/PrestaShop/hummingbird/assets/100699445/20e63efd-9393-450d-ae5c-9184aa87dd96)

Tested on develop and 8.1.x.

Thank you

AureRita (approved):

PR #570, opened 2023-12-19
Fix: Separate product reference and manufacturer

### LINKED ISSUE #569: Product reference only shown if manufacturer is set
If product does not have manufacturer set the product reference wont be shown. 

See here: https://github.com/PrestaShop/hummingbird/blob/develop/templates/catalog/_partials/product-details.tpl#L17

Also the block naming is kinda misleading as the block is named product_reference but it contains both manufacturer and reference information.

Also would be really nice to have specific class for each `detail` li element. Maybe the `detail__title` information could be used to generate class for the li (does smarty have a `slugify` formatter)? That would allow doing things like highlighting one specific detail with just css.

dennispw:
| Questions         | Answers
| Description?      | Separates product reference and manufacturer information to different blocks and if statements.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #569

### DIFF
--- a/templates/catalog/_partials/product-details.tpl
+++ b/templates/catalog/_partials/product-details.tpl
@@ -13,7 +13,7 @@
   <div id="product-details-collapse" class="info__content accordion-collapse collapse {if !$product.description}show{/if}" data-bs-parent="#product-details-heading" aria-labelledby="product-details-heading">
     <div class="accordion-body">
       <ul class="product__details">
-        {block name='product_reference'}
+        {block name='product_manufacturer'}
           {if isset($product_manufacturer->id)}
             <li class="detail">
               <div class="detail__left">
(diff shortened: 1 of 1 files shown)


templates/catalog/_partials/product-details.tpl
```diff
@@ -29,17 +29,19 @@
...
-                <div class="detail__left">
-                  <span class="detail__title">{l s='Reference' d='Shop.Theme.Catalog'}</span>
-                </div>
-                <div class="detail__right">
-                  <span>{$product.reference_to_display}</span>
-                </div>
-              </li>
-            {/if}
+        {block name='product_reference'}
+          {if isset($product.reference_to_display) && $product.reference_to_display neq ''}
```
SharakPL, line 36:
```suggestion
          {if !empty($product.reference_to_display)}
```
Should be enough

SharakPL (approved):
LGTM

boherm (approved):

florine2623 (approved):
Hello @tswfi ,

Tested with all types of products. LGTM
Tested with/out supplier, with/out supplier ref, with/out combination ref.

https://github.com/PrestaShop/hummingbird/assets/16019289/031530bd-7de4-424b-9391-599227342f0d

Thanks!

SharakPL:
Thanks @dennispw and @florine2623

PR #568, opened 2023-12-15
fix: productcomments issues

tblivet:
Make productcomments work again, and small display improvements has been added

| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Fix issues on productcomments modules.<br>- Display stars on product listing<br>- Bugfix when there were comment on a product they were not displayed<br>- Minor display improvements
| Type?             | bug fix / improvement
| BC breaks?        | no
| Deprecations?     | 
| Fixed ticket?     | 
| Sponsor company   | @PrestaShopCorp
| How to test?      | I think that a global functionality check has to be performed on productcomment module.

PR #584, opened 2024-01-31
Fix missing add to wishlist button and modals backdrop

SharakPL:
| Questions         | Answers
| Description?      | #568 prevented Add to wishlist button from loading on product list and caused error thrown in the console. `thumbnail-container` needs to be back in its original place.<br>It temporary breaks the product rating, but this needs to be fixed in the productcomments module, where wrong selector is being used in js (fix: https://github.com/PrestaShop/productcomments/pull/198)
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

### Before
obraz

### After
obraz

### DIFF
--- a/modules/blockwishlist/views/templates/components/modals/add-to-wishlist.tpl
+++ b/modules/blockwishlist/views/templates/components/modals/add-to-wishlist.tpl
@@ -26,8 +26,7 @@
             @click="toggleModal"
             data-dismiss="modal"
             aria-label="{l s='Close' d='Modules.Blockwishlist.Shop'}"
-          >
-          </button>
+          ></button>
         </div>
 
         <div class="modal-body">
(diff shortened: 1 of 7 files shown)


Hlavtox (dismissed):

SharakPL:
Further changes will be needed inside the productcomments module to avoid this empty space when no rating is available. It's meant for absolutely positioned Classic verstion and it only changes `visibility`.

IMO it's not the best place for it in Hummingbird. Products without rating will have either big empty space or different position of quantity and buttons and there will be layout shifts all over the page. Maybe we should keep it on the product cover like in Classic, but rather as a single star with score in it eg. 4.5/5, instead of 5 stars.

boherm (dismissed):

Hlavtox (approved):

AureRita (approved):
Hi @SharakPL 

[recording(92).webm](https://github.com/PrestaShop/hummingbird/assets/100699445/f5b55d9b-9706-41e6-822d-b43cefb94d93)

Tested on develop and 8.1.x 

Because the PR seems to works perfectly, it's QA :heavy_check_mark: 

Thank you

PR #586, opened 2024-02-02
WIP: Accessibility: Add skip to content button

yannicka:
Don't push it as it is: firstly, I don't know if the naming is right. Secondly, on some pages, like the category page for example, clicking on the link redirects to `/[object History]`. I haven't yet been able to find out why, so if anyone has any ideas, I'd be grateful for any help.

| Questions         | Answers
| Description?      | Add a "Skip to content" button for accessibility.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | No, but help #251
| Sponsor company   | N/A
| How to test?      | Load a page and press "tab"

Screenshot:
image

### DIFF
--- a/src/scss/custom/layout/_header-top.scss
+++ b/src/scss/custom/layout/_header-top.scss
@@ -25,3 +25,16 @@ $component-name: header-top;
     }
   }
 }
+
+.skip-link {
+  &__wrapper {
+      @include visually-hidden-focusable;
+  }
+
+  &__btn {
+      position: absolute;
+      top: 1em;
+      left: 1em;
+      z-index: 200;
+  }
+}
--- a/templates/layouts/layout-both-columns.tpl
+++ b/templates/layouts/layout-both-columns.tpl
@@ -18,6 +18,14 @@
       {hook h='displayAfterBodyOpeningTag'}
     {/block}
 
+    {block name='skip_link'}
+      <div class="skip-link__wrapper">
+        <a href="#wrapper" class="btn btn-primary skip-link__btn">
+          {l s='Skip to content' d='Shop.Theme.Actions'}
+        </a>
+      </div>
+    {/block}
+
     {block name='product_activation'}
       {include file='catalog/_partials/product-activation.tpl'}
     {/block}


Hlavtox:
I don't know what to think of this, I have never seen a button like this in my life. 🤷‍♂️

yannicka:
> I don't know what to think of this, I have never seen a button like this in my life. 🤷‍♂️

If you press "tab" on websites, this link will be displayed.

Example on GitHub:
image

On gov.uk:
image

On wordpress.org:
image

Etc.

https://css-tricks.com/how-to-create-a-skip-to-content-link/

Of course, my PR is useless as long as these bugs exist.

Hlavtox:
@yannicka Cool!!! I learned something new. :-)

Well, I think the bug with the `Object-history` thing is the same issue as with the profiler right? https://github.com/PrestaShop/PrestaShop/issues/33260

yannicka:
> @yannicka Cool!!! I learned something new. :-)
> 
> Well, I think the bug with the `Object-history` thing is the same issue as with the profiler right? [PrestaShop/PrestaShop#33260](https://github.com/PrestaShop/PrestaShop/issues/33260)

Yes, the issue seems very similar if not identical.

dennispw:
Where does the "content" start in this case? On catalog pages (categories, search, etc.) I *guess* the content would be the products? What if a category don't have any products but sub categories with products in them? Where is the "content" on the start page?

This will become *very* theme and module dependent. Even with hummingbird and standard PS modules on the start page (as an example) they can freely change positions and/or activate/deactivate modules.

I'm not opposed to the idea as such, but I think it's a quirky one to implement correctly since the very nature of PrestaShop is customizability the "content" in this case can (and will) different for different shops.

yannicka:
The link depends on the theme, as it leads to the tag with the id `wrapper`. But since we're in the theme itself, I don't see any problem with that.

The content may indeed be different between pages, but it's often considered to be what comes after the header. Which is the case here, the tag is just after the header:

https://github.com/PrestaShop/hummingbird/blob/733a70cf5f6de7188cf6dc4f3e4a7b3b44a64ab1/templates/layouts/layout-both-columns.tpl#L25-L36

On some sites, the skip link is more complex and may contain several links to take you to different places, but this would require more work.

image

As it stands, considering that the content is always in the `wrapper` tag seems fine to me, since we're not trying to avoid the page title or anything else. So it works just as well on the home page as on a category page, a product page or any other page.

PR #587, opened 2024-02-05
Changed lazyloading with highest fetchpriority in category header

buggyzap:
| Questions         | Answers
| Description?      | I have changed the lazyloading with a more accurate property for this element, the fetchpriority=high.
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | Open Chrome devtools in Performance tab and measure that image is loading first.

# Common scenario

We have to load an header image, the LCP of the page depends on it, so we cannot load it in lazy loading mode but in fetchpriority=high, so in network order load of resources we get an highest priority and a fastest rendering.

## Additional Resources

https://web.dev/articles/fetch-priority

### DIFF
--- a/templates/catalog/_partials/category-header.tpl
+++ b/templates/catalog/_partials/category-header.tpl
@@ -13,7 +13,7 @@
               <div class="category-cover mb-4">
                 <img src="{$category.image.large.url}" 
                   alt="{if !empty($category.image.legend)}{$category.image.legend}{else}{$category.name}{/if}" 
-                  loading="lazy" 
+                  fetchpriority="high" 
                   class="img-fluid"
                   width="{$category.image.large.width}" 
                   height="{$category.image.large.height}">


kpodemski (approved):
`fetchpriority` is not available in Firefox, but I don't think it's a blocker, since the image is above the fold I think the change is legit

edit: I wonder if we could mix loading="eager" and this? 🤔 

I'll ping other people to get some other opinions
cc @ga-devfront @tblivet @Hlavtox @SharakPL

ga-devfront:
Interesting, does this really benefit all devices? Do we really feel a change in performance? as @kpodemski  suggests I think we can mix that with `loading="eager" `. @tblivet  an opinion?

ga-devfront (approved):

Hlavtox (approved):
https://stackoverflow.com/questions/77744344/is-it-okay-to-use-both-fetchpriority-high-and-loading-eager-in-img-tag

We can also add loading eager 👍

tblivet:
I also agree with adding loading eager 👍

buggyzap:
> https://stackoverflow.com/questions/77744344/is-it-okay-to-use-both-fetchpriority-high-and-loading-eager-in-img-tag
> 
> We can also add loading eager 👍

Hi guys, loading eager is the default value so its redundant to add it: https://www.w3schools.com/tags/att_img_loading.asp

We can just add fetchpriority.

PR #593, opened 2024-02-09
Templates: fix formatting

yannicka:
| Questions         | Answers
| Description?      | Currently there is no consistency in formatting. Sometimes there are two spaces, sometimes four; sometimes the tags are not aligned; sometimes single quotes are used, sometimes double quotes; sometimes there are spaces between tags, sometimes not, etc. The goal of this PR is to improve the situation, even if there is still a lot to do.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| Sponsor company   | N/A
| How to test?      | Install the theme and browse the website.

### DIFF
--- a/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
+++ b/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
@@ -4,7 +4,7 @@
  *}
 <section class="best-sellers-products mt-3">
   <div class="container">
-    {include file='components/section-title.tpl' title={l s="Best Sellers" d="Shop.Theme.Catalog"}}
+    {include file='components/section-title.tpl' title={l s='Best Sellers' d='Shop.Theme.Catalog'}}
 
     {include file='catalog/_partials/productlist.tpl' products=$products productClass='col-6 col-lg-4 col-xl-3'}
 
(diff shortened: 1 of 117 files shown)


SharakPL:
I started something similar, but it's not worth it. You would only trigger lots of new translations for all languages so a lot of your changes are really unnecessary. Sorry
 
The rest can be fixed along the way while fixing other bugs without triggering updates of 100+ files.

Hlavtox:
@yannicka Hey, first off, it's very cool to fix all the formatting things, but PR big like this makes it very difficult to review. :(

There are some things I don't like, for example the empty lines everywhere and others, but that could be discussed.

Do you think we can separate the PR into smaller chunks? For example - a PR that will only fix quotes `"` to `'`? That would be easy to review and could be merged right away. :-)

Also, you need to rebase it anyway, because some changes were merged.

yannicka:
I'll see about cutting my PR into smaller chunks when I have time:
- Change of translations.
- Changes to single quotes/double quotes.
- Changing long lines to several lines.
- Add blank lines to space out.
- Changing indentations.
- Other small changes.

yannicka:
`"` to `'` done in #601. It's the least interesting change, but also one of the biggest (in terms of differential, not final cleanliness, as the change itself only brings consistency).

yannicka:
Blank lines added to improve readability: #602.

yannicka:
Fix search translation in #603.

PR #600, opened 2024-02-22
Release 0.2.0

matthieu-rolland:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Release 0.2.0
| Type?             | 
| BC breaks?        | 
| Deprecations?     | 
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      |

PR #606, opened 2024-03-08
Fix logo size

boherm:
| Questions         | Answers
| Description?      | After tests review for https://github.com/PrestaShop/hummingbird/pull/600#pullrequestreview-1898190625,<br>we need to fix the logo size when the uploaded image is taller than longer.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Not yet.
| Sponsor company   | PrestaShop SA
| How to test?      | See @florine2623 review in https://github.com/PrestaShop/hummingbird/pull/600#pullrequestreview-1898190625

-- 
### Additional info issue from the release tests

- Related to https://github.com/PrestaShop/hummingbird/issues/546
Tested with this file :
Screenshot 2024-02-23 at 14 59 44

Here's the result :x:

https://github.com/PrestaShop/hummingbird/assets/16019289/c8d1d6ad-7f02-4b0f-9c68-51f09b7b150c

### DIFF
--- a/src/scss/custom/layout/_header-bottom.scss
+++ b/src/scss/custom/layout/_header-bottom.scss
@@ -8,7 +8,14 @@ $component-name: header-bottom;
   border-bottom: var(--header-bottom-border-bottom);
 
   .navbar-brand {
+    max-width: 250px;
+    max-height: max(50px, 10vh);
     padding: 0;
+
+    .logo {
+      width: auto;
+      max-height: inherit;
+    }
   }
 
   .nav-link {


templates/_partials/helpers.tpl
```diff
@@ -6,7 +6,7 @@
 {function renderLogo}
   <a class="navbar-brand d-block" href="{$urls.pages.index}">
     <img
-      class="logo img-fluid"
+      class="logo"
```
SharakPL, line 9:
```suggestion
      class="logo img-fluid"
```
This class is required

SharakPL (changes requested):

tblivet (dismissed):
Looks good to me 👍
If we limit the height, maybe we also need to do the same for the width ?
Like `max-width: 150px;` ?

SharakPL:
> Like `max-width: 150px;` ?

Yes, but from my experience it should be around 250px

SharakPL (approved):

AureRita:
Hi @boherm 

Thank you for your PR, currently I tested it and, yes, the logo is a little bit smaller but I think it's not enought as you can see : 

image


FYI : you can scroll under the logo 

Can you correct it ? 

Waiting for your feedback

boherm:
Hi @AureRita, thanks for the review!

This is only on the homepage or on a specific page?

It looks like you haven't properly rebuild hummingbird theme or don't clear the browser cache. 🤔
Have you done this steps?

@SharakPL, @tblivet,
So, do you prefer adding `max-width: 250px;` property and also removing `.img-fluid` (that is totally bypassed by this rule)?

src/scss/custom/layout/_header-bottom.scss
```diff
@@ -9,6 +9,11 @@ $component-name: header-bottom;
 
   .navbar-brand {
     padding: 0;
+
+    .logo {
+      width: auto;
+      max-height: 50px;
+    }
```
SharakPL, line 16:
```suggestion
  .navbar-brand {
    padding: 0;
    max-width: 250px;
    max-height: max(50px, 12vh);

    .logo {
      max-height: inherit;
    }
```
This should work nicely.

SharakPL (changes requested):

SharakPL (approved):

AureRita (approved):
Hi @boherm 

[recording(140).webm](https://github.com/PrestaShop/hummingbird/assets/100699445/8a8222ca-a985-4411-948a-2cf12ce034b0)

Tested on develop and 8.1.x

Thank you

PR #609, opened 2024-03-14
Fix minimum qty input on products listing and product detail pages

### LINKED ISSUE #608: The Min quantity config is always preselected regardless of the quantity in the cart
### Issue 

As a merchant, I don't want to impose a min quantity in the product listing if the min quantity is already reached in the cart 

#### Detail issue
- Related to https://github.com/PrestaShop/hummingbird/pull/548
When you add a product with a minimal quantity to your cart, the product quantity should be updated. 
Example : my minimal quantity is 3, I add 3 products to my cart, I Continue shopping, I should be able to add just 1 product to my cart instead of 3. :x:
The behavior in the shopping cart is correct ✅ 

https://github.com/PrestaShop/hummingbird/assets/16019289/5d079315-3976-4022-ab30-7a1abb2ffb01

#### Technical
The feature works theortically on the theme but can't work on the PrestaShop 8.1 since we don't have the cart content on the product listing.

So, first step would be to add on the product listing and other related page the possibility to check the cart content from the front side

Waiting for more investigation, on the feasibility of the task

boherm:
| Questions         | Answers
| Description?      | With this PR, when you add a product with a minimal quantity to your cart, the product quantity should be updated in the quantity input of the product. This is for all products listing pages, product quick view and also product detail page.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/608 (related to https://github.com/PrestaShop/hummingbird/pull/600)
| Sponsor company   | PrestaShop SA
| How to test?      | 0. Don't forget to rebuild the theme before!<br>1. Set a minimal quantity for ordering a product in BO.<br>2. Then go to the FO and search your product in categories pages, and see the value displayed in the quantity input.<br>3. After added to the cart, you must have "1" in the quantity input instead of the minimal quantity required to put the product on your cart.

### DIFF
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -176,7 +176,7 @@
               {/block}
             </div>
 
-            {if $product.add_to_cart_url}
+            {if $product.add_to_cart_url && $product.product_type !== 'combinations'}
               <form action="{$urls.pages.cart}" method="post" class="d-flex flex-wrap flex-md-nowrap gap-3 align-items-center mt-3">
                 <input type="hidden" value="{$product.id_product}" name="id_product">
 
@@ -185,9 +185,9 @@
                 <div class="quantity-button js-quantity-button w-100 w-sm-auto">
                   {include file='components/qty-input.tpl'
                     attributes=[
-                      "id"=>"quantity_wanted_{$product.id_product}",
-                      "value"=>"{if $product.cart_quantity && $product.cart_quantity >= $product.minimal_quantity}1{else}{$product.minimal_quantity}{/if}",
-                      "min"=>"{if $product.cart_quantity && $product.cart_quantity >= $product.minimal_quantity}1{else}{$product.minimal_quantity}{/if}"
+                      "id" => "quantity_wanted_{$product.id_product}",
+                      "value" => "{$product.minimal_quantity}",
+                      "min" => "{$product.minimal_quantity}"
                     ]
                     marginHelper="mb-0"
                   }
--- a/templates/catalog/_partials/product-add-to-cart.tpl
+++ b/templates/catalog/_partials/product-add-to-cart.tpl
@@ -44,9 +44,9 @@
         <div class="product-actions__quantity quantity-button js-quantity-button col-md-auto">
           {include file='components/qty-input.tpl'
             attributes=[
-              "id"=>"quantity_wanted",
-              "value"=>"{if $product.quantity_wanted}{$product.quantity_wanted}{else}1{/if}",
-              "min"=>"{if $product.quantity_wanted}{$product.minimal_quantity}{else}1{/if}"
+              "id" => "quantity_wanted",
+              "value" => "{$product.minimal_quantity}",
+              "min" => "{$product.minimal_quantity}"
             ]
           }
         </div>
--- a/src/js/components/useQuantityInput.ts
+++ b/src/js/components/useQuantityInput.ts
@@ -278,8 +278,36 @@ const sendUpdateCartRequest = async (updateUrl: string, quantity: number) => {
   return response;
 };
 
+export const populateMinQuantityInput = (selector = quantityInputMap.default) => {
+  const qtyInputNodeList = document.querySelectorAll<HTMLElement>(selector);
+
+  // For each product in list
+  qtyInputNodeList.forEach((qtyInputWrapper: HTMLElement) => {
+    const idProduct = qtyInputWrapper.closest('form')
+      ?.querySelector<HTMLInputElement>(quantityInputMap.idProductInput)?.value;
+    const qtyInput = qtyInputWrapper.querySelector<HTMLInputElement>('input');
+    const qtyInputMin = qtyInput?.getAttribute('min');
+
+    // if the idproduct is set, and the input has a min attribute
+    if (idProduct && qtyInput && qtyInputMin) {
+      // we check if the product is already in the cart
+      const productInCart = window.prestashop.cart.products.filter(
+        (product: {id: number}) => product.id === parseInt(idProduct, 10),
+      ).shift();
+      // if the product is in the cart (and if the qty wanted is >= than the min qty, we set the minimal quantity to 1
+      const minimalQuantity = productInCart && productInCart.quantity_wanted >= qtyInputMin
+        ? 1 : qtyInputMin;
+      // we set the min attribute to the input
+      qtyInput.setAttribute('min', minimalQuantity.toString());
+      qtyInput.setAttribute('value', minimalQuantity.toString());
+    }
+  });
+};
+
 document.addEventListener('DOMContentLoaded', () => {
   const {prestashop, Theme: {events, selectors}} = window;
+
+  populateMinQuantityInput();
 
   prestashop.on(events.updatedCart, () => {
     useQuantityInput(cartSelectorMap.productQuantity);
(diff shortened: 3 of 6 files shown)


M0rgan01 (dismissed):

templates/catalog/_partials/product-add-to-cart.tpl
```diff
@@ -45,8 +45,8 @@
           {include file='components/qty-input.tpl'
             attributes=[
               "id"=>"quantity_wanted",
-              "value"=>"{if $product.quantity_wanted}{$product.quantity_wanted}{else}1{/if}",
-              "min"=>"{if $product.quantity_wanted}{$product.minimal_quantity}{else}1{/if}"
+              "value"=>"1",
+              "data-min"=>"{$product.minimal_quantity}"
```
SharakPL, line 49: Please change `=>` to ` => ` 😉

Why not use `min` attribute directly?

boherm: Oh yeah, that's a good remark!
I will change that to use the `min` attribute directly.

florine2623 (commented):
Hello @boherm ,

Tested with all types of products. 

Works well for standard and virtual products ! ✅ 

On a product with combination, it doesn't work as expected.
In the FO, the quantity minimum is always 1. There's an alert to tell the customer that there's a minimum quantity. :x:

https://github.com/PrestaShop/hummingbird/assets/16019289/88006a51-7875-42b0-b39c-2c81a7665893

Doesn't work on Pack product.  :x:
In the FO, the quantity is always the minimum quantity. After adding the product to my cart, the quantity is not updated.

https://github.com/PrestaShop/hummingbird/assets/16019289/081d6734-9289-4d3b-bd61-c2cf95f7f3a3

When I deleted the products in the cart and go back to category page, the minimum quantity is not updated directly. I have to add the product through my quickview page or
refresh again the page. :x:

https://github.com/PrestaShop/hummingbird/assets/16019289/d9c64300-b704-42a2-a60d-5c633c5f6528

Could you check ? ^^
Thanks!

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -185,9 +185,9 @@
                 <div class="quantity-button js-quantity-button w-100 w-sm-auto">
                   {include file='components/qty-input.tpl'
                     attributes=[
-                      "id"=>"quantity_wanted_{$product.id_product}",
-                      "value"=>"{if $product.cart_quantity && $product.cart_quantity >= $product.minimal_quantity}1{else}{$product.minimal_quantity}{/if}",
-                      "min"=>"{if $product.cart_quantity && $product.cart_quantity >= $product.minimal_quantity}1{else}{$product.minimal_quantity}{/if}"
+                      "id" => "quantity_wanted_{$product.id_product}",
+                      "value" => "1",
```
SharakPL, line 189: I'm not sure why you fixed the value to 1 instead of using `$product.minimal_quantity`. The same in other file.

boherm: For this, it's an issue with the data in cache before template generation.
The `cart_quantity`, that must be unique by users, is in cache too and for everybody! With your solution, if we do a `cache:clear` at any time, it's working well.
So yeah, it's a core issue, but force to 1 and add some javascript can do the trick as long as we fix this cache issue in the core.

boherm: In all cases, this isn't a fault to cache the product list 🤷‍♂️
But the `cart_quantity` don't be cached in a global state.

SharakPL: Could `{nocache}` help here?
https://www.smarty.net/docs/en/language.function.nocache.tpl

boherm: Unfortunately, it's seems not enough.

SharakPL:
IMO it would be better to use a different approach. Keep `min` exactly as it is set in product configuration and if the current product combination is already in the cart then the `value` should be updated to its amount in the cart and instead of the`Add to cart` button there would be `Update cart`. It would make all calculations obsolete. Just get the product config and its amount in the cart.

boherm:
Thanks for your comment @SharakPL, your approach with `Update cart` instead of `Add to cart` button seems cool, but I think that is a "feature" that need to be discuss together with @PrestaShop/product-council. I don't know if they had this possibility in mind.

florine2623 (commented):
Hello @boherm ,

Tested on branch develop on PS.

Tested on Product details page, Category related page, Homepage, Quickview and Search page.
Tested for standard product ✅ , combination :x: , virtual :x:, pack :x:, product with custom 🤷 .

The minimal quantity is always the same on the `Product Details page` for Combination, pack and virtual :x:
On other pages, it works as expected ✅ 

- Pack

https://github.com/PrestaShop/hummingbird/assets/16019289/065fd23c-d1e7-427b-a5dc-934a94fc125c

- Combination

https://github.com/PrestaShop/hummingbird/assets/16019289/7a6c01eb-79da-46d0-941d-f2928cc613c6

- Virtual

https://github.com/PrestaShop/hummingbird/assets/16019289/c8d397d9-5546-499a-b0be-365b06b31916

- For customized product, it doesn't work either, but I think that's a wanted behavior as the Customization might not be checked. So we can't know if if have 3 of the same product.
The issue is that I can't have 2 custom product with custom XX, and 1 other custom product with custom YY.
I guess that for @PrestaShop/product-council to decide ^^

https://github.com/PrestaShop/hummingbird/assets/16019289/14d567ff-1357-4019-9af6-d2e58d0de842

Thanks !
Waiting for your feedback !

boherm:
Thanks @florine2623 for your message!
After some tests, I have totally the same behaviour on my side.
But, they are not related to HB theme, and we have the same problems on classic theme.

In fact, this is a problem with the core here and after a small talk with @jolelievre, this could be fix in another PR and directly on the core side. 

What do you think? 🤔

florine2623:
Hello @boherm and @jolelievre ,

Ok for me to get it fixed on the core directly ^^
I'll let you decide on what to do with this PR ! I'll take off the Waiting for QA label.

Thanks!

templates/catalog/_partials/miniatures/product.tpl
jolelievre: I understand the cache problem, but then shouldn't the default value be `{$product.minimal_quantity}` It doesn't seem logic to have a value always to 1 when the min attribute under can be higher no?

It would mean the default cached page would use the min quantity for everyone and then your JS update handles updating/reducing it when relevant no?

templates/catalog/_partials/product-add-to-cart.tpl
```diff
@@ -44,9 +44,9 @@
         <div class="product-actions__quantity quantity-button js-quantity-button col-md-auto">
           {include file='components/qty-input.tpl'
             attributes=[
-              "id"=>"quantity_wanted",
-              "value"=>"{if $product.quantity_wanted}{$product.quantity_wanted}{else}1{/if}",
-              "min"=>"{if $product.quantity_wanted}{$product.minimal_quantity}{else}1{/if}"
+              "id" => "quantity_wanted",
+              "value" => "1",
```
jolelievre, line 48:
```suggestion
              "value" => "{$product.minimal_quantity}",
```
Same question why not use the min quantity for both min and value attributes?

matks (approved):

PR #627, opened 2024-05-17
Fix: issue-340

### LINKED ISSUE #340: Viewport check: 320px

tblivet:
| Questions         | Answers
| Description?      | Fix https://github.com/PrestaShop/hummingbird/issues/340
| BC breaks?        | no
| Deprecations?     | 
| Fixed ticket?     | Fix https://github.com/PrestaShop/hummingbird/issues/340
| How to test?      | ⬇️

Most of the points described in the issue have already been addressed. The remaining ones have been tackled as follows:
- Enhanced pagination to prevent overflow on mobile devices:
  image
- Added a Bootstrap breakpoint to provide more flexibility on small devices (use to improve product list on mobile).
- Updated product list display based on page layouts.
- Improved mobile display of the product list to feature only one product per row on mobile devices.
image

### DIFF
--- a/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
+++ b/modules/ps_bestsellers/views/templates/hook/ps_bestsellers.tpl
@@ -5,7 +5,7 @@
 <section class="best-sellers-products mt-3">
     <div class="container">
         {include file='components/section-title.tpl' title={l s='Best Sellers' d='Shop.Theme.Catalog'}}
-        {include file='catalog/_partials/productlist.tpl' products=$products productClass='col-6 col-lg-4 col-xl-3'}
+        {include file='catalog/_partials/productlist.tpl' products=$products productClass='col-12 col-xs-6 col-lg-4 col-xl-3'}
         <div class="best-sellers-products-footer text-center">
             <a class="all-product-link btn btn-outline-primary" href="{$allBestSellers}">
                 {l s='All best sellers' d='Shop.Theme.Catalog'}<i class="material-icons rtl-flip" aria-hidden="true">&#xE315;</i>
--- a/modules/ps_categoryproducts/views/templates/hook/ps_categoryproducts.tpl
+++ b/modules/ps_categoryproducts/views/templates/hook/ps_categoryproducts.tpl
@@ -10,6 +10,6 @@
       {include file='components/section-title.tpl' title={l s='%s other products in the same category' sprintf=[$products|@count] d='Shop.Theme.Catalog'}}
     {/if}
 
-    {include file='catalog/_partials/productlist.tpl' products=$products productClass='col-6 col-lg-4 col-xl-3'}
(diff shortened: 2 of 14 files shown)


nicosomb (dismissed):

Hlavtox (changes requested):
Please keep one column for mobile. In the issue, it was just a suggestion from the author.

tblivet:
> Please keep one column for mobile. In the issue, it was just a suggestion from the author.

Hi @Hlavtox, I don't really understand, you want me to remove one column for mobile?
I think it's a good optimization as it will concern only devices with screen width < 375. What do you think?

tblivet:
@Hlavtox, I think removing it is not a good idea because with 2 products per line on very small devices, we really lose in terms of visibility. As you say, we can fit more data, but we display more data in a bad way.
I suggest keeping this in place but reducing the breakpoint to 360.
With the previous value of 375, the new iPhones will display 2 products per line as it will trigger if < 375. If we lower the value to 360, we will also cover Galaxy, and some other phones, which have a width of 360 for their latest models. This way, one column will only display for small and older devices.
https://gs.statcounter.com/screen-resolution-stats/mobile/worldwide

I think it's a good compromise. What do you think?

Hlavtox:
@tblivet Yeah! Good solution. :-) Do you want to move the breakpoint or add another one?

tblivet:
@Hlavtox, Perfect 👍 No, just move it because it's already a new breakpoint!

nicosomb (approved):

Hlavtox (approved):

boherm (approved):

florine2623 (approved):
Hello @tblivet ,

Thanks for the PR.

When I try on desktop screen, the 3 dots button doesn't work :

https://github.com/PrestaShop/hummingbird/assets/16019289/a8704ec6-6de7-4737-8c33-288845bf32a4

On mobile 320px size, the pagination design is as expected. 

But it doesn't work so well. The 3-dots button doesn't always work.

https://github.com/PrestaShop/hummingbird/assets/16019289/89ec259d-2c96-4837-8c76-a3cb1a2b2888

Although this is not blocking, before this PR, it was already like this. It can be fixed later.

The products are well displayed on mobile size.

It is QA ✅

tblivet:
Thank you, @florine2623 🙏
The three dots are just there to indicate that there are pages between page X and page Y.
They allow us to keep all the pagination on a single line, instead of showing every page. So, it's normal if they do nothing, they're just visual elements.

Issue #581, opened 2024-01-25
Flag indicating the use of a new version of Bootstrap for developers


kpodemski:
We should add a flag to theme.yml and to the core, allowing the developer to check whether the theme is based on Bootstrap5. It is to distinguish the use of Classic vs. Hummingbird as a base of the theme. Why? Because it impacts front-end work from those developers.

I'm pretty sure I discussed it with @MatShir and @Hlavtox as an absolute must-have, but I cannot find an issue about it.

Hlavtox:
Yes, we should use a flag called `framework`.
This flag could be accessible out of the box by calling `$this->context->shop->theme->get('framework')`.

The only issue I see, is that the core is loading the config from the parsed json file in `config/themes` folder, which is not updated when updating the theme.

kpodemski:
> Second issue is that if you configure page layouts in backoffice, they get saved into this JSON. We will need to come up with a different place to store the layout config.

IMHO this is completely unrelated to the main point of this issue, which is the ability to distinguish bs5-based themes from bs4-based themes 👍🏻

Hlavtox:
@kpodemski I expanded on that topic because of

> The only issue I see, is that the core is loading the config from the parsed json file in config/themes folder, which is not updated when updating the theme.

So if you update a theme to use Boostrap 5 to Bootstrap 6, core will still tell you that this theme is using v5.

matks:
@kpodemski Something like this? https://github.com/PrestaShop/classic-theme/pull/149

Hlavtox:
One important point still left @matks @jolelievre, the core is caching the theme YML inside `config/themes/themename/shopX.json`.

Hlavtox:
@kpodemski What if you update the theme by uploading a new zip? (This is how would I do it.)

Hlavtox:
@kpodemski The problem is that page layout configurations are saved in that JSON, so when you clear it, it's fucked :D

jolelievre:
Hi guys here is the PR that introduces the modal based on the theme config https://github.com/PrestaShop/PrestaShop/pull/36731

The criteria to show the modal is if the theme IS bootstrap and is not version4 (maybe I should check if `>4` instead 🤔) The wording is not validated yet, and especially the doc URL is not correct For now it points to https://devdocs.prestashop-project.org/9/themes/hummingbird/ but it should be a dedicated page for bootstrap5 (with info about the compatibility layer)

jolelievre:
[...]

PR #649, opened 2024-08-20
Add meta framework into theme

matks:
| Questions         | Answers
| Description?      | Add meta framework into theme to allow module developers to find out what framework is used. See similar PR for Classic https://github.com/PrestaShop/classic-theme/pull/149
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes partially https://github.com/PrestaShop/hummingbird/issues/581
| Sponsor company   | 
| How to test?      |

### DIFF
--- a/config/theme.yml
+++ b/config/theme.yml
@@ -10,6 +10,7 @@ meta:
   compatibility:
       from: 8.1.0
       to: ~
+      framework: 'bootstrap-v5.2.0'
 
   available_layouts:
     layout-full-width:


Hlavtox:
I overlooked it in the second PR, do we want to put it into compatibility section? Isn't it more an general info about the theme?

kpodemski:
Maybe we could be less specific and use "bootstrap-v5.2" or even "bootstrap-v5", just not to force devs to update every time we update Bootstrap version, my understanding is that they do follow SemVer, so we should be safe.

jolelievre (approved):
Thanks @matks

ps-jarvis:
PR merged, well done!

kpodemski:
oh no @jolelievre 😅

matks:
@kpodemski Don't worry 😄 this is just a merged PR, not a "final act"

- code can evolve
- in the last resort, PRs can be reverted

I personally don't have a strong opinion on the content of that property, so every option is OK for me 👍

jolelievre:
Hi @kpodemski 

yeah I merged this early because I needed both themes to have the extra configuration to handle the specific modal, but we can always improve those two values before the next release don't worry.

I think having `bootstrap-v5.2` is better, I would even go further and integrate the patch version `bootstrap-v5.2.5` for example, the reason is because we can use semver comparison method to check this data.

If you look at this PR https://github.com/PrestaShop/PrestaShop/pull/36731 you can see I'm using some JS method to check the version so there's no need to split the version into major/minor/patch It's easier and more accurate to compare the whole version with semver rules like `satisfies(frameworkVersion, '^4.0.0')`

You can look into the JS library small docs here https://www.npmjs.com/package/compare-versions to see how more versatile it can be (`1.2.7 || >=1.2.9 <2.0.0`, `^10.0.0`, ...).

And if you think we also need this kind of comparison in twig and/or smarty I think we can add some functions to allow similar tests

kpodemski:
@jolelievre 

> And if you think we also need this kind of comparison in twig and/or smarty I think we can add some functions to allow similar tests

Yes, that would be great. Ultimately, we should be able to do such comparisons on the controller and view level.

jolelievre:
@kpodemski we already have this kind of tools in PHP thanks to composer https://github.com/composer/semver/blob/main/src/Semver.php#L32

We'd just need to add a twig and smarty functions to use it in the view

kpodemski:
@jolelievre ok, I thought we'd like to wrap it, so that we don't rely on composer's code directly 👍🏻

matks:
> Hi @kpodemski
> 
> yeah I merged this early because I needed both themes to have the extra configuration to handle the specific modal, but we can always improve those two values before the next release don't worry.
> 
> I think having `bootstrap-v5.2` is better, I would even go further and integrate the patch version `bootstrap-v5.2.5` for example, the reason is because we can use semver comparison method to check this data.
> 
> If you look at this PR [PrestaShop/PrestaShop#36731](https://github.com/PrestaShop/PrestaShop/pull/36731) you can see I'm using some JS method to check the version so there's no need to split the version into major/minor/patch It's easier and more accurate to compare the whole version with semver rules like `satisfies(frameworkVersion, '^4.0.0')`
> 
> You can look into the JS library small docs here https://www.npmjs.com/package/compare-versions to see how more versatile it can be (`1.2.7 || >=1.2.9 <2.0.0`, `^10.0.0`, ...).
> 
> And if you think we also need this kind of comparison in twig and/or smarty I think we can add some functions to allow similar tests

New PRs for themes
- https://github.com/PrestaShop/hummingbird/pull/651
- https://github.com/PrestaShop/classic-theme/pull/150

jolelievre:
@matks I don't know if you read my comment, but I don't understand the interest of your PRs?
We have tools that allow checking, comparing versions efficiently and very, so why not give as much detail as possible? 🤔

PR #651, opened 2024-08-28
Update theme.yml - better framework value

matks:
| Questions         | Answers
| Description?      | Update theme.yml - better framework value. See discussion [here](https://github.com/PrestaShop/hummingbird/pull/649#issuecomment-2301714452)
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/581
| Sponsor company   | 
| How to test?      |

### DIFF
--- a/config/theme.yml
+++ b/config/theme.yml
@@ -10,7 +10,7 @@ meta:
   compatibility:
       from: 8.1.0
       to: ~
-      framework: 'bootstrap-v5.2.0'
+      framework: 'bootstrap-v5.2'
 
   available_layouts:
     layout-full-width:


nicosomb (approved):

jolelievre (changes requested):
As explained in this comment https://github.com/PrestaShop/hummingbird/pull/649#issuecomment-2301714452

We can simply check a version using semver methods, so having the full version is only a plus I don't see why we wouldn't benefit from it

It doesn't mean all themes will respect this, maybe they will only add the minor version but it doesn't mean we can't use the full version

matks:
Following @jolelievre comment closing ;)

PR #662, opened 2024-11-27
In productlist: if price is not shown don't show add to cart either

### LINKED ISSUE #661: On product listing "add to cart" is shown even if prices are not.
When customer group has "show prices: no" set the product cannot be bought from the product page. The product listing still shows the add to cart buttons.

tswfi:
| Questions         | Answers
| Description?      | If customer group has "show prices: no" the product listing still shows the add to cart button. This adds the price check to the check if "add to cart" should be shown in the product listing
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     |no
| Fixed ticket?     | Fixes #661 
| Sponsor company   | 
| How to test?      | Change "guest" and "visitor" to "show prices: no". Check shop product listings that prices are not shown anymore and no "add to cart" button is shown. Login with a customer that belongs to a group that sees prices and check that the add to cart button is shown.

### DIFF
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -176,7 +176,7 @@
               {/block}
             </div>
 
-            {if $product.add_to_cart_url}
+            {if $product.show_price && $product.add_to_cart_url}
               <form action="{$urls.pages.cart}" method="post" class="d-flex flex-wrap flex-md-nowrap gap-3 align-items-center mt-3">
                 <input type="hidden" value="{$product.id_product}" name="id_product">
 


Hlavtox (changes requested):
This should be handled on LazyArray level. Do it here:

https://github.com/PrestaShop/PrestaShop/blob/2eecdda02daa262e6dcf760d1103abee38cbcded/src/Adapter/Presenter/Product/ProductLazyArray.php#L960

Hlavtox:
@tswfi No idea, but if we do it in core, it will work in all themes and will be centralized. 👍

Just do something like this:

```php
// Do not enable add to cart button if prices are hidden
if (!$this->shouldShowPrice($this->settings, $this->product)) {
  return false;
}
```

tswfi:
yup, core is the correct place.. the problem is that when searching for existing issues I get like 14 pages of results: https://github.com/PrestaShop/PrestaShop/issues?q=is%3Aissue%20state%3Aopen%20add%20to%20cart ...

PR #668, opened 2025-01-24
[IMPROVEMENTS - PART 1] Global quality improvement

tblivet:
| Questions         | Answers
| Description?      | This PR includes the first batch of quality improvements and refactoring of the theme, which contains:<br><br> - **Header refactoring**, including the integration of the following modules:<br> &nbsp;&nbsp;- `ps_contactinfo`<br> &nbsp;&nbsp;- `ps_searchbar`<br> &nbsp;&nbsp;- `ps_currencyselector`<br> &nbsp;&nbsp;- `ps_languageselector`<br> &nbsp;&nbsp;- `ps_customersignin`<br> &nbsp;&nbsp;- `ps_mainmenu`<br> &nbsp;&nbsp;- `ps_shoppingcart`<br> - The **ps_mainmenu** desktop display has been completely reworked.<br> - The **ps_searchbar** functionality has been improved.<br> - The product miniature component has been reworked.<br> - The “product grid” has been updated to use CSS Grid instead of multiple classes.<br> - The workflow has been updated to include Prettier for SCSS files.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      |


src/js/modules/ps_mainmenu.ts
```diff
@@ -1,105 +1,68 @@
...
-    dropdownToggleElement.addEventListener('keydown', (event: KeyboardEvent) => {
-      // Open the submenu when Arrow Right or Arrow Left key is selected
-      if (event.key === ARROW_LEFT_KEY || event.key === ARROW_RIGHT_KEY) {
-        dropdown.show();
-      }
-
-      // Hide the submenu when Escape key is pressed
-      if (event.key === ESCAPE_KEY) {
-        dropdown.hide();
+    /* Contrôle de l'affichage du sous-menu pour les événements de la souris à l'intérieur du conteneur du menu */
```
Hlavtox, line 7: French :-))

Hlavtox:
Hey @tblivet, I have written most of the code with Valentin, there is one thing I am not really happy about regarding product miniature.

We are doing column count in product list by passing boostrap classes into the templates, but this is just too much work and every module needs to do this.

I have come up with far simplier solution - https://danielhlavacek.cz/products.html, wdyt? It doesn't need to know anything about the context or container. :-)

Hlavtox (changes requested):
Good work, but I don't really like two things:
- You are moving a lot of things to `core` folder and deleting a lot of things from `custom` folder. The point of this was to separate the basic layout and our own custom design things. So developers can get a nice clean bootstrap base just by commeting out the custom folder. And not having to spend time with deleting things they dont need.
- There are container media queries used, I don't think the support is that big. Is it safe to use it?

src/js/modules/ps_mainmenu.ts
tblivet: Good catch 🙂

tblivet:
Hi @Hlavtox 👋 thanks for the review and feedback !

- Regarding the grid approach you proposed, I like the solution, but I don't think it’s the best fit for our case. I believe we need contextual control to better manage how many elements are displayed per line. This approach helps us avoid situations where a line is not fully populated with products, as shown in the screenshot below. What do you think?
products-line
- Yes, in fact, I have started moving files from `custom` to `core` because I don’t fully understand the benefit behind this separation. I believe it may lead contributors to confusion about where to write code. I’m not sure what you mean by **separating the basic layout from custom design**. From my perspective, it’s unclear, but maybe you can help me understand! Currently, when we comment out the import of the `custom.scss` file, the theme becomes completely non-functional.
- I think we can now use container queries, as they are currently supported by over 93% of browsers. Since Hummingbird is intended to be a modern theme and has not yet been officially released, I believe using container queries should be acceptable! We can also implement a fallback using the CSS `@supports` feature, but I’m not sure if it is necessary in this case, what do you think?

Hlavtox:
Hey @tblivet! :-)

> Regarding the grid approach you proposed, I like the solution, but I don't think it’s the best fit for our case. I believe we need contextual control to better manage how many elements are displayed per line. This approach helps us avoid situations where a line is not fully populated with products, as shown in the screenshot below. What do you think?

It does handle this case as well, see here: 
Snímek obrazovky 2025-01-27 123622

> Yes, in fact, I have started moving files from custom to core because I don’t fully understand the benefit behind this separation. I believe it may lead contributors to confusion about where to write code. I’m not sure what you mean by separating the basic layout from custom design. From my perspective, it’s unclear, but maybe you can help me understand! Currently, when we comment out the import of the custom.scss file, the theme becomes completely non-functional.

Check this article - https://build.prestashop-project.org/news/2022/new-theme-announce/, the point of this separation was, that it allows you to start creating much easier. I think that 90% of theme developers use classic theme as a base for their theme, and if you did this, you maybe spent more time deleting styles you don't want, other than building the new stuff. This way, you can just comment out `@import "custom/_custom";`, build it and voila, you have `somehow clean` boostrap layout. :-)

> I think we can now use container queries, as they are currently supported by over 93% of browsers. Since Hummingbird is intended to be a modern theme and has not yet been officially released, I believe using container queries should be acceptable! We can also implement a fallback using the CSS @supports feature, but I’m not sure if it is necessary in this case, what do you think?

I don't know what is the current standard, but I think we should aim more for 95 % to 97 % of support, it's ecommerce after all... I have only started using grid/flex gaps recently for example. But, no doubt that it's a cool feature that would allow to nicely style the product miniature sizing for example. :-)

ga-devfront (approved):

kpodemski (changes requested):
I am blocking to prevent accidental merge.

modules/ps_shoppingcart/ps_shoppingcart.tpl
```diff
@@ -2,22 +2,25 @@
...
-    {/if}
 
-    <i class="material-icons header-block__icon" aria-hidden="true">shopping_cart</i>
-    <span class="d-none d-md-flex header-block__title">{l s='Cart' d='Shop.Theme.Checkout'}</span>
-    <span class="header-block__badge">{$cart.products_count}</span>
+<div id="_desktop_ps_shoppingcart">
+  <div class="ps-shoppingcart">
+    <div class="header-block d-flex align-items-center blockcart cart-preview {if $cart.products_count> 0}header-block--active{else}inactive{/if}" data-refresh-url="{$refresh_url}">
+      {if $cart.products_count> 0}
+        <a class="header-block__action-btn pe-md-0" rel="nofollow" href="{$cart_url}" aria-label="{l s='View cart (%d products)' d='Shop.Theme.Checkout' sprintf=[$cart.products_count]}">
```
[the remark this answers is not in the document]
kpodemski: what do you mean? front office variables are automatically escaped

tblivet:
Hi @Hlavtox :)
- Sorry for the confusion, but this scenario is exactly what I wanted to avoid. A line of products that isn't fully populated can make the layout feel a bit buggy. However, I agree that an approach that doesn't require consideration of the page layout is beneficial, but in this case, I believe it’s not applicable. For example, by default, the module displays 8 products, so having 3 products per line on certain resolutions isn’t visually appealing, as it results in a row with only one product. A better alternative could be adding a carousel for modules that display products on the homepage, but that’s not the focus here.
> It does handle this case as well, see here: Snímek obrazovky 2025-01-27 123622

- Ok, but currently, when we comment out the `@import "custom/_custom";` file, the result on the develop branch doesn’t look like a clean Bootstrap layout to me. It may have been the case at the beginning of the project, but I don’t think so anymore? 
Homepage screenshot without `custom.scss`:
hummingbird-without-custom
> Check this article - https://build.prestashop-project.org/news/2022/new-theme-announce/, the point of this separation was, that it allows you to start creating much easier. I think that 90% of theme developers use classic theme as a base for their theme, and if you did this, you maybe spent more time deleting styles you don't want, other than building the new stuff. This way, you can just comment out `@import "custom/_custom";`, build it and voila, you have `somehow clean` boostrap layout. :-)
 
- From my point of view, this is acceptable, as customers generally use modern smartphones or laptops and are not stuck with outdated browsers, as is often the case in some companies. However, if anyone thinks it’s too early, I can implement a CSS `@supports` fallback for this feature.
> I don't know what is the current standard, but I think we should aim more for 95 % to 97 % of support, it's ecommerce after all... I have only started using grid/flex gaps recently for example. But, no doubt that it's a cool feature that would allow to nicely style the product miniature sizing for example. :-)

modules/ps_contactinfo/ps_contactinfo-rich.tpl
```diff
@@ -2,38 +2,38 @@
...
+    <i class="material-icons" aria-hidden="true">&#xF00F;</i>
+    <div class="ps-contactinfo__info">{$contact_infos.address.formatted nofilter}</div>
   </div>
   {if $contact_infos.phone}
     <hr/>
-    <div class="contact__item">
+    <div class="ps-contactinfo__item">
       <i class="material-icons" aria-hidden="true">&#xE0CD;</i>
-      <div class="contact__info"><a href="tel:{$contact_infos.phone}">{$contact_infos.phone}</a></div>
+      <div class="ps-contactinfo__info"><a href="tel:{$contact_infos.phone|escape:'htmlall':'UTF-8'}">{$contact_infos.phone|escape:'htmlall':'UTF-8'}</a></div>
```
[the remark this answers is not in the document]
tblivet: Hello @kpodemski 👋
This recommendation comes from: https://docs.cloud.prestashop.com/10-validation-checklist/#security
However, I can remove these occurrences in this PR and start a discussion in a separate PR about whether or not to include them in the entire project. Wdyt?

modules/ps_emailsubscription/views/templates/hook/ps_emailsubscription.tpl
```diff
@@ -14,7 +14,7 @@
 
       <div class="{$componentName}__content__right col-md-7">
         <form action="{$urls.current_url}#blockEmailSubscription_{$hookName}" method="post">
-          <div class="{$componentName}__content__inputs inline-items">
```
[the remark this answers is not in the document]
tblivet: Yes, the PR is already huge, which is why it doesn't include everything. However, I plan to redo this part as well in another PR. 👍

modules/ps_contactinfo/ps_contactinfo-rich.tpl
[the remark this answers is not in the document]
kpodemski: As far as I know, those are still relevant for the back office but not the front office. They were removed on purpose in 1.7 quite a long time ago.

Let me also ping @jolelievre, so he is also aware of this recommendation and share his opinion :)

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -6,11 +6,11 @@
 
 {block name='product_miniature_item'}
   <article
-    class="{$componentName} js-{$componentName}{if !empty($productClasses)} {$productClasses}{/if}"
```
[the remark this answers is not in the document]
tblivet: I have removed it to make the code more understandable and to handle the display of x products per line in the SCSS, avoiding overloading items with too many classes.

kpodemski:
> Ok, but currently, when we comment out the @import "custom/_custom"; file, the result on the develop branch doesn’t look like a clean Bootstrap layout to me.

Although it doesn't look like a clean design, it is a clean bootstrap, and that was its purpose: to serve as a skeleton for developers who dislike the default styling.

Hlavtox:
@tblivet 

> Sorry for the confusion, but this scenario is exactly what I wanted to avoid. A line of products that isn't fully populated can make the layout feel a bit buggy.

It's perfectly normal to me, in a real world, it just happens all the time. What would you want to do instead, make one product 3 columns wide? :-)

> However, I agree that an approach that doesn't require consideration of the page layout is beneficial, but in this case, I believe it’s not applicable. For example, by default, the module displays 8 products, so having 3 products per line on certain resolutions isn’t visually appealing, as it results in a row with only one product. A better alternative could be adding a carousel for modules that display products on the homepage, but that’s not the focus here.

I don't understand, I am keeping the same behavior as now - 3 products per row in left column layouts, 4 products per row in full width layouts. :-) If people don't want to have 1 products per line, they need to make sure to implement some conditions in their modules.

> Ok, but currently, when we comment out the @import "custom/_custom"; file, the result on the develop branch doesn’t look like a clean Bootstrap layout to me. It may have been the case at the beginning of the project, but I don’t think so anymore?
> Homepage screenshot without custom.scss:

Yeah, it doesn't look good. I wasn't active in this repo for a while. The original idea was to make it look something like Falcon - https://falcon-theme.dev/en/.

kpodemski:
@tblivet what about some other comments I made? :) if you want to move them to another PR then fine, but I think we lack of consistency with `{$componentName}` and hardcoded module names in the header are not good for the overall quality of the theme

tblivet:
@kpodemski I completely agree that a module should be handled through hooks and not directly inside the theme .tpl files, but that's not currently the case. What I’m proposing in this PR is a **"Part 1"** of the optimization. I believe there’s still a lot to do to improve the overall quality, but we can’t include everything in this single PR ?
> @tblivet what about some other comments I made? :) if you want to move them to another PR then fine, but I think we lack of consistency with `{$componentName}` and hardcoded module names in the header are not good for the overall quality of the theme

modules/ps_languageselector/ps_languageselector.tpl
```diff
@@ -2,11 +2,12 @@
...
-<div id="_desktop_language_selector">
-  <div class="language-selector__wrapper">
-    <select id="language-selector" aria-label="{l s='Language' d='Shop.Theme.Global'}" class="form-select js-language-selector">
+
+<div id="_desktop_ps_languageselector">
+  <div class="ps-languageselector">
+    <select class="form-select js-language-selector" aria-label="{l s='Language' d='Shop.Theme.Global'}">
       {foreach from=$languages item=language}
-        <option value="{url entity='language' id=$language.id_lang}"{if $language.id_lang == $current_language.id_lang} selected="selected"{/if} data-iso-code="{$language.iso_code}">{$language.name_simple}</option>
+        <option value="{url entity='language' id=$language.id_lang}"{if $language.id_lang == $current_language.id_lang} selected="selected"{/if} data-iso-code="{$language.iso_code}">{$language.iso_code|upper}</option>
```
Hlavtox, line 10: Why change name_simple here?

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -6,11 +6,11 @@
...
+    class="{$componentName} js-{$componentName}"
     data-id-product="{$product.id_product}"
     data-id-product-attribute="{$product.id_product_attribute}"
   >
-    <div class="card">
```
Hlavtox, line 13: Why do you remove all of the things we spent so much time on? The usage of card and card-body had their purpose, to automatically style it somehow...

modules/ps_languageselector/ps_languageselector.tpl
kpodemski: I think it may look better? :D

Hlavtox: Well, for czech the iso code is CS, which is not looking good. :-))) General public knows CZ, if something.

kpodemski: good point

templates/catalog/_partials/miniatures/product.tpl
kpodemski: that's true, it follows what has been done in Falcon and what Daniel mentioned before, if you "skip" the "custom" layer you will end up with a design that looks like a clean Boostrap theme

modules/ps_languageselector/ps_languageselector.tpl
tblivet: Yes, it was for styling purposes, but I have no problem reverting to `name_simple` if it’s more relevant for certain languages.

templates/catalog/_partials/miniatures/product.tpl
tblivet: @Hlavtox I have removed this because, in my opinion, `product-miniature` appears to be more of a fully custom component rather than a card component.

@kpodemski I understand, but currently, we don’t ended up with a clean Bootstrap theme. By "clean," you mean an unmodified Bootstrap version?

I believe this concept wasn’t clearly defined and shared among all contributors. If we look at the current implementation, if I understand well, we can see elements that don’t align with this vision?

### Example: Breadcrumb Component

1. Some SCSS variables from the Bootstrap breadcrumb are overridden in `abstract/variables/bootstrap`, which is good.
2. Then, we have `core/components/_breadcrumb.scss`, which seems to adapt the breadcrumb to our styling needs.
3. Finally, there's `custom/components/_breadcrumb.scss`, which, in this case, only contains `margin: 0`. However, I’m not sure if this is expected ? Should `custom/components` only contain components unrelated to Bootstrap?

### Another Example: Product Miniature

1. We have styles in `core/components/product/_product-miniature.scss`, which are related to custom classes defined in the template that handle product miniature.
2. We also have styles in `custom/components/product/_product-miniature.scss`, which again relate to custom classes in the same template.

In this case, I don’t really understand why both exist. Is there a specific reason for this? Or is it just, as I mentioned in another comment, that contributors are confused or unsure where to contribute?

Hlavtox: @tblivet Unfortunately, as with everything Prestashop SA invests in, the work was not finished and it was abandoned. @NeOMakinG left the company and I lost the will to work on it for free.

So yeah, I agree with you, it doesn't look like it now, but that was the vision. To provide developers with easy starter template. Something that will only contain the most basic styles.

So, for example, after commenting out the custom folder, you should get something like this https://getbootstrap.com/docs/5.3/examples/checkout/.

- Is it a style that makes the layout and makes thing work? = `core/`
- Is it a style that adds colors, rounded borders, margins, paddings, fonts etc.? = `custom/`

tblivet: @Hlavtox Thank you for this information, it's a bit clearer what was trying to be achieved. I think it's a great idea to provide developers with a clean base Bootstrap theme 👍 I also believe the design should have been entirely based on Bootstrap to achieve this, and I think it wasn't the case 😕

However, there's one thing I don't agree with: the use of `core/` and `custom/.`

I strongly believe that it creates confusion and slows down the integration process. Manually splitting layout and visual styles into two separate folders seems a bit inconvenient, doesn’t it?

Maybe `core/` should have been named `bootstrap/` to specifically style Bootstrap components, while `custom/` would be used only for PrestaShop specific custom components ?

ga-devfront: I think @tblivet 's solution is an excellent idea to clarify this part.

Hlavtox: What about `structure` and `styling` folders?

kpodemski (approved):

modules/ps_contactinfo/ps_contactinfo-rich.tpl
```diff
@@ -2,38 +2,38 @@
...
  *}
-<div class="contact__details">
-  <p class="h2 contact__title">{l s='Store information' d='Shop.Theme.Global'}</p>
-  <div class="contact__item">
-    <i class="material-icons" aria-hidden="true">&#xE55F;</i>
-    <div class="contact__info">{$contact_infos.address.formatted nofilter}</div>
+ <div class="ps-contactinfo">
+  <p class="h2 ps-contactinfo__title">{l s='Store information' d='Shop.Theme.Global'}</p>
+  <div class="ps-contactinfo__item">
+    <i class="material-icons" aria-hidden="true">&#xF00F;</i>
```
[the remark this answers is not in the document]
Hlavtox: @ga-devfront @kpodemski @tblivet I think that the reason to use the entity representations of the icons.

Using the Unicode entity instead of the icon name prevents the text from showing in search results AND flashing in default font before the font loads. This works because the browser renders it as a character, even if the correct font isn't applied yet.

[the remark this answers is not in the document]
kpodemski: @Hlavtox makes sense 👍🏻

tblivet:
> Hello @tblivet
> 
> If the plan to work on improvements is still in place (@ibahloul-ps ?), I would suggest you discuss with @PrestaShop/qa-functional not to test those PRs and agree on the complete tests after you finish improvements.

Hello @kpodemski, thank you for your message and approval.
I have a proposition for you, could you let me know what you think about it?
Do you think it would be better to create a new branch in the repo, something like `improvements`, and make all my PRs to this branch? This way, we can separate the "normal" updates from the improvements I’ve started introducing, which would make it easier to identify what has been done and might also be better for QA to test.

kpodemski:
@tblivet I love that idea 👍🏻

templates/catalog/_partials/miniatures/product.tpl
tblivet: @Hlavtox, from my point of view, I really think that simply separating `structure` and `styling` still causes confusion and might be one of the reasons we already have styles in the wrong place.

modules/ps_contactinfo/ps_contactinfo-rich.tpl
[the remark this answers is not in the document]
tblivet: @Hlavtox, I think it needs to be tested, but you're right, if using hex code prevents flashing text, it should be optimized from text to hex code!

tblivet:
**FYI:** As we briefly discussed with @kpodemski, I have created a new branch named `improvements` to serve as the branch for the enhancements I will propose for the theme. I have also updated the base branch for this PR to point to `improvements`.

ps-jarvis:
PR merged, well done!

PR #699, opened 2025-05-13
Add displayCartExtraProductInfo hook

jf-viguier:
| Questions         | Answers
| Description?      | Add displayCartExtraProductInfo hook
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/discussions/38692
| Sponsor company   | Creabilis.com
| How to test?      | See issue

### DIFF
--- a/templates/checkout/_partials/cart-detailed-product-line.tpl
+++ b/templates/checkout/_partials/cart-detailed-product-line.tpl
@@ -92,6 +92,8 @@
       </div>
     {/foreach}
 
+    {hook h='displayCartExtraProductInfo' product=$product}
+
     <div class="product-line__prices product-line__item">
       <div class="product-line__current">
         <span class="price">{$product.price}</span>


Hlavtox (approved):

ps-jarvis:
PR merged, well done!

nicosomb:
No QA here? 🤔 

@matks

Hlavtox:
@nicosomb Not needed, it's only a smarty hook and the syntax is identical to all other simmilar hooks. I verified it on a live store.

Hlavtox:
@nicosomb But thanks for keeping an eye on things, as always.

jolelievre:
Sorry, @Hlavtox but even though we agree there's little risk in this modification it's not up to you to unilaterally decide that QA is not needed and ignore this step by using your privileges to merge...

Remember, with great power comes great responsibility 😅 Just a kind warning: if you keep abusing those privileges, I'm afraid we'll be forced to revoke your merge authorisations

Hlavtox:
@jolelivre Jo, this topic again. :-) I’ve seen much more impactful PRs merged without QA and no warnings were issued. If we’re going to strictly enforce rules, I’m all for it — but it should be done consistently, not selectively.

ShaiMagal:
@Hlavtox @jolelievre @nicosomb 
I think that this PR does not need any QA.
**However, I am not against QA having to be done on every PR.**
But it would be good to define it clearly so that there are no mistakes. If we are not able to distinguish which PRs need QA and which do not, then why not just solve QA everywhere and only then do the merge after QA?

WDYT guys?
We need to stick together and not argue about someone "abusing" their powers.
**Let's be united and move PrestaShop forward together.**

matks:
@ShaiMagal @Hlavtox One simple solution that has been followed in the past is

1. We see a PR that seems eligible to be tested by a developer or does not require testing at all because CI tests it all
2. We ask a QA analyst whether he/she agrees with that opinion
3. QA analyst says "Yes, I agree, let's go" and PR can be merged without QA

At the end of the day QA analysts are the ones deciding whether or not they should test it 😉

jolelievre:
Yep, to put it simply we just need to ask the QA if its ok to merge it

PR #769, opened 2025-09-11
[A11Y][ps_searchbar] improve accessibility

ga-devfront:
| Questions         | Answers
| Description?      | improve ps_searchbar module accessibility
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | --

### DIFF
--- a/src/js/modules/ps_searchbar.ts
+++ b/src/js/modules/ps_searchbar.ts
@@ -89,10 +89,12 @@ const initSearchbar = () => {
 
           if (productLink && productTitle && productImage) {
             productLink.href = product.canonical_url;
+            productLink.id = "result_product_option_" + product.id_product.toString();
             productTitle.innerHTML = product.name;
-
+            
             if (product.cover) {
               productImage.src = product.cover.small.url;
+              productImage.alt = product.cover.legend;
             } else {
               productImage.innerHTML = '';
             }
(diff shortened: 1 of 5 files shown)


tblivet (changes requested):
- When the search is cleared using clear button with keyboard, focus is lost, so we need to reassign it to the search input.
- Add an `aria-label` to the product link to prevent screen readers from announcing both the image name and the link text, which would cause duplicate announcements.

src/js/modules/ps_searchbar.ts
```diff
@@ -94,28 +138,208 @@ const initSearchbar = () => {
...
+          // Make clear button tabbable when search results are shown
+          searchClear?.setAttribute('tabindex', '0');
+          searchDropdown?.classList.remove('d-none');
+          currentResultIndex = -1; // Reset navigation index
+
+          // Update ARIA expanded state
+          searchInput.setAttribute('aria-expanded', 'true');
+
+          // Add keyboard navigation to result links and make them non-tabbable
+          const resultLinks = searchResults.querySelectorAll<HTMLAnchorElement>('.ps-searchbar__result-link');
```
tblivet, line 303: I think we should add this inside selector map ?

tblivet (approved):

tblivet (approved):

AureRita (approved):
Hi @ga

Thank you for your PR, I tested it and it seems to works : 

- All interactive elements in the search component are navigable and usable with the keyboard. :heavy_check_mark: 
- Visible focus indicators appear on all interactive components. :heavy_check_mark: 
- The close button is usable via keyboard and includes an accessible name. :heavy_check_mark: 
- Role="search" is present on the appropriate container. :heavy_check_mark: 
- No visual or functional regressions occur on desktop or mobile devices. :heavy_check_mark: 

Because the PR seems to works as expected, It's QA :white_check_mark:  

Thank you

ps-jarvis:
PR merged, well done!

PR #792, opened 2025-10-20
Fix the page position after submitting email subscription form

SharakPL:
| Questions         | Answers
| Description?      | When the email subscription form is submitted from the bottom of the page:<br>obraz<br>... it tries to return the page at `#blockEmailSubscription_{$hookName}` but such id doesn't exist in the template. This PR fixes it.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| Sponsor company   | ~
| How to test?      |  1. Subscribe with a valid email<br>  2. Notice the page reloads with specific `#` pointer in the URL, but stays on top nevertheless, preventing the user from seeing the feedback right away.<br>  3. Apply this PR and repeat 1-2. This time page will position itself at the newsletter block with the feedback alert.

### DIFF
--- a/modules/ps_emailsubscription/views/templates/hook/ps_emailsubscription.tpl
+++ b/modules/ps_emailsubscription/views/templates/hook/ps_emailsubscription.tpl
@@ -5,7 +5,7 @@
 
 {$componentName = 'email-subscription'}
 
-<div class="{$componentName} px-0 py-4">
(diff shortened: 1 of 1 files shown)


ga-devfront:
Yes, we do need to remove it, thank you for pointing that out.

Issue #734, opened 2025-07-08
Mobile Search bar fault


MerseyRay:
Im mobile view, upon page load, the following is present

<img width="436" height="56" alt="Image" src="https://github.com/user-attachments/assets/d8acd6c1-f5ef-46b4-84be-c9f8bd3e19b0" />

Then when a search is performed and jquery fills up the results we get this

<img width="432" height="223" alt="Image" src="https://github.com/user-attachments/assets/ba54720f-2b48-46c7-97ef-b594978332f6" />

But if the cancel button is pressed, the results are removed, but also the ul container is removed, so the search will not work again until a page reload has happened.

<img width="456" height="21" alt="Image" src="https://github.com/user-attachments/assets/927702df-9fd2-4561-be0d-6b9248f73b0f" />

tblivet:
Resolved in Hummingbird v2 ✅

PR #796, opened 2025-10-25
On mobile: focus on searchinput on offcanvas open

tswfi:
| Questions         | Answers
| Description?      | On mobile view focus on the searchinput when the offcanvas search is shown. Also clear the search when the offcanvas is hidden
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Touches on issue #734 (the actual issue was already fixed before but when testing I noticed that this could be improved a bit)
| Sponsor company   | 
| How to test?      | On mobile view click on the search icon to get the offcanvas search to show, the focus should be on the input field and if a mobile phone is used the keyboard is shown. If input field contains text when closing the view then reopening the view should have the old search cleared and the focus again for a new search.

Note: someone with better understanding on the offcanvas functionality could chime in on if the timeout is even necessary or not :)

### DIFF
--- a/src/js/modules/ps_searchbar.ts
+++ b/src/js/modules/ps_searchbar.ts
@@ -98,6 +98,16 @@ const initSearchbar = () => {
     }
   });
 
+  // Focus on searchField on mobile canvas open
+  searchCanvas?.addEventListener('shown.bs.offcanvas', () => {
+    if (searchInput) {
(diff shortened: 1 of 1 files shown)


tswfi:
rebasing this is not practical. opened a new pr https://github.com/PrestaShop/hummingbird/pull/805

PR #797, opened 2025-10-25
Poweruser search, use / to focus on searchinput

tswfi:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Poweruser search. Use `/` to focus on the searchInput field
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | 
| How to test?      | Open a page and type `/` , check that search input is focused (and on mobile the offCanvas is open) and ready for searching. Check that the / can still be typed on input fields elsewhere (the search focus should only happen if you are not in a input field already)

PR #806, opened 2025-11-10
Poweruser search, use / to focus on searchinput

tswfi:
| Questions         | Answers
| Description?      | Poweruser search. Use / to focus on the searchInput field
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | 
| How to test?      | Open a page and type / , check that search input is focused (and on mobile the offCanvas is open) and ready for searching. Check that the / can still be typed on input fields elsewhere (the search focus should only happen if you are not in a input field already)

this replaces pr #797

### DIFF
--- a/src/js/modules/ps_searchbar.ts
+++ b/src/js/modules/ps_searchbar.ts
@@ -29,9 +29,9 @@ const initSearchbar = () => {
   const searchClear = document.querySelector<HTMLElement>(SearchBarMap.searchClear);
   const searchUrl = searchWidget?.dataset.searchControllerUrl;
 
-  // For powerusers `/` goes to search :)
-  // TODO: this should probably be in a totally new "shortcuts" module that does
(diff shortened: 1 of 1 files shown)


nicosomb:
FYI, this shortcut is used on some browsers (for example, Firefox) to open the browser's search form.

tswfi:
> FYI, this shortcut is used on some browsers (for example, Firefox) to open the browser's search form.

the browser probably wins there..

I wonder if there is a best practice for this defined somewhere...

ga-devfront:
However, ` ctrl ` **+** ` k ` opens the search function in the **Chrome browser** (same as ` / ` for **Firefox**).
I completely understand the PR's intention, which I think is very good, but I don't know if it's going to backfire...

Hlavtox:
I think I can call myself a poweruser, I navigate most of everything by keyboard shortcuts, but I have never heard or used of a slash for search.

Search is CTRL+F. :-)

PR #810, opened 2025-11-12
Update rating based on the future release of the module.

### LINKED ISSUE #809: Product comments gets hidden after js refresh - bad div?
This button "Write your opinion" gets removed after changing a combination for example. I think the hook could be in a wrong div.

Image

tblivet:
| Questions         | Answers
| Description?      | Update rating based on the future release of the module productcomments.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | Should use `dev` version of `productcomments` module to test the fix: https://github.com/PrestaShop/hummingbird/issues/809

### DIFF
--- a/src/js/modules/productcomments.ts
+++ b/src/js/modules/productcomments.ts
@@ -799,6 +799,7 @@ class ProductListReviews {
 
     prestashop.on(events.updatedProduct, () => {
       this.loadProductListReviews();
+      ProductCommentsRating.initProductRatingSystem();
     });
   }
 


ga-devfront (approved):

boherm (approved):

AureRita (approved):
Hi @tblivet 

Thank you for your PR, I tested it and it seems to works as you can see, we still have correctly the stars for each combination : 

https://github.com/user-attachments/assets/9b37b6d3-42cf-45c0-9328-bf3b16bd4052



Tested on : 
9.0.x 
develop

Thank you

PR #897, opened 2026-01-12
Feat add missing hook displayAddressSelectorBottom

tblivet:
| Questions         | Answers
| Description?      | Feat add missing hook displayAddressSelectorBottom
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | --

### DIFF
--- a/templates/checkout/_partials/steps/addresses.tpl
+++ b/templates/checkout/_partials/steps/addresses.tpl
@@ -116,5 +116,14 @@
         {/if}
       </div>
     </form>
+
+    {capture name="address_selector_bottom"}{hook h='displayAddressSelectorBottom'}{/capture}
+    {if $smarty.capture.address_selector_bottom}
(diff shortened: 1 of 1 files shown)


ga-devfront (approved):

Hlavtox:
Ahhh, ok, I see. No need to worry about margins for example.

But! Watch out with this in case of the JS refreshed blocks by core.js, there must always be (for example) `<div class="product-discounts"></div>` even if empty.

tblivet:
@Hlavtox Yes, in some cases, due to how the core refreshes data, removing some blocks cannot be done.

ingridusta (approved):
Hi @tblivet, 

Your bug fix works perfectly. The hook is added : 

Capture d’écran 2026-01-13 à 17 39 15

Capture d’écran 2026-01-13 à 17 39 38

It's QA approved ✅

PR #916, opened 2026-01-27
Converted UPC to gtin13 in product schema

mgielecinski:
| Questions         | Answers
| Description?      | Converted UPC to gtin13 by adding leading 0
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      |

### DIFF
--- a/templates/_partials/microdata/product-jsonld.tpl
+++ b/templates/_partials/microdata/product-jsonld.tpl
@@ -35,7 +35,7 @@
     "sku": "{if $product.reference}{$product.reference}{else}{$product.id}{/if}",
     "mpn": "{if $product.mpn}{$product.mpn}{elseif $product.reference}{$product.reference}{else}{$product.id}{/if}"
     {if $product.ean13},"gtin13": "{$product.ean13}"
-    {elseif $product.upc},"gtin13": "0{$product.upc}"
+    {elseif $product.upc},"gtin12": "{$product.upc}"
     {/if}
     {if isset($product_manufacturer) && $product_manufacturer->name},
     "brand": {
@@ -80,7 +80,7 @@
       {/if}
       "sku": "{if $product.reference}{$product.reference}{else}{$product.id}{/if}",
       "mpn": "{if $product.mpn}{$product.mpn}{elseif $product.reference}{$product.reference}{else}{$product.id}{/if}",
-      {if $product.ean13}"gtin13": "{$product.ean13}",{elseif $product.upc}"gtin13": "0{$product.upc}",{/if}
+      {if $product.ean13}"gtin13": "{$product.ean13}",{elseif $product.upc}"gtin12": "{$product.upc}",{/if}
       {if !empty($product.show_condition) && !empty($product.condition)}"itemCondition": "{$product.condition.schema_url}",{/if}
       "availability": "{$product.seo_availability}",
       "seller": {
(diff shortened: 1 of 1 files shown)


templates/_partials/microdata/product-jsonld.tpl
```diff
@@ -35,7 +35,7 @@
     "sku": "{if $product.reference}{$product.reference}{else}{$product.id}{/if}",
     "mpn": "{if $product.mpn}{$product.mpn}{elseif $product.reference}{$product.reference}{else}{$product.id}{/if}"
     {if $product.ean13},"gtin13": "{$product.ean13}"
-    {elseif $product.upc},"gtin13": "{$product.upc}"
+    {elseif $product.upc},"gtin13": "0{$product.upc}"
```
tblivet: Hi @mgielecinski, I’m not an expert on product codes, but it seems we can use gtin12 instead. This appears to be valid according to Schema.org: https://schema.org/Product

```suggestion
    {elseif $product.upc},"gtin12": "{$product.upc}"
```

tblivet (commented):
Hi @mgielecinski, I left a comment here, @kpodemski or @Hlavtox do you have an opinion on this?

templates/_partials/microdata/product-jsonld.tpl
```diff
@@ -80,7 +80,7 @@
       {/if}
       "sku": "{if $product.reference}{$product.reference}{else}{$product.id}{/if}",
       "mpn": "{if $product.mpn}{$product.mpn}{elseif $product.reference}{$product.reference}{else}{$product.id}{/if}",
-      {if $product.ean13}"gtin13": "{$product.ean13}",{elseif $product.upc}"gtin13": "0{$product.upc}",{/if}
+      {if $product.ean13}"gtin13": "{$product.ean13}",{elseif $product.upc}"gtin12": "{$product.upc}",{/if}
```
tblivet: Sorry, I hadn’t noticed that the code was in two different places. I believe sku, mpn, ean13, and upc inside the Offers type can be removed, as they are duplicates.

tblivet (changes requested):

templates/_partials/microdata/product-jsonld.tpl
```diff
@@ -34,8 +34,8 @@
     {if !empty($product.cover)}"image" :"{$product.cover.bySize.home_default.url}",{/if}
     "sku": "{if $product.reference}{$product.reference}{else}{$product.id}{/if}",
     "mpn": "{if $product.mpn}{$product.mpn}{elseif $product.reference}{$product.reference}{else}{$product.id}{/if}"
-    {if $product.ean13},"gtin13": "{$product.ean13}"
-    {elseif $product.upc},"gtin13": "{$product.upc}"
+    {if $product.ean13},"gtin13": "{$product.ean13}"{/if}
+    {if $product.upc},"gtin12": "{$product.upc}"{/if}
     {/if}
```
tblivet: @mgielecinski Thanks for the changes, but you forgot to remove a {/if}.
```suggestion
```

tblivet (changes requested):

tblivet (approved):
Perfect thank you 👍

mgielecinski:
I made changes but I still think that gtin13 should be changed to gtin because in the ean13 field we can enter a 14-digit code

tblivet (changes requested):
Hmm, you’re right, let’s go with `gtin` for `ean13`. Sorry for the back and forth 😕

tblivet (approved):
Thank you again 🙂

AureRita (approved):
Hi @mgielecinski 

Image

Thank you

Issue #839, opened 2019-07-10
Display stock label in the cart


omar2886:
Hi,
I remember in PS 1.6 that before proceeding to check-out, customers can first see the complete cart and there was a sign clearly marking for each product wether it was in stock or not.
Now in PS 1.7 we only see in the cart (photo, title, qty, price and delete button). In the entire checkout process customer can't explicitly see wether all products are actually available in stock or not. I'm having problems with customers not being aware of this and asking for refunds, as the delay for orders containing products that we don't have immediate availability is considerably longer.
Regards,

---
# Display stock label in the cart
## Specs
If stock management is enabled in BO: in FO, in the cart, display the stock label according to the case for each product/combination:
- Icon in green #70B580 + Label of in-stock products in green #5A9166 or
- Icon in orange #FF5E24 + Label of out-of-stock products with allowed backorders in orange #FF5E24 or 
- Icon in red #F44336 + Label of out-of-stock products with denied backorders in red #C3362B

[Detail of specs and mockups](https://github.com/PrestaShop/prestashop-specs/blob/master/content/1.7/back-office/shop-parameters/product-settings/page.md)

Hlavtox:
@tblivet @ga-devfront Guys, could this be done for the final version of hummingbird? It could step up the game once more. :-)

Category
<img width="1027" height="750" alt="Image" src="https://github.com/user-attachments/assets/0973d116-d2df-47b1-b6ac-4bbdbf7fbb79" />

Cart
<img width="973" height="400" alt="Image" src="https://github.com/user-attachments/assets/3be601d0-96d7-4805-a2e8-e918b6b67a96" />

Full design - https://www.figma.com/design/aCBf2HlypjT0ZxP2XcsT68/Core-Github-issues?node-id=207-4875&p=f&t=UZLZCQIsAKzcxiPT-0

tblivet:
Hi @Hlavtox, I took some time to create a small POC for stock information in the cart. Here is more information: https://github.com/PrestaShop/hummingbird/pull/919

PR #919, opened 2026-01-30
Add stock informations on cart page

tblivet:
| Questions         | Answers
| Description?      | Add stock informations on cart page + refactor delivery informations 
| BC breaks?        | --
| Deprecations?     | --
| Fixed ticket?     | https://github.com/PrestaShop/hummingbird/issues/839
| How to test?      | --

**NOTE:** this PR should be test with https://github.com/PrestaShop/PrestaShop/pull/40653 and https://github.com/PrestaShop/PrestaShop/pull/40673 and https://github.com/PrestaShop/PrestaShop/pull/40662

This PR is only a POC to display stock information on the cart page. Core-side changes will be required to properly implement it.

This is a preview of what could be achieved if the core is adapted. I have identified some issues with error messages returned by the core. For example, when a product in the cart is no longer in stock, the current message is something like:
"You can only add 0 qty of this product."

A more appropriate message would be something like:
"Please remove the product (XXX) from the cart." when the available quantity is ≤ 0.

Capture d’écran 2026-01-30 à 11 30 45

### DIFF
--- a/templates/catalog/_partials/product-add-to-cart.tpl
+++ b/templates/catalog/_partials/product-add-to-cart.tpl
@@ -41,7 +41,6 @@
               {/if}
             </div>
           </div>
-
         {/if}
 
         {block name='product_delivery_times'}
--- a/templates/catalog/_partials/product-add-to-cart.tpl
+++ b/templates/catalog/_partials/product-add-to-cart.tpl
@@ -12,7 +12,7 @@
           hidden
         {/if}
       >
-        {if $product.show_availability && $product.availability_message}
+        {if !empty($product.availability_message)}
           {** First, we prepare the icons and colors we want to use *}
           {if $product.availability == 'in_stock'}
             {assign 'availability_icon' 'E5CA'}
--- a/templates/catalog/_partials/product-add-to-cart.tpl
+++ b/templates/catalog/_partials/product-add-to-cart.tpl
@@ -5,11 +5,14 @@
 {if !$configuration.is_catalog}
   <div class="product__add-to-cart-container product-add-to-cart js-product-add-to-cart">
     {block name='product_availability'}
-      {if $product.show_availability && $product.availability_message}
-        <div
-          id="product-availability"
-          class="product__availability js-product-availability"
-        >
+      <div
+        id="product-availability"
+        class="product__availability js-product-availability"
+        {if empty($product.availability_message) && empty($product.delivery_information)}
+          hidden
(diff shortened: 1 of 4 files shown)


templates/catalog/_partials/product-add-to-cart.tpl
```diff
@@ -5,11 +5,14 @@
...
     {block name='product_availability'}
-      {if $product.show_availability && $product.availability_message}
-        <div
-          id="product-availability"
-          class="product__availability js-product-availability"
-        >
+      <div
+        id="product-availability"
+        class="product__availability js-product-availability"
+        {if empty($product.availability_message) && empty($product.delivery_information)}
```
Touxten: There is the same problem: `hidden` no longer displays delivery times if messages about stock are not displayed.

I also think we can simplify this with a single condition.

{if empty($product.availability_message) && empty($product.delivery_information)}

{if $product.show_availability && $product.availability_message}

ga-devfront (dismissed):

Touxten (dismissed):

Hlavtox (approved):
😍😍😍

templates/checkout/_partials/cart-detailed-product-line.tpl
```diff
@@ -91,6 +91,35 @@
...
+        {assign 'availability_class' 'text-warning'}
+      {elseif $product.availability == 'last_remaining_items'}
+        {assign 'availability_icon' 'E002'}
+        {assign 'availability_class' 'text-warning'}
+      {else}
+        {assign 'availability_icon' 'E14B'}
+        {assign 'availability_class' 'text-danger'}
+      {/if}
+
+      {if $product.show_availability && $product.availability_message}
```
Hlavtox: `show_availability` not needed. It's already handled by the core. :-)

Hlavtox (changes requested):
One last tweak, I will reapprove immediately 🙏

And one question @tblivet @ga-devfront - guys, why is there the condition to put `hidden`? Isn't the block automatically rendered and replaced as a whole? Cannot we just not render it at all if both availability and delivery message is missing?

Hlavtox (approved):

AureRita (approved):
Hi @

Image

https://github.com/user-attachments/assets/ee1af959-1061-4b4f-9f08-79103a34e896

Thank you

ps-jarvis:
PR merged, well done!

PR #926, opened 2026-02-04
Refactor quantity inputs following recent CORE changes

### LINKED ISSUE #915: Logic that manually handled min quantity in inputs can be now removed
cc @tblivet @ga-devfront Hello guys, I have some technical tasks that should be done after last PRs fixing availabilities in the core.

### Logic that manually altered quantity inputs with some custom logic
After last PRs, the logic for handling minimal quantity, required quantity, wanted quantity is now properly handled in the core and passed to the template. The input now receives proper minimal required quanity (after reducing cart items) directly from the core - for both listing and product page.

You can now completely remove many of the logic in https://github.com/PrestaShop/hummingbird/blob/develop/src/js/product.ts
and
https://github.com/PrestaShop/hummingbird/blob/develop/src/js/components/useQuantityInput.ts

Whole populateMinQuantityInput can go for example.

### quantity_wanted from url is not working - but will after previous point

Try that if you add ?quantity_wanted=10 to the end of URL on product page, it works on classic, but doesnt on hummingbird (because of the logic in previous point). It affects some parts of the page, like availability, but not the quantity input.

I think it will work after the previous point is fixed.

tblivet:
| Questions         | Answers
| Description?      | Refactor quantity inputs following recent CORE changes
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/hummingbird/issues/915
| How to test?      | Should be tested with the recent changes added to the CORE. https://github.com/PrestaShop/PrestaShop/pull/40538 **NOTE for QA:** Check that quantity selection and updates work correctly in different cases, such as minimum quantity...

### DIFF
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -84,7 +84,7 @@
                   {include file='components/qty-input.tpl'
                     attributes=[
                       "id" => "quantity_wanted_{$product.id_product}",
-                      "value" => "{$product.quantity_required}",
+                      "value" => "{$product.quantity_wanted}",
                       "min" => "{$product.quantity_required}"
                     ]
                   }
--- a/src/js/components/useQuantityInput.ts
+++ b/src/js/components/useQuantityInput.ts
@@ -322,41 +322,8 @@ const sendUpdateCartRequest = async (updateUrl: string, quantity: number) => {
   return response;
 };
 
-export const populateMinQuantityInput = (selector = quantityInputMap.default) => {
-  const qtyInputNodeList = document.querySelectorAll<HTMLElement>(selector);
-
-  const products = window.prestashop?.cart?.products;
-
-  if (!Array.isArray(products) || products.length === 0) {
-    return;
-  }
-
-  qtyInputNodeList.forEach((qtyInputWrapper: HTMLElement) => {
-    const form = qtyInputWrapper.closest('form');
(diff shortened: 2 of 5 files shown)


templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -84,7 +84,7 @@
                   {include file='components/qty-input.tpl'
                     attributes=[
                       "id" => "quantity_wanted_{$product.id_product}",
-                      "value" => "{$product.quantity_required}",
+                      "value" => "{$product.quantity_wanted}",
```
Hlavtox: This will never work anyway, it will always resolve to `quantity_required`.

Before https://github.com/PrestaShop/PrestaShop/pull/40570, it wasn't even reliable, that's why I just used `quantity_required`.

Hlavtox (commented):
Testing right now

Hlavtox (approved):
Works perfectly, even quantity wanted from URL works now.

AureRita (approved):
Hi @

https://github.com/user-attachments/assets/edcc0ceb-b4d8-4c01-bd50-2ec7d06ff572

https://github.com/user-attachments/assets/8c95ad23-a825-4f16-8ca5-5600ca85332c

https://github.com/user-attachments/assets/5c99dac6-2eba-4619-a5ba-acb8112f0277

Thank you

ps-jarvis:
PR merged, well done!

PR #936, opened 2026-02-18
Accessibility: add some `aria-hidden`, `label`s...

yannicka:
| Questions         | Answers
| Description?      | Improve accessibility. Add label, add aria-hidden, add aria-label, etc.
| Type?             | improvement (on accessibility)
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| Sponsor company   | N/A
| How to test?      | N/A

### DIFF
--- a/templates/catalog/_partials/sort-orders.tpl
+++ b/templates/catalog/_partials/sort-orders.tpl
@@ -2,9 +2,9 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  *}
-<label for="sort_dropdown_button" class="products__sort-label">
-  {l s='Sort by:' d='Shop.Theme.Global'}
-</label>
+<div class="products__sort-label">
+  <span class="align-middle">{l s='Sort by:' d='Shop.Theme.Global'}</span>
+</div>
 
 <div class="products__sort-dropdown">
   <button
--- a/templates/catalog/_partials/sort-orders.tpl
+++ b/templates/catalog/_partials/sort-orders.tpl
@@ -2,9 +2,9 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  *}
-<div class="products__sort-label">
-  <span class="align-middle">{l s='Sort by:' d='Shop.Theme.Global'}</span>
-</div>
+<label for="sort_dropdown_button" class="products__sort-label">
+  {l s='Sort by:' d='Shop.Theme.Global'}
+</label>
 
 <div class="products__sort-dropdown">
   <button
--- a/templates/customer/password-email.tpl
+++ b/templates/customer/password-email.tpl
@@ -28,10 +28,10 @@
 
     <section class="form-fields">
       <div class="mb-3">
-        <label class="form-label required">{l s='Email address' d='Shop.Forms.Labels'}</label>
-        <input type="email" name="email" id="email" value="{if isset($smarty.post.email)}{stripslashes($smarty.post.email)}{/if}" class="form-control" autocomplete="email" required>
+        <label for="email" class="form-label required">{l s='Email address' d='Shop.Forms.Labels'}</label>
+        <input type="email" name="email" id="email" value="{if isset($smarty.post.email)}{stripslashes($smarty.post.email)}{/if}" class="form-control" autocomplete="username" required>
       </div>
-      
+
       <div class="buttons-wrapper">
         <button id="send-reset-link" class="btn btn-primary" name="submit" type="submit" data-ps-action="form-validation-submit">
           {l s='Send reset link' d='Shop.Theme.Actions'}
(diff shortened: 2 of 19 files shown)


ps-jarvis:

- `Modules.Emailsubscription.Shop`
    - [ ] `Subscribe to our newsletter`
- `Shop.Theme.Catalog`
    - [ ] `Search products...`
    - [ ] `Clear search`
- `Shop.Theme.Global`
    - [ ] `Product availability:`
    - [ ] `Open zoomed product image gallery`

yannicka:
I also have a few questions:

1. Is `tabindex="0"` a good practice on alerts? For example:
  https://github.com/PrestaShop/hummingbird/blob/fd25e7f98d19924d671ecb9b7a8b26ab421606c2/templates/_partials/notifications.tpl#L11

2. Are `aria-label` useful on contact info? For example:
  https://github.com/PrestaShop/hummingbird/blob/fd25e7f98d19924d671ecb9b7a8b26ab421606c2/modules/ps_contactinfo/ps_contactinfo-rich.tpl#L47

templates/catalog/_partials/sort-orders.tpl
```diff
@@ -2,9 +2,9 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  *}
-<div class="products__sort-label">
-  <span class="align-middle">{l s='Sort by:' d='Shop.Theme.Global'}</span>
-</div>
+<label for="sort_dropdown_button" class="products__sort-label">
+  {l s='Sort by:' d='Shop.Theme.Global'}
+</label>
```
tblivet: Even though `<button>` can be associated with a `<label>`, that pattern is mainly meant for form inputs. I think we should keep the file as it is.

templates/customer/password-email.tpl
```diff
@@ -28,10 +28,10 @@
 
     <section class="form-fields">
       <div class="mb-3">
-        <label class="form-label required">{l s='Email address' d='Shop.Forms.Labels'}</label>
-        <input type="email" name="email" id="email" value="{if isset($smarty.post.email)}{stripslashes($smarty.post.email)}{/if}" class="form-control" autocomplete="email" required>
+        <label for="email" class="form-label required">{l s='Email address' d='Shop.Forms.Labels'}</label>
+        <input type="email" name="email" id="email" value="{if isset($smarty.post.email)}{stripslashes($smarty.post.email)}{/if}" class="form-control" autocomplete="username" required>
```
tblivet, line 32: Is there a specific reason for using `autocomplete="username"` here instead of email?

tblivet (changes requested):
Hi @yannicka, thanks for the PR! Here’s some feedback.

tblivet:
> I also have a few questions:
> 
> 1. Is `tabindex="0"` a good practice on alerts? For example:
>    https://github.com/PrestaShop/hummingbird/blob/fd25e7f98d19924d671ecb9b7a8b26ab421606c2/templates/_partials/notifications.tpl#L11
> 2. Are `aria-label` useful on contact info? For example:
>    https://github.com/PrestaShop/hummingbird/blob/fd25e7f98d19924d671ecb9b7a8b26ab421606c2/modules/ps_contactinfo/ps_contactinfo-rich.tpl#L47

Hi @yannicka,
1. I believe tabindex="0" is appropriate in this case because the notification template may display errors or important messages. Without it, the alert wouldn’t be focusable via the Tab key, which could make it harder for keyboard and screen reader users to access.
2. Yes, this aria-label is used to clearly indicate that the link is intended to send an email.

templates/customer/password-email.tpl
yannicka: "email" is generic while "username" is intended to autocomplete the field with the username (which can be an email address).

templates/catalog/_partials/sort-orders.tpl
yannicka: I think you're right, I'll see about cancelling that.

templates/customer/password-email.tpl
tblivet: I’m fine with this approach. When you have a moment, could you also update the field on the login form of the Sign In page to keep everything consistent? Thanks a lot.

yannicka: Here it is! :)
https://github.com/PrestaShop/PrestaShop/pull/40849

tblivet (changes requested):
Thank you for the update! It’s okay for me, but we need to rebase here. Yesterday, I changed the target branch to `2.x`

yannicka:
> Thank you for the update! It’s okay for me, but we need to rebase here. Yesterday, I changed the target branch to `2.x`

Can you do that?

tblivet (dismissed):

tblivet (approved):

ingridusta (approved):
Hi @yannicka, 

I have tested your pr. No regression on keyboard nav + voice over. 

It's a QA approveeeed ✅ 

Thanks for your contribution.

PR #963, opened 2026-03-23
Dynamically generate checkout navigation steps

### LINKED ISSUE #719: hookActionCheckoutRender cannot be used anymore to manage (add / remove) checkout steps
The hook actionCheckoutRender cannot be used to add a new checkout step. This is a regression from the classic theme because this following code is not enough anymore to add a step :
```php
    public function hookActionCheckoutRender($hookParams)
    {
        $checkoutTestStep = new CheckoutTestStep();
        $checkoutProcess = $hookParams['checkoutProcess'];

        $steps = $checkoutProcess->getSteps();

        $checkoutTestStep->setCheckoutProcess($checkoutProcess);

        $newSteps = [
            $steps[0],
            $checkoutTestStep,
            $steps[1],
            $steps[2],
            $steps[3],
        ];

        $checkoutProcess->setSteps($newSteps);
    }
```

Regarding the hummingbird theme, when we changed the checkout steps, the navigation menu was not updated, meaning customers could not access the additional steps.

Image

The problem is that the default steps are hardcoded in [checkout-navigation.tpl ](https://github.com/PrestaShop/hummingbird/blob/develop/templates/checkout/checkout-navigation.tpl), rather than being based on the steps in `CheckoutProcess` class.

ga-devfront:
| Questions         | Answers
| Description?      | Dynamically generate checkout navigation steps from the `checkout_process` object instead of hardcoding them (hooked steps would be now added properly)
| Type?             | bug fix / improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #719
| How to test?      | All checkout process work fine and we can now add a step with hook.

### DIFF
--- a/templates/checkout/checkout-navigation.tpl
+++ b/templates/checkout/checkout-navigation.tpl
@@ -12,21 +12,14 @@
   <div class="{$componentName} {if isset($notifications) && isset($hasNotifications) && $hasNotifications} {$componentName}--has-notifications{/if}">
     <div class="{$componentName}__desktop">
       <ul class="{$componentName}__list" role="tablist">
-        {* Personal Information *}
-        {include file='checkout/_partials/checkout-navigation-step.tpl' number="{l s='1' d='Shop.Theme.Checkout'}"
-        step="checkout-personal-information-step" title="{l s='Personal Information' d='Shop.Theme.Checkout'}"}
-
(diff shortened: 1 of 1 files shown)


Hlavtox (approved):
🚀

PR #977, opened 2026-04-01
fix ps_customersignin.tpl Sign in url

### LINKED ISSUE #973: Infinite redirect loop via `back` parameter in ps_customersignin
### Prerequisites

**Description:**
There is a issue in the `ps_customersignin` module where the `back` parameter is recursively appended, leading to infinite redirect loops and exponentially growing URLs.

This results in extremely long query strings and can cause severe server load when crawlers (e.g. Facebook, other bots) follow these links.

---

**Steps to reproduce:**

1. Visit the login page with a `back` parameter that already contains another `back` parameter.
2. The module appends the current URL again as `back`.
3. Repeat redirects cause nested `back` parameters to accumulate.

Example:

```
/login?back=/login?back=/login?back=...
```

---

**Expected behavior:**

* The `back` parameter should be added only once.
* Existing `back` parameters should be sanitized or replaced.
* Nested or recursive `back` parameters should be prevented.

---

**Actual behavior:**

* The `back` parameter is appended repeatedly without validation.
* URLs grow exponentially with each redirect.
* Bots can trigger massive request loops, leading to high server load and outage.

---

**Impact:**

* Potential denial-of-service–like behavior due to crawler loops
* Excessive CPU and memory usage
* Extremely long URLs (several KB)
* SEO risks (duplicate / malformed URLs)

---

**Suggested fix:**

* Remove back parameter in the .tpl file.
OR
* Before appending `back`, check if it already exists.
* Strip existing `back` parameters from the current URL.
* Validate and limit the length of the `back` parameter.
* Ensure the `back` URL does not point to another login page.

Example (pseudo):

```
if (current_url contains "back=") {
    remove existing back parameter
}
```

---

**Affected file:**

* `modules/ps_customersignin/views/templates/hook/ps_customersignin.tpl`

---

**Additional notes:**
This issue can be easily exploited by bots and should be considered a high-priority fix.

---

**Workaround:**
We mitigated this server-side by blocking requests with multiple `back` parameters via `.htaccess`, but the root cause remains in the module.

---

### Steps to reproduce

See above

### Expected behavior

_No response_

### Actual Result

_No response_

8-9

1-2

7-8

_No response_

mcaldex:
| Questions         | Answers
| Description?      | While logged out, if you click on the Sign in button the first time, you'll be redirected to the authentication page. You'll see that the url is correct, but if you click on that button again while staying in the authentication page, you will see that the current_url ( /login?back=http... ) will be appended to the "back" query parameter, so it will grow everytime you click on the Sign in button while staying in the authentication page.
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | Fixes [973](https://github.com/PrestaShop/hummingbird/issues/973)
| Sponsor company   | www.hostinato.it
| How to test?      | check if the href in the Sign in button stays the same if you're currently in the authentication page

### DIFF
--- a/modules/ps_customersignin/ps_customersignin.tpl
+++ b/modules/ps_customersignin/ps_customersignin.tpl
@@ -119,7 +119,7 @@
 
           <div class="dropdown-divider"></div>
 
-          <a 
+          <a
             href="{$logout_url}"
             class="dropdown-item"
             rel="nofollow"
@@ -132,7 +132,11 @@
     {else}
       <div class="header-block">
         <a
-          href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"
+          {if $page.page_name == 'authentication'}
+            href="{$urls.pages.authentication}?back={$urls.pages.my_account|urlencode}"
+          {else}
+            href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"
+          {/if}
           class="header-block__action-btn"
           rel="nofollow"
           aria-label="{l s='Sign in' d='Shop.Theme.Actions'}"


ps-jarvis:
Hello @mcaldex!

modules/ps_customersignin/ps_customersignin.tpl
```diff
@@ -132,7 +132,11 @@
     {else}
       <div class="header-block">
         <a
-          href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"
+          {if $page.page_name == 'authentication'}
+            href="{$urls.current_url}"
```
tblivet:
```suggestion
            href="{$urls.pages.authentication}"
```

tblivet (changes requested):
Hi, thank you for the fix 👍

I just left a comment. I believe it’s cleaner.

modules/ps_customersignin/ps_customersignin.tpl
mcaldex: Hi,
by applying your suggestion the user won't be redirected to their intended destination if they click this anchor again while being in the authetication page already.
I'm not entirely sure this would be correct, what do you think?

MerseyRay: > Hi, by applying your suggestion the user won't be redirected to their intended destination if they click this anchor again while being in the authetication page already. I'm not entirely sure this would be correct, what do you think?

That will be correct then.
If they come from a product page and sign in, they will go back to the product page.
But if they come from a product page, then click the login again, they essentially visited the login page from the login page, so will be directed back there.

The only other way would  be to disable the link on the login button when the user is already on the login page.

tblivet: Thanks @mcaldex, @MerseyRay for the feedback 🙂 after further thought, I see 2 approaches:                                                                     
                                                                                                                              
**Option 1:** Preserve shopping flow:
```
{if $page.page_name == 'authentication'}                                                                                    
  href="{$urls.pages.authentication}?back={$urls.pages.my_account|urlencode}"
{else}    
  href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"                                       
{/if}
```                                                                   
From any page, the user is redirected back to their original context after login. When already on the auth page, they land on their account page instead. This preserves the shopping flow (e.g. product page → login → back to product page).
        
**Option 2:** Always redirect to account:
```
href="{$urls.pages.authentication}?back={$urls.pages.my_account|urlencode}"
```
Simpler, no conditional. After login, the user always lands on their account page regardless of where they came from. A user coming from a product page would need to navigate back manually after login.

WDYT about this ?

MerseyRay: Or how about
```
{if $page.page_name == 'authentication'}
  href="#"
{else}
  href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"
{/if}
```

Then the user cant click the sign in button again, and they will be redirected back to the page they came from.

Ive just implemented that on my site and it seems to work fine.

mcaldex: I think it would be wise to always have a link to the login button, in case the user wants to "reset" the situation while being in the authentication, in any case.
I'd go for option 1 proposed by @tblivet :)

MerseyRay: > I think it would be wise to always have a link to the login button, in case the user wants to "reset" the situation while being in the authentication, in any case. I'd go for option 1 proposed by @tblivet :)

When im using any sites, if something doesnt work I would automatically refresh the page. I would never click the same button again.

MerseyRay: > I think it would be wise to always have a link to the login button, in case the user wants to "reset" the situation while being in the authentication, in any case. I'd go for option 1 proposed by @tblivet :)

if you want to reload the page, instead of using the reload/refresh button then why not this

```
{if $page.page_name == 'authentication'}
  href="#" onclick="location.reload(); return false;"
{else}
  href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"
{/if}
```

that works like we need, and the user can click the sign in button to their hearts content and it will return to the originating page :)

tblivet: @mcaldex @MerseyRay Thanks for your feedback 🙂                                                                                             
I agree that we should go with **Option 1**. If the redirect behavior needs deeper changes, it's the module's responsibility to implement them.
   
I'm also not in favor of using JavaScript with `onclick` in plain HTML, and I believe `href` should always point to a real URL rather than `#`, except in specific cases.

MerseyRay:
Ahh, if we use an empty href, then it links to the current page, so this is ideal

```
{if $page.page_name == 'authentication'}                                                                                    
  href=""
{else}    
  href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"                                       
{/if}
```

This doesnt use a #, or javascript. It returns the real current url

tblivet:
@MerseyRay, thanks for taking the time to look for a solution. I think this could be a good middle ground. Using an empty `href=""` seems to be well [documented here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/href), and it would fix the initial issue. If we want to improve it further, that should probably be handled directly in the module. Are you okay with this, @mcaldex?

mcaldex:
While it's true that `href=""` is documented there and it's a valid option, making `href` to always point to a real URL is much more bulletproof (clicking on it will always reset the state of the login controller no matter what), and also a certainity in terms of accessibility.

Imagine we have a module that customizes the login page a bit, and it adds additional query params to the URL to add some custom logic while the tries to login.
Leaving `href=""` would mean that the page will be refreshed while maintaining all of those query params.
On the other hand, if we explicitly redirect the user to a freshly generated login URL, we make sure that he will always start from scratch, potentially clearing up the old state given by those query params.

In my opinion if we always set an URL we only have pros and no cons, no?

tblivet:
Hi @mcaldex, your feedback makes sense. To avoid blocking this PR, I suggest we go with this option. It fixes the bug and also improves the UX by redirecting users to the account page.
```
{if $page.page_name == 'authentication'}
  href="{$urls.pages.authentication}?back={$urls.pages.my_account|urlencode}"
{else}
  href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"
{/if}
```

MerseyRay:
> Hi @mcaldex, your feedback makes sense. To avoid blocking this PR, I suggest we go with this option. It fixes the bug and also improves the UX by redirecting users to the account page.
> 
> ```
> {if $page.page_name == 'authentication'}
>   href="{$urls.pages.authentication}?back={$urls.pages.my_account|urlencode}"
> {else}
>   href="{$urls.pages.authentication}?back={$urls.current_url|urlencode}"
> {/if}
> ```

I don't think any users want redirecting the account page after logging in.  Maybe after creating an account that may be ideal.
Imagine your browsing a store and after some time find something you want to order. Then you log in, and it sends you to your account page so you have to search for that item again. That annoys me on stores when that happens.

Ill be implementing the href="" on my stores as that seems the most logical.  They are always redirected to their originating page.

tblivet:
> I don't think any users want redirecting the account page after logging in. Maybe after creating an account that may be ideal. Imagine your browsing a store and after some time find something you want to order. Then you log in, and it sends you to your account page so you have to search for that item again. That annoys me on stores when that happens.
> 
> Ill be implementing the href="" on my stores as that seems the most logical. They are always redirected to their originating page.

Hi @mcaldex , I see your point, but that situation would only happen if the user clicked the links twice, so I do not think it is a real issue. In a normal flow, if a user clicks the login link from a product page and then logs in, they will be redirected back to that same product page.

MerseyRay:
> > I don't think any users want redirecting the account page after logging in. Maybe after creating an account that may be ideal. Imagine your browsing a store and after some time find something you want to order. Then you log in, and it sends you to your account page so you have to search for that item again. That annoys me on stores when that happens.
> > Ill be implementing the href="" on my stores as that seems the most logical. They are always redirected to their originating page.
> 
> Hi @mcaldex , I see your point, but that situation would only happen if the user clicked the links twice, so I do not think it is a real issue. In a normal flow, if a user clicks the login link from a product page and then logs in, they will be redirected back to that same product page.

I was taking into account what marsaldev said earlier...

"mcaldex : I think it would be wise to always have a link to the login button, in case the user wants to "reset" the situation while being in the authentication, in any case."

ie, if they loaded the sign in page and it didnt look right for some reason. they either refresh, or click login again to reload the page.  Also, the less techy people still click things twice on web pages because they think its like a desktop.

Either fix is good as it eliminates the bug which is the main thing.

tblivet (approved):
Thanks @mcaldex 🙏

tblivet (changes requested):
Sorry, I’m blocking to update the branch destination to 2.x

tblivet (changes requested):
@mcaldex, could you please rebase on `2.x` so the branch contains only your commit?

tblivet (approved):
Thanks for your patience 🙏

ingridusta (approved):
Hi @mcaldex, 

Thanks for your contribution. 

I reproduced the bug : 

https://github.com/user-attachments/assets/e324d524-252f-468c-bbec-eaae91bb24f5

And with your fix : 

https://github.com/user-attachments/assets/4cf1b6db-ed3c-4de0-a06c-340893ae4a04

Also the user is redirected to the page he was before signing in.

It's QA approved ✅

ps-jarvis:
PR merged, well done!

Issue #731, opened 2025-06-27
Module blockwishlist errors


adorade:
The Hummingbird theme utilizes BS 5, whereas the module relies on BS 4, which I believe is the source of these errors:
- products-sort-order (dropdown not working)
- wishlist product pagination

PrestaShop v8.2.1, Hummingbird v1.0.1

![Image](https://github.com/user-attachments/assets/2d479158-4b0e-4ec5-9c3d-5c8d5524b350)
![Image](https://github.com/user-attachments/assets/61c27104-694f-4f66-81a9-c5f343365501)

PR #996, opened 2026-04-28
fix : make wishlist-footer-links clickable

kookline:
| Questions         | Answers
| Description?      |  The links in the footer of the product list page of the blockwishlist module are not clickable, they are under the ul.wishlist-products-list . I did the same thing as what is done in the classic theme, set the pagination's display to inline-block. This is volontarily a conservative change, applied only to .wishlist-pagination, because we don't want to create side effects. It may be better to apply it to the .pagination class globally if other modules encounter the same issue. 
| Type?             | bug fix 
| BC breaks?        | no
| Deprecations?     |no
| Fixed ticket?     | n/a
| Sponsor company   | Kookline
| How to test?      |  On a new prestashop 9.1 install, with Hummingbird 2.0.0, enable blockwishlist module, create a wishlist in the customer account, add a product to it and navigate to the list. Try to click on either link under the product, home / back to wishlist

### DIFF
--- a/src/scss/prestashop/modules/_blockwishlist.scss
+++ b/src/scss/prestashop/modules/_blockwishlist.scss
@@ -40,4 +40,8 @@ $component-name: wishlist;
       color: inherit !important;
     }
   }
+
+  &-pagination {
+    display: inline-block;
+  }
 }


ps-jarvis:
Hello @kookline!

tblivet (changes requested):
Hi @kookline, thank you for the PR.

I am not in favor of adding code to support the **blockwishlist** module, as it has been disabled in the Hummingbird configuration. The module is buggy and does not meet the quality bar we set for Hummingbird, so I believe this fix either should not exist in the theme or should live inside the module itself. The module needs a proper refactor before it can be considered usable with Hummingbird.

Progi1984:
Thanks for the fix, @kookline. It works on my shops.

tblivet:
Hi @kookline, thanks for taking the time to fix it!

The blockwishlist module is currently disabled by default in Hummingbird v2 due to multiple compatibility issues (see [#731](https://github.com/PrestaShop/hummingbird/issues/731) and [blockwishlist#395](https://github.com/PrestaShop/blockwishlist/issues/395)). A global contribution on the module would be needed to properly address these.

I'm closing this PR for now, the issues you found are real and your analysis will be useful when someone takes on a full rework of the module.

Thanks again for the contribution 🙏

PR #999, opened 2026-05-05
Feat: add support for one-page checkout

ThbPS:
| Questions         | Answers |
| ------------------| ------------------------------------------------------- |
| Description?      | Adds support for switching between the classic 4-step checkout and one-page checkout. The checkout navigation is hidden when OPC is enabled, and a hidden toast hook is added for the account-created message after registration. |
| Type?             | new feature |
| BC breaks?        | no |
| Deprecations?     | no |
| Fixed ticket?     | N/A |
| Sponsor company   | PrestaShop |
| How to test?      | Enable one-page checkout, open the checkout page, and verify that the classic step navigation is not displayed. Create an account from checkout and verify that the account-created toast can be displayed after redirect. Then disable one-page checkout and verify that the classic 4-step checkout navigation still appears. |

### DIFF
--- a/templates/checkout/checkout.tpl
+++ b/templates/checkout/checkout.tpl
@@ -11,16 +11,24 @@
 {/block}
 
 {block name='content_columns'}
-  {include file='checkout/checkout-navigation.tpl'}
+  {if isset($is_one_page_checkout_enabled) && !$is_one_page_checkout_enabled}
+    {include file='checkout/checkout-navigation.tpl'}
+  {/if}
 
   {block name='checkout_notifications'}
     {include file='_partials/notifications.tpl'}
   {/block}
+  <div
+    id="js-account-created-toast"
+    class="d-none"
+    data-show="{if !empty($show_account_created_toast)}1{else}0{/if}"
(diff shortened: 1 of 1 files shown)


ps-jarvis:
Hello @ThbPS!

Hlavtox (changes requested):
I think that there should be a separate template for OPC, this is a recipe for hell.

ThbPS:
[question](https://pullpo.io/cc?l=question) 

@Hlavtox

Since a module cannot cleanly override the theme’s checkout page template, achieving zero OPC dependency in `checkout.tpl` requires a Core change so that `OrderController` selects the page template directly.
Or are you suggesting a different approach?

kpodemski:
@ThbPS 

In my opinion, the OPC controller can incorporate its own `templates/checkout/checkout.tpl` and all the needed templates.

This means `checkout.tpl` would live directly in the OPC module, and you can have all control over the front-office part :)

Alternative option: I'm pretty sure you can extend this template from the module, and correctly use `append` or maybe even override certain blocks of this page, but, the cleanest way, as mentioned by @Hlavtox is to have a dedicated template, and, to follow my suggestion: directly inside the OPC module.

PR #1015, opened 2026-05-20
docs: add PRODUCT.md with product role, users, flows, and scope bound…

hwaseonchoi:
…aries

| Questions         | Answers
| Description?      | Add `PRODUCT.md` as part of PS starter pack 
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | N/A

### DIFF
--- a/PRODUCT.md
+++ b/PRODUCT.md
@@ -1,0 +1,85 @@
+# Hummingbird — Product Context
+
+## Product Role
+
+Hummingbird is the official default theme for PrestaShop 9.1+. Its primary purpose
+is to serve as the production-ready reference implementation that ships with the
+platform — not a blank starter, not a showcase, but the theme a merchant gets
+out of the box and a developer forks as their baseline.
+
+It defines what "correct" looks like for PrestaShop frontend development:
+accessibility, performance, separation of concerns, and compatibility with the
+Core's Smarty variable contracts. All other themes, modules, and integrations
+are implicitly measured against it.
+
+## Users
+
+**Theme developers (primary)**
+Agencies, freelancers, and in-house teams who fork Hummingbird to build custom
+themes for merchants. They need the codebase to be predictable, well-structured,
+and safe to extend without fighting the architecture.
+
+**PrestaShop Core contributors**
+Members of the PS team and open-source contributors who evolve the theme
+alongside the Core. They need clear boundaries between what belongs in the theme
+versus the Core, and reliable conventions for adding features without regressions.
+
+**Module developers**
+Developers building PrestaShop modules who need their front-office templates to
+integrate cleanly with the default theme. They rely on Hummingbird's HTML
+structure, CSS layers, and JS architecture to be stable and documented.
+
+**Merchants (indirect)**
+End users of PrestaShop stores running Hummingbird. They are not direct consumers
+of this repo but are the ultimate beneficiaries of its quality. Performance,
+accessibility, and browser compatibility decisions are made with them in mind.
+
+## Key Flows
+
+These are the storefront flows Hummingbird owns end-to-end:
+
+- **Product catalog** — category listing, search results, filters
+- **Product page** — detail, gallery, variants, add-to-cart
+- **Cart** — line items, quantity update, coupon, summary
+- **Checkout** — address, shipping, payment selection, order confirmation
+- **Customer account** — login, registration, order history, returns
+- **CMS pages** — static content, contact form
+- **Error pages** — 404, maintenance
+
+## Business Rules
+
+- **Strict presentation boundary.** Hummingbird is a rendering layer only. It
+  consumes Smarty variables injected by the Core and renders HTML. It never
+  queries the database, overrides controllers, or duplicates business logic.
+
+- **Core contract compliance.** The theme must remain compatible with every
+  Smarty variable and hook provided by the supported PrestaShop version (9.1+).
+  If a variable or hook is missing, the fix belongs in the Core — not in a
+  Smarty plugin workaround inside the theme.
+
+- **Module template isolation.** Modules may override theme templates via the
+  standard PrestaShop override mechanism. The theme must not break this contract
+  by hardcoding assumptions about module output.
+
+- **Accessibility is non-negotiable.** All rendered markup must meet W3C
+  WAI-ARIA standards. Accessibility regressions are treated as bugs, not
+  cosmetic issues.
+
+- **No jQuery.** The theme targets modern browsers and must not introduce jQuery
+  as a dependency, even transitively through third-party scripts.
+
+## Out of Scope
+
+The following are explicitly not the theme's responsibility:
+
+- **Business logic** — pricing rules, promotions, tax calculations, inventory.
+  These belong in the PrestaShop Core or dedicated modules.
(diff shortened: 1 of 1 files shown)


ps-jarvis:
Hello @hwaseonchoi!

kpodemski (dismissed):

tblivet:
Hi @hwaseonchoi, this file doesn't seem to follow a naming convention used by AI tools, so I think it should be referenced from another file like CONTEXT.md to actually be picked up.

Moreover, while the content itself looks fine, I feel it adds some redundancy with what's already in CONTEXT.md. It would be worth checking what's truly new here versus what's already covered there.

On top of that, I think the file could be more product-oriented. Sections like "Business Rules" (no jQuery, core contract compliance, module template isolation) are technical and architectural constraints that belong in CONTEXT.md. A PRODUCT.md should focus purely on who the users are, what flows they own, and what's in or out of scope from a product perspective and not implementation rules, what do you think ?

hwaseonchoi:
@tblivet Thanks for the review ! 

> Hi @hwaseonchoi, this file doesn't seem to follow a naming convention used by AI tools, so I think it should be referenced from another file like CONTEXT.md to actually be picked up.

Can you give us an example please ? 

> 
> Moreover, while the content itself looks fine, I feel it adds some redundancy with what's already in CONTEXT.md. It would be worth checking what's truly new here versus what's already covered there.

Indeed, we're handling some of the same themes because many of HB2 features are also a bit "technical". I wish that CONTEXT.md is served as implementation guide based on product guidelines from PRODUCT.md. That's the main goal beyond the repetition of same theme titles.

> 
> On top of that, I think the file could be more product-oriented. Sections like "Business Rules" (no jQuery, core contract compliance, module template isolation) are technical and architectural constraints that belong in CONTEXT.md. A PRODUCT.md should focus purely on who the users are, what flows they own, and what's in or out of scope from a product perspective and not implementation rules, what do you think ?

I updated a little bit by removing "business rules" section which have the most redundant info.

Overall, I'm open to discuss about the value/direction of this new product. 🙂

tblivet:
Hi @hwaseonchoi!

I was thinking of something like this, right after the title of the CONTEXT.md file.

```
Before proceeding, also read `PRODUCT.md` for product context, user personas, storefront flows, and scope boundaries.
```

To make sure AI doesn’t just ignore the PRODUCT.md file.

PRODUCT.md
```diff
@@ -0,0 +1,68 @@
...
+Developers building PrestaShop modules who need their front-office templates to
+integrate cleanly with the default theme. They rely on Hummingbird's HTML
+structure, CSS layers, and JS architecture to be stable and documented.
+
+**Merchants (indirect)**
+End users of PrestaShop stores running Hummingbird. They are not direct consumers
+of this repo but are the ultimate beneficiaries of its quality. Performance,
+accessibility, and browser compatibility decisions are made with them in mind.
+
+## Key Flows
```
tblivet, line 44: I don’t really see the value of this “Key flows” section as it is. What do you think about making it more product-oriented?

```
**Product page**: highest user and developer impact.
The most visited page and the most extended by developers. Any regression here affects both
merchant conversion and the developer experience of forking the theme. Accessibility is
non-negotiable: variant changes, availability updates, and gallery interactions must work
without a mouse.

**Category / listing**: performance-sensitive, display-heavy.
Critical for large catalogs. The theme owns product grid rendering, sorting, pagination, and
subcategory display.

**Cart**: conversion-sensitive, feedback-critical.
Every interaction (quantity change, item removal, voucher) should give immediate, accessible
feedback to keep the experience smooth and trustworthy.

**Checkout**: most conversion-critical flow; zero regression tolerance.
A broken or degraded checkout directly loses orders. This flow must be treated with the
highest stability priority.

**Customer account**: retention-focused, low complexity.
Covers login, registration, order history, and returns. Lower risk than transactional flows
but directly affects customer trust. Form interactions (password policy, validation) must
be clear and accessible.

**CMS and error pages**: low priority, lightweight by design.
Static content and error states. Should remain minimal and simple.
```

PRODUCT.md
```diff
@@ -0,0 +1,68 @@
...
+  These belong in the PrestaShop Core or dedicated modules.
+- **Module-owned features** — payment, B2B workflows, loyalty, marketplace, and
+  any other feature delivered by a module. Modules bring their own templates;
+  the theme provides no dedicated UI for them.
+- **Multi-shop configuration** — shop group logic, domain routing.
+- **Back-office UI** — admin panels, product editing, order management.
+- **Session state and data persistence** — the theme has no ownership over user
+  state or stored data.
+- **SEO meta strategy** — meta content and structured data are the Core's
+  responsibility; the theme renders what it receives.
```
M0rgan01, line 84: Shouldn't this kind of context file be in an AI folder? If we all start adding files, the root directory will quickly become full of files.

I also don't really understand the purpose of this file. How will it help the AI?

kpodemski: I'm assuming it's more about discussing product's future with AI :) I agree it would be better to have it somewhere in the directory. Maybe follow the same pattern as core, and have it under .ai/

M0rgan01: The context available to an AI can quickly become overloaded. Adding generic product descriptions like this may:

- unnecessarily increase token usage
- increase processing time for each interaction
- dilute project-specific information that actually improves output quality

Most of this content describes broad architectural concepts (theme role, users, scope boundaries) rather than implementation details or repository-specific constraints. For an AI assistant, high-signal project conventions, coding standards, architecture decisions, or custom rules usually provide more value than generic product descriptions.

hwaseonchoi: This is all new to everyone, so I understand we question ourselves the value of this initiative. 

We try to put a global, agreed & efficient operating system in all squads with the use of IA, so for this moment it's only a foundation but tomorrow we may create agents which require only product part without charging all architectural & implementation rules (which can costs tokens), we may precise CONTEXT.md file that this is a pointer not the mandatory reading to do for IA.

I suggest to start with, and can improve the way it's charged / used and even re-question the value of the separation of different AI document.

hwaseonchoi:
@tblivet Hi 👋🏻 are you waiting for @ga-devfront review or could we move forward without him ?

tblivet:
Hi @hwaseonchoi,

Based on the last discussion with @M0rgan01 and your answer, I believe we should finally remove the `CONTEXT.md` change and leave the new `PRODUCT.md` file "alone".

As it stands, pointing to it from `CONTEXT.md` means the AI loads it on every interaction, which adds token cost and dilutes the implementation rules, without bringing real value for coding at this moment.

As you mentioned, `PRODUCT.md` is more of a foundation that will be useful later for dedicated skills/agents, so it makes sense to keep it standalone for now and decide how it gets consumed (and where it lives) in the future.

So, let's drop the `CONTEXT.md` edit, keep `PRODUCT.md`, and I'll approve it ! No need for @ga-devfront to review. Does that work for you? 🙂

PRODUCT.md
hwaseonchoi: Great idea, better detailed and product oriented about each part. ✅

tblivet (approved):

Issue #1025, opened 2026-05-29
Mobile offcanvas menu remains open when resizing to desktop layout


Codencode:
### Prerequisites

- [x] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md).
- [x] I have already [searched in existing issues](https://github.com/PrestaShop/hummingbird/issues?q=is%3Aissue+label%3ABug) and found no previous report of this bug.

### Describe the bug and add attachments

When using the Hummingbird theme on mobile, the main menu opens as an offcanvas menu.

I noticed that if the mobile menu is open and the viewport is then resized to the desktop layout, the offcanvas menu remains open instead of being closed automatically.

https://github.com/user-attachments/assets/f1d4d549-b358-4916-bf6d-d0dd2a4fc66c

### Steps to reproduce

1. Open the Hummingbird theme on a mobile viewport.
2. Open the mobile menu.
3. Resize the viewport to desktop width.
4. Observe that the offcanvas mobile menu remains open.

### PrestaShop version where the bug happens

all

### Theme version where the bug happens

2.0.0

### Your company or customer's name goes here (if applicable).

Codencode snc

PR #1026, opened 2026-05-29
Close mobile menu when switching to desktop layout

Codencode:
Automatically closes the mobile offcanvas menu when the viewport switches to the desktop layout, preventing the mobile menu from remaining open after resizing.

| Questions         | Answers
| Description?      | This PR closes the mobile offcanvas menu automatically when switching from the mobile layout to the desktop layout.<br><br>The visibility of the desktop menu is checked using the actual computed style of `.js-menu-desktop`, avoiding hardcoded breakpoints.<br><br>When the desktop menu becomes visible after a viewport resize, the existing Bootstrap Offcanvas instance is closed using `Offcanvas.getInstance(menuCanvas)?.hide()`.<br><br>This prevents the mobile menu from remaining open in an inconsistent state after resizing the viewport to desktop.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/hummingbird/issues/1025
| How to test?      | 1. Open the Hummingbird theme on a mobile viewport.<br>2. Open the mobile menu.<br>3. Resize the viewport to a desktop width.<br>4. Check that the mobile offcanvas menu is closed automatically when the desktop menu becomes visible.

### DIFF
--- a/src/js/mobile-menu.ts
+++ b/src/js/mobile-menu.ts
@@ -3,6 +3,7 @@
  * file that was distributed with this source code.
  */
 
+import {Offcanvas} from 'bootstrap';
 import {isHTMLElement} from '@helpers/typeguards';
 
 const initMobileMenu = () => {
@@ -14,6 +15,7 @@ const initMobileMenu = () => {
   const backButton = document.querySelector(MobileMenuMap.backButton);
   const menuCanvas = document.querySelector(MobileMenuMap.menuCanvas);
   const defaultBackTitle = backTitle?.innerHTML;
+  const desktopMenu = document.querySelector('.js-menu-desktop');
 
   const backToParent = () => {
     if (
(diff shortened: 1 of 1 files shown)


Codencode:
Hello @tblivet @ga-devfront 
after opening this PR I noticed that some other websites/components with a similar mobile offcanvas menu behavior also keep the menu open when resizing from mobile to desktop.<br><br>Even if this PR fixes the behavior described in #1025, I’m not sure whether this should actually be considered a bug or an acceptable behavior.<br><br>Could you please confirm if we want to automatically close the mobile menu when the desktop menu becomes visible, or if we should keep the current behavior?<br><br>Thank you

ga-devfront:
> Hello @tblivet @ga-devfront after opening this PR I noticed that some other websites/components with a similar mobile offcanvas menu behavior also keep the menu open when resizing from mobile to desktop.Even if this PR fixes the behavior described in #1025, I’m not sure whether this should actually be considered a bug or an acceptable behavior.Could you please confirm if we want to automatically close the mobile menu when the desktop menu becomes visible, or if we should keep the current behavior?Thank you

Hello @Codencode !
This is just my opinion, but I think it's a "fix" that isn't really a fix. This behavior is specific to developers; average users use their mobile devices 80% of the time, and when they're on desktop, they almost never resize to mobile. This adds code and therefore complexity for very little gain. What do you think?

Codencode:
Hello @ga-devfront, yes, as mentioned, after creating the PR I checked a few other websites from major brands and in this situation they also do not close the mobile menu, so we can leave it as it is.

I just wanted to get feedback from someone who worked on the Hummingbird theme :)

We are currently running an accessibility audit and I will open other PRs about it.

Thank you

tblivet:
Hi @Codencode, I share the same opinion as @ga-devfront.
I believe this fix only adds unnecessary code. In practice, when users navigate on mobile, they usually do so in portrait mode. Also, in most cases, switching a mobile device from portrait to landscape will still not provide enough width to display the desktop menu mode.
So I also believe this fix is not needed, and I would suggest closing the PR.

Codencode:
@ga-devfront @tblivet 
Okay, that's fine.
Thank you for your considerations!

PR #1028, opened 2026-05-29
Improve rich text link accessibility in Hummingbird theme

Codencode:
| Questions         | Answers
| Description?      | Improves accessibility of links inside rich text content by restoring a persistent underline and making hover/focus states perceivable beyond color alone.<br/>This helps users distinguish links more reliably in CMS pages, product descriptions, and category descriptions.<br/>It aligns the theme with better accessibility practices by not relying on color as the only visual cue for interactive text.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | @Codencode 
| How to test?      | 1. Open a CMS page, a product page with description content, and a category page with description content, then verify that links inside rich text are underlined by default.<br/>2. Hover those links and verify that the hover state is perceivable not only through color change, but also through the thicker and more offset underline.<br/>3. Navigate with keyboard using Tab and verify that focused links display a visible focus outline.<br/>4. Confirm that links outside rich text areas keep their existing styling unchanged.

### DIFF
--- a/src/scss/prestashop/components/_rich-text.scss
+++ b/src/scss/prestashop/components/_rich-text.scss
@@ -5,6 +5,17 @@
 .rich-text {
   line-height: 1.5;
 
+  a {
+    text-decoration: underline;
+    text-decoration-thickness: 0.1em;
+    text-underline-offset: 0.15em;
+
+    &:hover {
+      color: var(--bs-link-hover-color);
+      text-decoration-thickness: 0.2em;
+    }
+  }
+
   table {
     @include table;
     @include table-bordered;
--- a/src/scss/prestashop/components/_rich-text.scss
+++ b/src/scss/prestashop/components/_rich-text.scss
@@ -7,7 +7,6 @@
 
   a {
     text-decoration: underline;
-    text-decoration-thickness: 0.1em;
     text-underline-offset: 0.15em;
 
     &:hover {
--- a/src/scss/prestashop/components/_rich-text.scss
+++ b/src/scss/prestashop/components/_rich-text.scss
@@ -7,11 +7,10 @@
 
   a {
     text-decoration: underline;
-    text-underline-offset: 0.15em;
+    text-underline-offset: 0.125em;
 
     &:hover {
-      color: var(--bs-link-hover-color);
-      text-decoration-thickness: 0.2em;
+      text-underline-offset: 0.25em;
     }
   }
 


src/scss/prestashop/components/_rich-text.scss
```diff
@@ -1,13 +1,20 @@
...
+  a {
+    text-decoration: underline;
+    text-decoration-thickness: 0.1em;
+    text-underline-offset: 0.15em;
+
+    &:hover {
+      color: var(--bs-link-hover-color);
+      text-decoration-thickness: 0.2em;
+    }
+  }
```
tblivet, line 15: Hi @Codencode, thank you for this PR 🙏 but I believe we should stick to this minimal code, which is cleaner in terms of rendering.

Could you please rebase your PR and change the target branch to `2.x`?

```suggestion
  a {
    text-decoration: underline;
    text-underline-offset: 0.125em;
  }
```

tblivet (changes requested):

src/scss/prestashop/components/_rich-text.scss
Codencode: Hi @tblivet, thanks for the feedback.

I agree with simplifying the default underline style, but I would prefer to keep a visible hover state for links inside rich text.

The goal of this change is related to WCAG 1.4.1 “Use of Color”, which requires that color is not used as the only visual means of distinguishing a visual element or indicating an action.

Reference:
https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html

WCAG technique G183 also provides guidance about adding visual cues for links, including `:hover` and `:focus` states, so that users can more easily identify links and confirm the interactive element when using a mouse or keyboard.

Reference:
https://www.w3.org/WAI/WCAG22/Techniques/general/G183

For this reason, I would suggest keeping a visible hover state while simplifying the default underline style:

```scss
a {
  text-decoration: underline;
  text-underline-offset: 0.15em;

  &:hover {
    color: var(--bs-link-hover-color);
    text-decoration-thickness: 0.2em;
  }
}
```

The default underline ensures that links are not identified by color alone, while the hover state provides an additional visible interaction feedback for pointer users.

What do you think?

I have also rebased the PR on the `2.x` branch.

tblivet: Hi @Codencode, thank you for the clarification. Could you also change the target branch to `2.x`?

I would go for something like this, what do you think?
Bootstrap already provides a subtle color change on hover, so we don’t need to define a new one.

```css
a {
  text-decoration: underline;
  text-underline-offset: 0.125em;

  &:hover {
    text-decoration: none;
    /* OR */
    text-underline-offset: 0.25em;
  }
}
```

Codencode: Hi @tblivet, yes, I had forgotten to change the target branch and I’ve done it now.
I went with this solution — we would prefer to have an underlined effect on hover as well.
What do you think?

```css
  a {
    text-decoration: underline;
    text-underline-offset: 0.125em;

    &:hover {
      text-underline-offset: 0.25em;
    }
  }
```

tblivet (approved):
@Codencode Perfect, thank you 🙏

ingridusta (approved):
Hi @Codencode, 

Before your improvement : 

https://github.com/user-attachments/assets/70e6a0fd-e3ce-452c-91d8-18e8c6484983

<br>With your improvement : 

https://github.com/user-attachments/assets/f778bfe1-7b29-472e-95fb-a7cc0b730bd1

<br>Links in category description & product description are more visible ✅ 

It's QA approved ✅, thanks for your contribution to Hummingbird project !

PR #1029, opened 2026-05-29
add more displayProductPriceBlock hooks in product templates

Markus-Gurkcity:
| Questions         | Answers
| Description?      | we need more displayProductPriceBlock hooks in the product templates to display legal informations
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      |

### DIFF
--- a/templates/catalog/_partials/miniatures/product.tpl
+++ b/templates/catalog/_partials/miniatures/product.tpl
@@ -50,6 +50,9 @@
                     {/if}
                   </div>
 
+                  {hook h='displayProductPriceBlock' product=$product type="price"}
+                  {hook h='displayProductPriceBlock' product=$product type="after_price"}
+
                   {hook h='displayProductPriceBlock' product=$product type='unit_price'}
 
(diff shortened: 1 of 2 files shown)


ps-jarvis:
Hello @Markus-Gurkcity!

templates/catalog/_partials/miniatures/product.tpl
```diff
@@ -50,6 +50,9 @@
                     {/if}
                   </div>
 
+                  {hook h='displayProductPriceBlock' product=$product type="price"}
```
Hlavtox, line 53: What sense does it make to put two exact hooks here?

Markus-Gurkcity: The hooks differ from the parameter 'type'.

Hlavtox: Still the same question, why you cannot use the ones there, if they are on the same place?

PR #1033, opened 2026-05-29
Remove duplicate IDs from customer account module links in Hummingbird

### LINKED ISSUE #1032: `displayCustomerAccount` module links can generate duplicate HTML ids between sidebar and main account page
### Prerequisites

In the Hummingbird theme, the `displayCustomerAccount` hook is rendered in two different account contexts:

- the account sidebar
- the main "My account" page

Some modules attached to this hook render links with static HTML ids such as `emailalerts_link`, `wishlist_link`, or `psgdpr_link`.

Because the hook output is reused in both locations, the resulting page contain duplicate `id` attributes.

### Steps to reproduce

1. Perform login in the front office
2. Access the account
3. View the HTML and you will see 2 links with ID `emailalerts_link`

### Expected behavior

_No response_

### Actual Result

_No response_

9.0.0

2.0.0

_No response_

Codencode snc

Codencode:
| Questions         | Answers
| Description?      | This PR removes static HTML ids from Hummingbird module account-link templates rendered through `displayCustomerAccount`. Since the same hook output can appear in both the account sidebar and the main **My account** page, static ids may be duplicated in the DOM. Removing those ids avoids invalid duplicated markup without requiring coordinated changes across modules.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/1032
| How to test?      | 
| ~~Related PR~~ | ~~https://github.com/PrestaShop/ps_emailalerts/pull/159~~

### Notes
~~This PR is part of a two-step fix. It introduces the account rendering context in Hummingbird by passing `isMainAccount` to the `displayCustomerAccount` hook. A companion PR in ps_emailalerts consumes that parameter to generate distinct link IDs and complete the fix.~~

### DIFF
--- a/modules/blockwishlist/views/templates/hook/displayCustomerAccount.tpl
+++ b/modules/blockwishlist/views/templates/hook/displayCustomerAccount.tpl
@@ -18,7 +18,7 @@
  *}
 
 <a
-  class="account-menu__link {if $urls.current_url === $url} account-menu__link--active{/if}"
+  class="account-menu__link account-menu__link--wishlist{if $urls.current_url === $url} account-menu__link--active{/if}"
   href="{$url}"
   {if $urls.current_url === $url}aria-current="page"{/if}
 >
--- a/modules/blockwishlist/views/templates/hook/displayCustomerAccount.tpl
+++ b/modules/blockwishlist/views/templates/hook/displayCustomerAccount.tpl
@@ -19,7 +19,6 @@
 
 <a
   class="account-menu__link {if $urls.current_url === $url} account-menu__link--active{/if}"
-  id="wishlist_link"
   href="{$url}"
   {if $urls.current_url === $url}aria-current="page"{/if}
 >
--- a/modules/ps_emailalerts/views/templates/hook/my-account.tpl
+++ b/modules/ps_emailalerts/views/templates/hook/my-account.tpl
@@ -4,7 +4,7 @@
  *}
 
 <a
-  class="account-menu__link {if $urls.current_url === $link->getModuleLink(ps_emailalerts, 'account')} account-menu__link--active{/if}"
+  class="account-menu__link account-menu__link--emailalerts{if $urls.current_url === $link->getModuleLink(ps_emailalerts, 'account')} account-menu__link--active{/if}"
   href="{$link->getModuleLink(ps_emailalerts, 'account')}"
   {if $urls.current_url === $link->getModuleLink(ps_emailalerts, 'account')}aria-current="page"{/if}
 >
--- a/modules/ps_emailalerts/views/templates/hook/my-account.tpl
+++ b/modules/ps_emailalerts/views/templates/hook/my-account.tpl
@@ -5,7 +5,11 @@
 
 <a
   class="account-menu__link {if $urls.current_url === $link->getModuleLink(ps_emailalerts, 'account')} account-menu__link--active{/if}"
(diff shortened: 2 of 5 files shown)


Hlavtox:
I would just delete the id and call it a day. There can be more than 2 links - header, my account, footer, what ID will you use then? :-)

Codencode:
@Hlavtox 
Yes, actually deleting the ID would be the best choice. I didn't do it because I thought it was necessary, but it's fine with me. That way there's no need to add the parameter to the hook. Shall I proceed this way?

Codencode:
@Hlavtox
Applied your suggestion: I removed the static IDs from the Hummingbird module account links to avoid duplicate IDs without adding context-specific naming.

tblivet (commented):
Hi @Codencode, I’m in favor of removing these IDs, but this would certainly break UI tests based on what I see here: https://github.com/PrestaShop/ui-testing-library/blob/main/src/versions/develop/pages/FO/hummingbird/myAccount/index.ts

What do you think @PrestaShop/qa-automation?

Hlavtox (dismissed):

Codencode:
> Hi @Codencode, I’m in favor of removing these IDs, but this would certainly break UI tests based on what I see here: https://github.com/PrestaShop/ui-testing-library/blob/main/src/versions/develop/pages/FO/hummingbird/myAccount/index.ts
> 
> What do you think @PrestaShop/qa-automation?

@tblivet good catch!

Maybe we could solve this by moving the current ID values to the `class` attribute instead?

This would still remove the duplicated IDs from the DOM, but it would keep a stable selector that could be used by the UI tests, for example by updating selectors from `.account-menu__nav #wishlist_link`  to account-menu__nav .wishlist_link`.

What do you think?

tblivet:
@Codencode We have some elements that use specific classes in Hummingbird and are used in UI tests. They are defined as CSS modifiers, like `account-menu__link--wishlist`, so I think this is the best approach to maintain consistency.

Codencode:
> @Codencode We have some elements that use specific classes in Hummingbird and are used in UI tests. They are defined as CSS modifiers, like `account-menu__link--wishlist`, so I think this is the best approach to maintain consistency.

@tblivet These classes already exist, so the PR is fine, meaning we just need to remove the IDs, right?

tblivet:
No, we don’t have it for now. We have `account-menu__link--active`, but we don’t have `account-menu__link--wishlist`.

Codencode:
@tblivet, if I understood your suggestion correctly, I added dedicated classes to these module account links so we can keep stable selectors without relying on duplicated IDs.

tblivet (approved):
@Codencode Yes, that’s exactly what I meant 👍 We just need @PrestaShop/qa-automation to update the UI tests.

Codencode:
Perfect!
Should the tests be updated after this PR is merged, or before?

tblivet:
Good question 🤷‍♂️ I’ll try to create a PR this afternoon if you want 👍

tblivet:
Here is the associated PR for UI tests: https://github.com/PrestaShop/ui-testing-library/pull/991

ingridusta (approved):
Hi @Codencode,

IDs for modules have been deleted : 

Before : 
Image

<br>After : 
Image


Thanks for your contribution ! It's QA approved ✅

PR #1041, opened 2026-06-09
Remove product_without_taxes block as it is never displayed

### LINKED ISSUE #1031: Incorrect condition for displaying the price without tax
### Prerequisites

https://github.com/PrestaShop/hummingbird/blob/114d9eb5001963cc94715a34dad1098f2b7518f1/templates/catalog/_partials/product-prices.tpl#L86

From what I checked, the getTaxCalculationMethod method of the Product class returns only 0/1

https://github.com/PrestaShop/PrestaShop/blob/2adc1ad13d007deebb2f660571073d911075b7cc/classes/Product.php#L915

### Steps to reproduce

Toggle the price display method and check if the price without tax is displayed

### Expected behavior

The price without tax can be shown as additional information (b2b customers) when the price display method is the with tax ( == 0) but the price without tax requires formatting in core

### Actual Result

_No response_

dev

2.0.0

_No response_

_No response_

tblivet:
| Questions         | Answers
| Description?      | Remove the `product_without_taxes` block, as it is never displayed due to an incorrect condition (`$priceDisplay == 2`, while only returns `0` or `1`). I believe this feature displaying at the time both prices (with and without taxes) is not needed in the default theme. WDYT?
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/hummingbird/issues/1031
| How to test?      | --

### DIFF
--- a/src/scss/prestashop/pages/_product.scss
+++ b/src/scss/prestashop/pages/_product.scss
@@ -68,6 +68,12 @@ $component-name: product;
       color: var(--bs-tertiary-color);
     }
 
+    &__taxless-price {
+      font-size: 0.875rem;
+      line-height: 1.2;
+      color: var(--bs-tertiary-color);
+    }
+
     &__prices {
       font-size: 1.25rem;
       line-height: 1;
--- a/templates/catalog/_partials/product-prices.tpl
+++ b/templates/catalog/_partials/product-prices.tpl
@@ -81,12 +81,6 @@
             {/if}
           {/block}
         </div>
-
-        {block name='product_without_taxes'}
-          {if $priceDisplay == 2}
-            <span class="product__taxless-price">{l s='%price% tax excl.' d='Shop.Theme.Catalog' sprintf=['%price%' => $product.price_tax_exc]}</span>
-          {/if}
-        {/block}
       </div>
     {/block}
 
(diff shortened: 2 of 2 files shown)


Hlavtox:
I think this should be fixed instead, a loooooot of clients want this - to display price both with and without tax.

Codencode:
Perhaps I would leave it so that it can be linked to the new B2B feature?
Even though as of today it would never be used.

ps-jarvis:

- `Shop.Theme.Catalog`
    - [ ] `%price% tax excluded`

tblivet:
Well, thank you @Codencode & @Hlavtox for your point of view 👍 I pushed a new commit. There's still a small formatting issue that needs to be addressed on the core side for the price.

image

Codencode (approved):

ingridusta (approved):
Hi @tblivet, 

Thanks for your fix, it works ! 


- When b2b activated + tax included both prices are displayed : 

Image
The only thing is that the currency ain't displayed for the price without taxes

- <br>And when b2b activated + tax excluded only off taxes price is displayed : 

Image

It's QA approved ! ✅

PR #1049, opened 2026-06-19
Fix searchbar not searching on paste

### LINKED ISSUE #974: ps_searchbar: product search functionality isn't triggered when the user pastes in something in the input
### Prerequisites

https://github.com/user-attachments/assets/623fa5e2-56e8-4f5a-96eb-2c1db484d55a

### Steps to reproduce

1. paste in something in the searchbar (ps_searchbar module needs to be enabled)
2. the results won't get updated in any case: if you had results visible from a search you did initiate before the paste in then the results won't update, if the input was clear before the paste in then no result will show up in any case

### Expected behavior

_No response_

### Actual Result

_No response_

9.1

2.0.0

8.4

www.hostinato.it

tblivet:
| Questions         | Answers
| Description?      | The product search only fired on `keydown`, so pasting via mouse/context-menu (and drag-drop/autofill) never triggered a search.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/hummingbird/issues/974
| How to test?      | Follow the issue

### DIFF
--- a/src/js/modules/ps_searchbar.ts
+++ b/src/js/modules/ps_searchbar.ts
@@ -44,7 +44,7 @@ const initSearchbar = () => {
   const clearSearch = () => {
     if (searchInput?.value) {
       searchInput.value = '';
-      searchInput.dispatchEvent(new KeyboardEvent('keydown'));
+      searchInput.dispatchEvent(new Event('input', {bubbles: true}));
     }
     searchClear?.classList.add('d-none');
     // Ensure clear button is not tabbable when hidden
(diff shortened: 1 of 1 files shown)


Quetzacoalt91 (approved):

src/js/modules/ps_searchbar.ts
```diff
@@ -254,7 +254,70 @@ const initSearchbar = () => {
...
+              clearTimeout(blurTimeout);
+              blurTimeout = null;
+            }
+          });
+
+          link.addEventListener('blur', handleBlur);
+        });
+
+        // Close dropdown when clicking outside
+        window.addEventListener('click', (event: Event) => {
```
Codencode, line 297: @tblivet 
This event listener is registered every time a search returns products. As a result, multiple identical listeners are added to window after multiple searches, and every click will execute all of them. Could we move it outside `triggerSearch()` so it is registered only once during initialization?

ingridusta (approved):
Hi @tblivet, 

Fix works perfectly : 

https://github.com/user-attachments/assets/10ff2925-4021-40ff-9153-706334e6c1ed

https://github.com/user-attachments/assets/6d8b8e3a-5339-48e3-ae97-0c228760100b

https://github.com/user-attachments/assets/0e071d54-7e82-4f03-a465-1a752194e48b

It's QA approved ✅

src/js/modules/ps_searchbar.ts
Codencode: @tblivet, sorry, I just realized that my previous comment was still pending, so it wasn’t visible before the pull request was merged.

PR #1053, opened 2026-06-23
Add subcategories hook

Hlavtox:
| Questions         | Answers
| Description?      | Adds a new hook to display content in subcategory block. You can use it for blog posts, special filtered landing pages, appending links to "On sale" page etc.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | Try to hook something to it and see if it displays. Try it in categories that have no subcategories.

### Example usage
hook category

### Hummingbird usage

```
<a class="subcategory__link subcategory__link--with-image" 
   style="background: #d4e5ff;border: 1px solid #8cbaff;" href="#">
    <span class="subcategory__name">
        <i class="material-icons" style="font-size: 2.2rem;color: #1765dd;">book</i>
        How to choose a sink
    </span>
</a>
```

hook

### DIFF
--- a/templates/catalog/_partials/category-header.tpl
+++ b/templates/catalog/_partials/category-header.tpl
@@ -34,9 +34,7 @@
         </div>
       {/if}
 
-      {if isset($subcategories) && $subcategories|@count > 0}
-        {include file='catalog/_partials/subcategories.tpl' subcategories=$subcategories}
-      {/if}
+      {include file='catalog/_partials/subcategories.tpl' subcategories=$subcategories|default:[]}
     </div>
   {/if}
 </div>
--- a/templates/catalog/_partials/subcategories.tpl
+++ b/templates/catalog/_partials/subcategories.tpl
@@ -4,58 +4,73 @@
  *}
 {$componentName = 'subcategory'}
 
+{**
+ * Determine if at least one subcategory has a thumbnail image.
+ * If so, set 'displaySubcategoryImages' to true so the list will render with images.
+ *}
+{assign var=displaySubcategoryImages value=false}
 {if !empty($subcategories)}
-  {**
-   * Determine if at least one subcategory has a thumbnail image.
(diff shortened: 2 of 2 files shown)


tblivet (changes requested):
Hi @Hlavtox nice addition, the hook is a useful extension point 👍 Does this hook already exist in PrestaShop core?

The partial is only included from `category-header.tpl` behind this guard:
```
{if isset($subcategories) && $subcategories|@count > 0}
  {include file='catalog/_partials/subcategories.tpl' subcategories=$subcategories}
{/if}
```

So on a category with no subcategories, `subcategories.tpl` is never included so `{hook h='displaySubcategories'}` never runs and nothing displays.

Two changes needed:
1. Maybe we should always include the partial and let it decide what to render: `{include file='catalog/_partials/subcategories.tpl' subcategories=$subcategories|default:[]}`
2. Output the hook content with nofilter so the module HTML isn't escaped: `{$smarty.capture.displaySubcategoriesContent nofilter}`

With these changes i think we are good 👍

tblivet:
@Hlavtox this should target the `2.x` branch rather than `develop` i think, could you update the base branch? 🙏

Hlavtox:
@tblivet Hello! :-) I applied the fixes, rebased the PR and I added a parameter to the hook, if it should render with images or not. What do you think now?

> Does this hook already exist in PrestaShop core?

No no, I need to create a PR for that so it's added to `hook.xml`.

Hlavtox:
@tblivet I just tested it with some code and it seems to work perfectly, I added a screenshot to the PR description. 👍

tblivet (approved):
Thank you @Hlavtox looks good 👍

ingridusta (approved):
Hi @Hlavtox,

Thanks for your contribution. 

I tested your PR with @tblivet 's help and it works as expected : 

Image

It's QA approved ✅

PR #1054, opened 2026-06-23
Fix product variants legend markup in Hummingbird

Codencode:
| Questions         | Answers
| Description?      | This PR fixes the product variants markup in the Hummingbird theme to keep the `legend` as a direct child of `fieldset`, which is the correct semantic and accessible HTML structure.<br/>Previously, the legend was wrapped inside a `<div>`, making the markup invalid. This change removes that wrapper and slightly adjusts the related spacing in SCSS by moving the margin to `.product-variant__legend`, preserving the same visual separation with valid markup.
| Type?            | bug fix / improvement / new feature / refactor
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 1. Open a product page using the Hummingbird theme.<br/>2. Verify each variant group renders legend directly inside fieldset.<br/>3. Confirm the selected value still displays correctly and spacing is unchanged.

### DIFF
--- a/src/scss/prestashop/components/product/_product-variant.scss
+++ b/src/scss/prestashop/components/product/_product-variant.scss
@@ -5,9 +5,12 @@ $component-name: product-variant;
   &__selected {
     display: inline-block;
     width: auto;
-    margin-block-end: 0.5rem;
     font-size: var(--bs-body-font-size);
     line-height: var(--bs-body-line-height);
+  }
+
+  &__attributes {
+    margin-block-start: 0.5rem;
   }
 
   // Radios
--- a/src/scss/prestashop/components/product/_product-variant.scss
+++ b/src/scss/prestashop/components/product/_product-variant.scss
@@ -5,6 +5,7 @@ $component-name: product-variant;
   &__selected {
     display: inline-block;
     width: auto;
+    margin-block-end: 0;
     font-size: var(--bs-body-font-size);
     line-height: var(--bs-body-line-height);
   }
--- a/src/scss/prestashop/components/product/_product-variant.scss
+++ b/src/scss/prestashop/components/product/_product-variant.scss
@@ -1,15 +1,11 @@
 $component-name: product-variant;
 
 .#{$component-name} {
-  &__label {
-    margin-block-end: 0.5rem;
-  }
-
   &__legend,
   &__selected {
     display: inline-block;
     width: auto;
-    margin-block-end: 0;
+    margin-block-end: 0.5rem;
     font-size: var(--bs-body-font-size);
     line-height: var(--bs-body-line-height);
   }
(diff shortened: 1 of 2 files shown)


ps-jarvis:

- `Shop.Theme.Catalog`
    - [ ] `:`

tblivet (commented):
Hi @Codencode, thanks!

My only concern is the spacing with margin-block-end now on both `product-variant__legend` and `product-variant__selected` (both inline-block), we can get unwanted vertical space when the selected value is long enough to wrap to a new line.

What about moving the `product-variant__selected` span inside the `<legend>`? That way the label and its selected value are a single block, we keep valid markup (legend stays the first child of the fieldset), and we can put the margin-block-end on `product-variant__legend` only.

And if we go with this, don't forget to set `font-weight: 400` on `product-variant__selected`!

WDYT ?

Codencode:
Hi @tblivet,

Although we can technically include the selected value inside the `<legend>` and use `aria-hidden="true"` to prevent it from becoming part of the accessible name, I think it is semantically more appropriate to keep it outside the legend.

The legend should contain only the stable name of the group, such as `Color`, while `Black` represents the current selected value rather than the group label itself.

A possible alternative could be to wrap the attribute controls in a common container and apply the vertical spacing to that container instead. We could then remove the `margin-bottom` from both the label and the legend, for example:

```html
<fieldset class="product-variant">
  <legend>Color</legend>
  <span class="product-variant__selected" aria-hidden="true">: Black</span>

  <div class="product-variant__attributes">
    <!-- select, color swatches or radio buttons -->
  </div>
</fieldset>
```

```scss
.product-variant__legend,
.product-variant__selected {
  margin-block-end: 0;
}

.product-variant__attributes {
  margin-block-start: 0.5rem;
}
```

This would keep the group name and the selected value semantically separate, while applying the spacing consistently before the actual attribute controls.

WDYT?

tblivet:
@Codencode, I'm okay with using the `<div class="product-variant__attributes">` wrapper and `margin-block-start` 👍

Codencode:
@tblivet done.

src/scss/prestashop/components/product/_product-variant.scss
```diff
@@ -1,19 +1,18 @@
...
 .#{$component-name} {
-  &__label {
-    margin-block-end: 0.5rem;
-  }
-
   &__legend,
   &__selected {
     display: inline-block;
     width: auto;
-    margin-block-end: 0;
```
tblivet, line 12: @Codencode, we should keep `margin-block-end: 0;` to reset the` __legend` margin applied by bootstrap reboot and `.form-label`.

tblivet (changes requested):
Thank you @Codencode  🙏 
I only have one suggested change.

src/scss/prestashop/components/product/_product-variant.scss
Codencode: @tblivet you're right!
I've added it, can you check?

Thanks!

tblivet (approved):
Perfect 🙏

ingridusta (approved):
Hi @Codencode, 

I reproduced the behaviour before your fix : 

Image

And with your fix : 

Image

No regression detected ! It's QA approved ✅

PR #1055, opened 2026-06-23
Add a hook to display content under cart summary

Hlavtox:
| Questions         | Answers
| Description?      | Adds a hook that displays content under the summary on cart page. I don't want to mix it with reassurance hook because that would need conditioning out to avoid displaying it on other pages.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 

### Example usage
Snímek obrazovky 2026-06-23 141411

### DIFF
--- a/templates/checkout/cart.tpl
+++ b/templates/checkout/cart.tpl
@@ -53,7 +53,9 @@
 
         <hr>
 
-        {hook h='displayCartBelowSummary'}
+        {block name='cart_below_summary'}
+          {hook h='displayCartBelowSummary'}
+        {/block}
 
         {block name='hook_reassurance'}
           {hook h='displayReassurance'}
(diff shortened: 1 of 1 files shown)


templates/checkout/cart.tpl
```diff
@@ -53,6 +53,8 @@
 
         <hr>
 
+        {hook h='displayCartBelowSummary'}
```
tblivet:
```suggestion
        {block name='cart_below_summary'}
          {hook h='displayCartBelowSummary'}
        {/block}
```

tblivet (changes requested):
HI @Hlavtox, I think we should wrap it inside a {block} !

Hlavtox:
@tblivet Done! :)

tblivet (approved):
Thank you @Hlavtox 😄

ingridusta (approved):
Hi @Hlavtox,

Works as expected : 

Image

Thanks for your contribution, it's QA approved ✅

PR #1056, opened 2026-06-24
Fix register searchbar outside-click listener once instead of on every search

tblivet:
| Questions         | Answers
| Description?      | Register the searchbar outside-click listener once at init instead of re-adding it on every search
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | --


Codencode (approved):
It seems to work well.
Awesome!
Thanks @tblivet!

PR #1063, opened 2026-07-05
Render the form field attr attributes in the form fields partial

boo-code:
| Questions         | Answers
| Description?      | Core's `FormField` carries an `attr` array (exported by `FormFieldCore::toArray()`), and core runs the `additionalHtmlAttributesFormFields` hook so modules can add HTML attributes to a form field — but `_partials/form-fields.tpl` never renders them, so everything placed in that slot is silently dropped. This renders the `attr` map on the generic input branch (`form_field_item_other`): each entry becomes an HTML attribute (name and value escaped), with `|default:[]` keeping the output byte-identical for callers that set no attributes. Concrete consumer: ps_onepagecheckout sets `inputmode="numeric"` on the postcode field for digits-only zip formats (PrestaShop/ps_onepagecheckout#101), so touch devices open the numeric keyboard — a standard mobile checkout keyboard optimization (https://baymard.com/blog/mobile-checkout). Verified end-to-end on a 9.2 shop against this change.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     |
| Sponsor company   |
| How to test?      | Render any form through `{form_field field=$field}` — unchanged markup. Set `$field.attr = ['inputmode' => 'numeric']` server-side (`FormField::setAttr()`) — the input renders `inputmode="numeric"`. With PrestaShop/ps_onepagecheckout#101 on a 9.2 shop, the checkout postcode inputs carry `inputmode="numeric"` for France.

### DIFF
--- a/templates/_partials/form-fields.tpl
+++ b/templates/_partials/form-fields.tpl
@@ -2,10 +2,15 @@
  * For the full copyright and license information, please view the
  * LICENSE.md file that was distributed with this source code.
  *}
+{capture assign='field_attr'}
+  {foreach from=$field.attr|default:[] key="attrName" item="attrValue"}
+    {$attrName|escape:'html'}="{$attrValue|escape:'html'}"
+  {/foreach}
+{/capture}
 {if $field.type == 'hidden'}
 
   {block name='form_field_item_hidden'}
-    <input type="hidden" name="{$field.name}" value="{$field.value}">
+    <input type="hidden" name="{$field.name}" value="{$field.value}" {$field_attr nofilter}>
   {/block}
 
 {else}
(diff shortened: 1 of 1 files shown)


ps-jarvis:
Hello @boo-code!

tblivet (changes requested):
Hi @boo-code 👋
A few suggestions before merge!

1.  I think we should apply it to all the other single-element field types too, and make the loop into one `{capture}` at the top to avoid repeating it:
```smarty
{capture assign='field_attr'}
  {foreach from=$field.attr|default:[] key="attrName" item="attrValue"}
    {$attrName|escape:'html'}="{$attrValue|escape:'html'}" 
  {/foreach}
{/capture}
```
2. Then put {$field_attr nofilter} into each element.
3. And place it at the end of the tag. Since HTML keeps the first occurrence of a duplicate attribute, putting the result of the loop last means `attr` can only add attributes and never override name/type/required/etc.

Note: radio-buttons, checkbox and birthday are different, as they're kind of multiple inputs, so we should leave those out.

boo-code:
Done in f84328b2.

- Hoisted the loop into a single `{capture assign='field_attr'}` at the top.
- Output `{$field_attr nofilter}` last on each single-element field, so `attr` can only add attributes and never override `name`/`type`/`required`.
- Left `radio-buttons`, `checkbox` and `birthday` out, as they render multiple inputs.

One call I made: I also applied it to the `hidden` input, since it's a single element too. Happy to drop that if you'd rather keep `hidden` untouched.

tblivet (approved):
Thank you 👍

tblivet:
@ingridusta this could be usefull to test the PR [hbqaformattr.zip](https://github.com/user-attachments/files/29849740/hbqaformattr.zip)

ingridusta (approved):
Hi, 

Image

Thanks @tblivet for the module to test,, so useful ! 
It's QA approved ✅

PR #1065, opened 2026-07-06
Replace outer nav element with div in pagination template

Codencode:
| Questions         | Answers
| Description?      | The pagination template contained a nested `<nav>` element, where the outer `<nav>` was used only as a layout container.<br/>This PR replaces the outer `<nav>` with a `<div>`, keeping the inner `<nav>` with its `aria-label`. This improves the semantic HTML structure and avoids nested navigation landmarks while preserving the existing layout and behavior.| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 1. Open any front office page with pagination (for example a category with multiple product pages).<br/>2. Verify that the pagination is displayed and works as before.<br/>3. Inspect the generated HTML and check that the outer pagination container is now a `<div>` while the pagination navigation is still wrapped in a single `<nav aria-label="Products pagination">`.

### DIFF
--- a/templates/_partials/pagination.tpl
+++ b/templates/_partials/pagination.tpl
@@ -4,7 +4,7 @@
  *}
 {$componentName = 'pagination'}
 
-<nav class="{$componentName}__container">
+<div class="{$componentName}__container">
   <div class="{$componentName}__number">
     {block name='pagination_summary'}
       {l s='Showing %from%-%to% of %total% item(s)' d='Shop.Theme.Catalog' sprintf=['%from%' => $pagination.items_shown_from ,'%to%' => $pagination.items_shown_to, '%total%' => $pagination.total_items]}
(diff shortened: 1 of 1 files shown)


tblivet (changes requested):
Hi @Codencode 👋

Good catch on the nested `<nav>` 👍

But only the opening tag was changed, so the last line still closes with `</nav>` could you also change that final `</nav>` to `</div>` please?

Codencode:
You're right, I hadn't noticed that 🤦‍♂️ !
Thank you.

I've applied the change.

tblivet (approved):
Perfect thank you @Codencode 🙏

ingridusta (approved):
Hi @Codencode, 

Thanks for your PR. 

I reproduced the bug : 

Image

<br>And with your PR it's all fixed ✨ : 

Image

It's all good and QA approved ✅

PR #1067, opened 2026-07-07
Render radio attribute groups as buttons on the product page

### LINKED ISSUE #1066: Render radio attribute groups as buttons on the product page (Baymard: use buttons for size selection)
### Summary

For a `radio`-type attribute group, `_partials/product-variants.tpl` renders each option as a plain `form-check` radio (a small radio circle plus a text label). Baymard's product-page research recommends that variation options — size in particular — be exposed as button-like selectors so all options are visible at a glance and are easy to tap: "Always Use 'Buttons' for Size Selection (28% of Desktop Sites Don't)" (https://baymard.com/blog/use-buttons-for-size-selection). Plain radios are not that pattern.

The theme already renders the `color` group type as swatches and `select` as a dropdown; only the `radio` type is left as bare radios, so a merchant who follows the guidance and configures size as a radio attribute still does not get a button UI.

### Proposed fix

Render the `radio` branch with Bootstrap's `btn-check` pattern (which the theme already uses elsewhere): a visually-hidden radio input (`class="btn-check"`, keeping `type="radio"`, `name`, `value`, `data-product-attribute`, `aria-labelledby` and the checked state) plus a `<label class="btn btn-outline-primary">`. The `role="radiogroup"` container and radio semantics are preserved, so keyboard/arrow-key selection and the variation-sync JS (which listens on `input[data-product-attribute]`) keep working; `select` and `color` are untouched.

A pull request implementing this is linked below; verified on a 9.2 shop by setting the demo Size attribute group to `radio` — the four sizes render as an outline-button group, selecting one still syncs price/URL/availability, and add-to-cart still works.

### Note

For the default Baymard-compliant experience the demo Size attribute group would also need to ship as `radio` rather than `select` (currently a dropdown) — that is a PrestaShop core install-data decision, separate from this theme change.

boo-code:
| Questions         | Answers
| Description?      | A `radio`-type attribute group renders each option as a plain `form-check` radio (small circle + text), while Baymard's product-page research recommends variation options — size in particular — be exposed as button-like selectors (https://baymard.com/blog/use-buttons-for-size-selection, "Always Use 'Buttons' for Size Selection (28% of Desktop Sites Don't)"). The theme already renders `color` as swatches and `select` as a dropdown; only `radio` was left as bare radios. This renders the `radio` branch with Bootstrap's `btn-check` pattern (already used elsewhere in the theme): a visually-hidden radio input (keeping `type="radio"`, `name`, `value`, `data-product-attribute`, `aria-labelledby` and the checked state) plus a `<label class="btn btn-outline-primary">`. `role="radiogroup"` and radio semantics are preserved, so keyboard/arrow-key selection and the variation-sync JS keep working; `select` and `color` are untouched. Fixes #1066.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #1066
| Sponsor company   |
| How to test?      | Set an attribute group used by a product to `group_type = radio` in the BO (or `UPDATE ps_attribute_group SET group_type='radio'` for e.g. the demo Size group), open that product: the options render as an outline-button group with all values visible, the selected value shows a filled button, selecting one syncs price/URL/availability, and add-to-cart still works. Verified live on a 9.2 shop. `select` and `color` group types render exactly as before.

### DIFF
--- a/src/scss/prestashop/components/product/_product-variant.scss
+++ b/src/scss/prestashop/components/product/_product-variant.scss
@@ -18,8 +18,7 @@ $component-name: product-variant;
   &__radios {
     display: flex;
     flex-wrap: wrap;
-    row-gap: 0.5rem;
-    column-gap: 0.875rem;
+    gap: 0.5rem;
     margin-bottom: 0;
   }
 
--- a/templates/catalog/_partials/product-variants.tpl
+++ b/templates/catalog/_partials/product-variants.tpl
@@ -71,19 +71,20 @@
                   {assign var=inputId value="input_{$id_attribute_group}_{$id_attribute}_{$product.id}"}
                   {assign var=labelId value="label_{$id_attribute_group}_{$id_attribute}_{$product.id}"}
 
-                  <div class="product-variant__radio form-check">
+                  <div class="product-variant__radio">
                     <input
-                      class="form-check-input"
+                      class="btn-check"
                       type="radio"
                       id="{$inputId}"
                       data-product-attribute="{$id_attribute_group}"
                       name="group[{$id_attribute_group}]"
                       value="{$id_attribute}"
                       aria-labelledby="{$labelId}"
+                      autocomplete="off"
                       {if $group_attribute.selected} checked="checked" aria-checked="true"{/if}
                     >
-                    <label for="{$inputId}">
-                      <span class="form-check-label" id="{$labelId}"><span class="visually-hidden">{$group.group_name} - </span>{$group_attribute.name}</span>
+                    <label class="btn btn-outline-primary" for="{$inputId}">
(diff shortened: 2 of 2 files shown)


ps-jarvis:
Hello @boo-code!

tblivet (changes requested):
Hi @boo-code, thanks for the improvement 👍 
I only see one thing to adapt in the `_product-variant.scss` file, I believe we should replace `row-gap: 0.5rem;` and `column-gap: 0.875rem;` with `gap: 0.5rem;` please!

boo-code:
Done in c6c003bf — replaced `row-gap`/`column-gap` with a single `gap: 0.5rem;` on `.product-variant__radios`.

tblivet (approved):
Thank you @boo-code !

ingridusta (approved):
Hi, 

Before : 

Image

After, with ur pr : 

Image

It's QA approved ✅

PR #1087, opened 2026-07-29
Use new image feature

Hlavtox:
| Questions         | Answers
| Description?      | Puts the new https://github.com/PrestaShop/PrestaShop/pull/41977 feature to use. Adds new image dimension for manufacturers and uses it on the product page.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | See that the new image doesn't have whitespace anymore.

### DIFF
--- a/config/theme.yml
+++ b/config/theme.yml
@@ -219,6 +219,11 @@ global_settings:
       width: 2000
       height: 400
       scope: [categories]
+    manufacturer_product_page:
+      width: 98
+      height: 98
+      scope: [manufacturers]
+      image_fitment: bound
 
 theme_settings:
   rtl_generation: false
--- a/templates/catalog/_partials/product-details.tpl
+++ b/templates/catalog/_partials/product-details.tpl
@@ -25,14 +25,14 @@
                 </div>
 
                 <div class="details__right">
-                  {if isset($product_manufacturer.image.bySize.small_default.url)}
+                  {if isset($product_manufacturer.image.bySize.manufacturer_product_page.url)}
                     <a href="{$product_manufacturer->url}">
-                      <img src="{$product_manufacturer.image.bySize.small_default.url}"
+                      <img src="{$product_manufacturer.image.bySize.manufacturer_product_page.url}"
                         class="img-fluid details__manufacturer-logo"
                         alt="{$product_manufacturer->name}"
                         loading="lazy"
-                        width="{$product_manufacturer.image.bySize.small_default.width}"
-                        height="{$product_manufacturer.image.bySize.small_default.height}"
+                        width="{$product_manufacturer.image.bySize.manufacturer_product_page.width}"
+                        height="{$product_manufacturer.image.bySize.manufacturer_product_page.height}"
                         aria-label="{l s='Brand: %brand_name%' sprintf=['%brand_name%' => $product_manufacturer->name] d='Shop.Theme.Catalog'}"
                       >
                     </a>


SharakPL (dismissed):

Hlavtox:
@Codencode @kpodemski @tblivet Guys, one more thing we could improve in the image fitment feature.

Right now, width and height of the image type are passed into the FO, but in case of images set to "bound", they will not contain right values. I think a good solution would be to `@getimagesize` of the real thumbnail. WDYT?

SharakPL:
The change is required but I suspect this change would cause a performance drop on long product lists. Isn't `width="auto" height="auto"` enough for bound thumbs?

To prevent layout shifts we could add `aspect-ratio` to parent link based on the dimensions from image settings.

For singular images like the manufacturer on the product page `@getimagesize` should be ok.

Codencode:
I would avoid using `getimagesize()` at rendering time, as it would introduce additional filesystem reads for every image and could become a performance issue, especially on product listings.

Using `width: auto` and `height: auto` could be an option, but only as CSS properties. The HTML `width` and `height` attributes are expected to contain numeric values. However, this would prevent image distortion without necessarily solving layout shifts, since the browser would still not know the actual image dimensions in advance.

Ideally, the actual generated dimensions should be stored when the thumbnail is created. They could then be exposed by the image presenter without having to inspect each file again on every request.

Codencode (dismissed):

SharakPL:
I agree. Storing image dimensions in DB would be the best option.

BTW. a reminder to also consider a fix for [#34665](https://github.com/PrestaShop/PrestaShop/issues/34665) while on this subject.

Hlavtox:
@Codencode I tested it once and the dimensions retrieval take like 0.00002 ms or something. I will do some measurements.

> I agree. Storing image dimensions in DB would be the best option.

You would need to store dimensions of all thumbnails, it would insanely complicated and slower than getting the image header.

Codencode:
@Hlavtox

We could consider retrieving the dimensions of just one image (for example, the original one) and then calculating the dimensions of all the other images from it. I think this approach would be preferable to retrieving the dimensions of every single image.

Another option would be to store the dimensions of only one image and derive the others through calculation. This shouldn't be an overly complex change: we would only need to store `width` and `height` by adding two fields to the `image` table.

What do you think?

Hlavtox:
So, I did some measurements.

### Test 1
```php
$images = Db::getInstance()->executeS('
  SELECT id_image
  FROM `' . _DB_PREFIX_ . 'image`
  LIMIT 1000
');

foreach ($images as $row) {
	$imagePath = implode(DIRECTORY_SEPARATOR, [
    rtrim(_PS_PRODUCT_IMG_DIR_, DIRECTORY_SEPARATOR),
    rtrim(Image::getImgFolderStatic((int) $row['id_image']), DIRECTORY_SEPARATOR),
  ]) . (int) $row['id_image'] . '-large_default.jpg';


  echo (int) $row['id_image'] . ' => ' . $imagePath . "<br/>";
}
```
1.162 ms

### Test 2

```php
$images = Db::getInstance()->executeS('
  SELECT id_image
  FROM `' . _DB_PREFIX_ . 'image`
  LIMIT 1000
');

foreach ($images as $row) {
	$imagePath = implode(DIRECTORY_SEPARATOR, [
    rtrim(_PS_PRODUCT_IMG_DIR_, DIRECTORY_SEPARATOR),
    rtrim(Image::getImgFolderStatic((int) $row['id_image']), DIRECTORY_SEPARATOR),
  ]) . "/" . (int) $row['id_image'] . '-medium_default.jpg';

	$size = @getimagesize($imagePath);
	if ($size !== false) {
		echo (int) $row['id_image'] . ' => ' . $size[0] . ' x ' . $size[1] . ' px' . "<br/>";
	} else {
		echo (int) $row['id_image'] . ' => Cannot load dimensions.' . "<br/>";
	}
}
```
85.508 ms

So, it takes 0,085 ms to get dimensions of one image.

Codencode:
The benchmark is reassuring, and `getimagesize()` may be perfectly acceptable in many cases.

My only concern is that filesystem access can behave differently depending on the environment, for example with virtual machines, Docker volumes, network storage, or slower hosting.

This could matter mainly for product images, since a single page may render many of them through product listings, homepage carousels, related products, or product-page recommendations.

For this reason, I was considering storing only the original product image dimensions in the `image` table. The dimensions of `bound` thumbnails could then be calculated from the original size and the image type configuration, without storing every generated size or repeatedly accessing the filesystem.

tblivet (changes requested):
Hi 👋 sorry, I blocked the PR just to avoid an accidental merge before a few questions
get answered 👍 But it's nice to see the new fitment being used 🙂

Small question about removing `width` / `height` on the ``: would it be possible
to keep them?

Without them the browser can't reserve space for the image, so it shifts the layout
when it loads (CLS). It's not really an issue in this specific case, but if `bound`
ends up being widely used across the theme, systematically dropping the dimensions
would become a habit worth avoiding.

For that to work, i believe both real values would need to be populated by the core with the actual generated dimensions, since today `bySize` returns the ones configured in the theme (98×98) rather than those of the generated file. WDYT ?

Hlavtox:
@tblivet Exactly, if I use the width and height provided by the image type, I am not getting the real image dimensions.

Since the usage will be very low by default, I think we can start safely by adapting ImageRetriever to get the dimensions with `@getimagesize`, will it be OK for 9.2?

tblivet:
Well thank you @Hlavtox 🙂
So from the theme side, all we need is `bySize` returning the real dimensions.

@kpodemski maybe you can confirm whether it's OK to add this to the 9.2 scope?

tblivet (dismissed):

Hlavtox:
@tblivet The real dimensions for `bound` thumbnails are ready in PrestaShop/PrestaShop#42349. I restored the `width` and `height` attributes here so the theme can use them. 🚀

SharakPL (approved):

tblivet (approved):
Perfect 👌 Thank you @Hlavtox !

Codencode (approved):

tblivet:
✅ **QA: approved.** Tested on PrestaShop 9.3.0, PHP 8.1.33, Hummingbird 2.1.0, front office.

Measured on the merge base, then on this branch, same scenario both times. Fixture: product 1, brand Studio Design, uploaded logo 189x88.

| | brand logo served | decoded | aspect ratio |
| --- | --- | --- | --- |
| merge base | `1-small_default.jpg` | 98x98 | 1.000 |
| this branch | `1-manufacturer_product_page.jpg` | 98x46 | 2.130 |

The padding is gone and the thumbnail now matches the shape of the uploaded file. The brands page, the front page, a product page and the cart all still render, and both narrow widths (375 and 768) stay at 0px sideways overflow on both sides.

Two things worth a look before merge, neither of which changes the result:

1. The markup still declares `height="{$product_manufacturer.image.bySize.manufacturer_product_page.height}"`, and core returns the image type's height (98), not the generated file's (46). On a cold load the browser reserves a 98x98 box and then collapses it: the Brand row goes from 131px to 79px, a 52px layout shift. Dropping the `height` attribute, or deriving it from the generated thumbnail, would avoid it.
2. The new image type only reaches `ps_image_type` when the theme is applied, and this PR does not bump the theme version, so no update is offered. On a shop that just pulls the new files the `{else}` branch fires and the brand renders as a plain text link instead of a logo. Worth a line in the release notes.

clip-04-the-brand-row-of-the-product-details-block

clip-04-the-brand-row-of-the-product-details-block

Hlavtox:
@tblivet Thanks!

About the update - let's bump the theme version to 2.1.1 then?

tblivet:
Hi @Hlavtox ! Yes, we will need 2.1.1 soon. I want to review and QA the last PRs first. No need to rush on it!

fdonnet:
@Hlavtox  Hi, don't know if I can ask you that here. But for people like me that forked the theme (for custom things) and actually taking care of cherry picking all the new commits from you (upstream). Do I need to do something special (create new image types in DB, regenerate thumbnails) ?

Hlavtox:
Hello @fdonnet, if you want to use this feature, you need be running current 9.2 code, with the https://github.com/PrestaShop/PrestaShop/pull/41977 feature.

Then, if you apply a change from this PR, make sure to go to Image settings and create the new image dimension for manufacturers

```
manufacturer_product_page
width: 98
height: 98
scope: [manufacturers]
image_fitment: bound
```

And it should work. :-)

BTW, we need to do one last change in https://github.com/PrestaShop/PrestaShop/pull/42349, for the width and height of the image to be correct for these new "dynamic" thumbnails.

PR #1089, opened 2026-08-07
Add productcomments summary template

mgielecinski:
| Questions         | Answers
| Description?      | This PR provides template changes to the productcomments module for this contribution PrestaShop/productcomments#222
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | 

obraz

### DIFF
--- a/modules/productcomments/views/templates/hook/product-comments-list.tpl
+++ b/modules/productcomments/views/templates/hook/product-comments-list.tpl
@@ -10,20 +10,69 @@
   </script>
 
   <div id="product-comments-list-header">
-    <h2 class="section-title">
-      {l s='Comments' d='Modules.Productcomments.Shop'} ({$nb_comments})
-    </h2>
-    {include file='module:productcomments/views/templates/hook/average-grade-stars.tpl' grade=$average_grade showGradeAverage=true showNbComments=false}
+    <h2>{l s='Comments' d='Modules.Productcomments.Shop'}</h2>
+    {if $nb_comments > 0 && $post_allowed}
+      <div id="product-comments-list-btn-group">
+        <button class="w-100 w-sm-auto btn btn-outline-primary post-product-comment" id="product-comments-list-review-button" type="button" data-bs-toggle="modal" data-bs-target="#post-product-comment-modal" data-ps-ref="product-post-review-button">
+          {l s='Write your review' d='Modules.Productcomments.Shop'}
+        </button>
+      </div>
+    {/if}
   </div>
+ 
+  {if $nb_comments > 0}
+    <div class="product-comments-summary">
+      <div class="product-comments-summary__left">
+        <div class="product-comments-summary__score-container">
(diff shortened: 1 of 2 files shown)


tblivet (changes requested):
Thanks @mgielecinski for this, it's a really nice addition to the product page 👌

I have a set of small changes, mostly theme consistency (keeping the section-title component, rem instead of px, logical properties) plus a guard so the summary degrades cleanly on older versions of the module.

Most of them are easier to read as a diff than to explain. Want me to push them to your branch, or open a PR against `productcomments/add-grade-summary` so you can review first? 

 If you'd rather handle it yourself, I'll detail everything in a review instead 👍

mgielecinski:
Hi @tblivet 
I tried to improve everything you mentioned in your previous message. If I missed something or there's still something that needs to be fixed, I'd be very grateful if you could push the corrections.
Thanks :heart:

Hlavtox (dismissed):

ps-jarvis:

- `Modules.Productcomments.Shop`
    - [ ] `/5.0`

tblivet:
Thanks @mgielecinski for addressing most of the feedback 🙏
I've just pushed a few styles improvements 👍

mgielecinski:
It looks ok with and without the summary :thumbsup:
Thank you @tblivet  for improving this contribution! :rocket: :heart:

tblivet:
Hi 👋 this report is an AI-generated QA review, currently in a testing phase.

--

**QA: 🟢 approved** — but tested **with `productcomments#222` applied**, since all the new markup is gated on `isset($summary)`. **Do not merge before https://github.com/PrestaShop/productcomments/pull/222**: alone, this PR is inert (verified — legacy layout intact, empty state fine, no errors).

Env: PS 9.3.0, PHP 8.1.33, hummingbird 2.1.0, Chromium headless at 1280×900 and 390×844. Seeded 12 approved reviews (5×5, 4×3, 3×2, 2×1, 1×1 — avg 3.8).

Verified in the browser:
- breakdown renders 5→1, one line per rating
- per-rating counts sum to the stated total
- `3.8` matches the seeded reviews, shown `/5.0`
- all five bars proportional within 2% of count/total (measured geometry, not read off the `style` attribute)
- review button moves into the section header
- product with no reviews → empty state, no breakdown
- 390px keeps bar widths, no sideways scroll
- 0 console errors and 0 failed requests, with and without the dependency

Not a blocker here, but worth knowing before both ship: grades average across criteria, so they can be `4.5`, and #222 does `$summary[(int) $grade] = [...]` over a float `GROUP BY` — buckets overwrite. Adding one 4.5 review to the twelve gave "Based on 13 reviews" with counts summing to 10 and the 4-star row at 1. Reported on #222.

Nits: `{l s='/5.0'}` is a translatable string with no words in it (#222 hardcodes the same text, so the two disagree); the per-rating counts are `aria-hidden`, so screen readers get percentages and never the counts.

PR #1090, opened 2026-08-10
Update breadcrumb styles

tblivet:
| Questions         | Answers
| Description?      | Regarding this CORE PR https://github.com/PrestaShop/PrestaShop/pull/42189, a small adjustment is needed to make the breadcrumb display correctly on error pages.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | https://github.com/PrestaShop/PrestaShop/issues/40369

### DIFF
--- a/src/scss/bootstrap/components/_breadcrumb.scss
+++ b/src/scss/bootstrap/components/_breadcrumb.scss
@@ -11,10 +11,6 @@ $component-name: breadcrumb;
   overflow-x: scroll;
   scrollbar-width: none;
 
-  &[data-depth="1"] {
-    display: none;
-  }
-
   &-item {
     padding: 0;
     white-space: nowrap;
--- a/src/scss/prestashop/pages/_index.scss
+++ b/src/scss/prestashop/pages/_index.scss
@@ -9,6 +9,5 @@
 @import "login";
 @import "order-details";
 @import "order-returns";
-@import "page-not-found";
 @import "pages";
 @import "product";
--- a/src/scss/prestashop/pages/_page-not-found.scss
+++ b/src/scss/prestashop/pages/_page-not-found.scss
@@ -1,9 +1,0 @@
-/**
- * For the full copyright and license information, please view the
- * LICENSE.md file that was distributed with this source code.
- */
-#pagenotfound {
-  .wrapper {
-    padding-top: 3rem;
-  }
-}
--- a/templates/errors/404.tpl
+++ b/templates/errors/404.tpl
@@ -4,8 +4,6 @@
  *}
 {extends file='page.tpl'}
 
-{block name='breadcrumb'}{/block}
-
 {block name='container_class'}container container--limited-md text-center{/block}
 
 {block name='page_header_container'}
--- a/templates/errors/410.tpl
+++ b/templates/errors/410.tpl
@@ -4,8 +4,6 @@
  *}
 {extends file='page.tpl'}
 
-{block name='breadcrumb'}{/block}
-
 {block name='container_class'}container container--limited-md text-center{/block}
 
 {block name='page_header_container'}


Hlavtox (approved):
If I understand correctly, it will only hide breadcrumbs with only "Home" right?

mattgoud (approved):

Codencode:
@tblivet I think this block should also be removed/modified: https://github.com/PrestaShop/hummingbird/blob/c8229621c7b9e432a8ec8498f282ea9452cc9130/templates/checkout/checkout.tpl#L10-L11

Because the PR https://github.com/PrestaShop/PrestaShop/pull/42189 adds the breadcrumb to the OrderController.php controller as well, or am I missing something?

tblivet:
Hi @Codencode, you're right that https://github.com/PrestaShop/PrestaShop/pull/42189 adds it to OrderController too, but I'd rather keep that block empty. `checkout-navigation.tpl` already tells the customer where they are in the checkout, and a breadcrumb whose only link goes back to Home is a kind of distraction / exit point on a page that should stay distraction-free. What do you think?

@Hlavtox to answer yours, the `&[data-depth="1"]` rule I removed was never applied at all, due to an incorrect selector. And now that the controller declares a proper breadcrumb, there's nothing left to hide anyway (the home page is already covered by the empty block).

Codencode:
@tblivet ok, you're right, better not to show it in that order, but at this point the code related to the Order controller should be removed in the PR: https://github.com/PrestaShop/PrestaShop/pull/42189?

tblivet:
@Codencode I'd keep the core PR as is. Core's job is to expose a correct breadcrumb and rendering it is a theme decision, i believe it's better to have the data available than to have themes unable to show it, this can be useful for custom themes.

Codencode:
@tblivet Okay, that's fine.

Codencode (approved):

tblivet:
🟢 **Approved.** The not-found page and the gone-product page now show a breadcrumb, like every other page of the shop.

Measured on PrestaShop 9.3.0, PHP 8.1.33, Hummingbird 2.1.0, at the merge base (`c1465f74`) and then at the PR head (`c9e33087`), rebuilding the theme in each state.

This result covers the pair, not this PR alone: core PR PrestaShop/PrestaShop#42189 was applied to the shop for both measurements. Without it the not-found page carries a one-level breadcrumb, and removing the theme suppression would show a lone "Home", which is the symptom of PrestaShop/PrestaShop#40369.

| Page | Before | After |
| --- | --- | --- |
| `/this-page-does-not-exist` | no breadcrumb | `Home / 404 error` |
| a product set to Gone (410) | no breadcrumb | `Home / Clothes / Men / Hummingbird printed t-shirt` |

Three further observations:

1. The `.breadcrumb[data-depth="1"] { display: none; }` rule this PR deletes never matched anything. `data-depth` is rendered on the `nav.breadcrumb__wrapper`, while `.breadcrumb` is the inner `ol`. That is why Hummingbird displays the empty breadcrumbs in PrestaShop/PrestaShop#40369 where Classic hides them, and deleting the rule changes nothing on its own.
2. Deleting `_page-not-found.scss` does not collapse the layout. The `#wrapper` element starts at the same offset in both states (122px), with the 48px `padding-top` replaced by the breadcrumb bar and its own margin.
3. The 410 page needs no core change: its breadcrumb comes from the product category path and already had four entries.

`checkout.tpl` still empties the breadcrumb block, so the checkout half of PrestaShop/PrestaShop#42189 has no visible effect in this theme. Read from the template, not observed: with an empty cart, `controller=order` redirects to the cart page.

Home page unchanged (no breadcrumb, as `index.tpl` empties the block). Smoke pass and the 375/768 checks are identical in both states.

Before:
03-open-a-url-that-does-not-exist
04-open-a-product-that-is-gone

After:
03-open-a-url-that-does-not-exist
04-open-a-product-that-is-gone

kpodemski:
@tblivet let's make sure this label is also green 👍🏻

tblivet:
@kpodemski Done! I’ve updated the QA labels in the repo 👍

PR #1091, opened 2026-08-12
Fix product schema image and price validity

mihailovs2000:
## Summary

- use the original product cover URL in `Product.image` instead of the 250 x 250 `home_default` thumbnail
- omit `Offer.priceValidUntil` when the catalog has no actual price-expiry date

## Why

`priceValidUntil` means the date after which the displayed price is no longer available. Generating an arbitrary date 15 days in the future publishes data that is not backed by catalog state and changes on every render. Omitting it is accurate when no validity date exists.

The full product cover is also a better structured-data image than the small listing thumbnail and is already exposed by the product presenter.

Reference: https://developers.google.com/search/docs/appearance/structured-data/product-snippet#offer-properties

## Test

- reviewed the rendered JSON-LD template paths for products with and without a cover
- verified the trailing commas remain valid with `priceValidUntil` removed

### DIFF
--- a/templates/_partials/microdata/product-jsonld.tpl
+++ b/templates/_partials/microdata/product-jsonld.tpl
@@ -31,7 +31,7 @@
     "name": {$product.name|json_encode nofilter},
     "description": {$page.meta.description|json_encode nofilter},
     "category": {$product.category_name|json_encode nofilter},
-    {if !empty($product.cover)}"image": {$product.cover.bySize.home_default.url|json_encode nofilter},{/if}
+    {if !empty($product.cover)}"image": {$product.cover.large.url|json_encode nofilter},{/if}
     "sku": {if $product.reference}{$product.reference|json_encode nofilter}{else}{$product.id|json_encode nofilter}{/if},
     "mpn": {if $product.mpn}{$product.mpn|json_encode nofilter}{elseif $product.reference}{$product.reference|json_encode nofilter}{else}{$product.id|json_encode nofilter}{/if}
     {if $product.ean13},"gtin": {$product.ean13|json_encode nofilter}{/if}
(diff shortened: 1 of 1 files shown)


ps-jarvis:
Hello @mihailovs2000!

Hlavtox:
Hello @mihailovs2000, starting with 9.2, [structured data are now sent from the core directly](https://github.com/PrestaShop/PrestaShop/pull/37552/changes). The images are made differently.

priceValidUntil is still there, but you can easily override it via a module. See the example module I made - https://github.com/PrestaShop/example-modules/tree/master/demoseo.

But, maybe we can remove it indeed. Or maybe improve it so that it reflects the validity of specific price, if the product is discounted?

tblivet:
Hi @mihailovs2000, thanks for the PR!

As @Hlavtox mentioned above, the images are built differently since 9.2 the core now does it in `ProductController::getStructuredData()` so the image part can be dropped.

And on the second point @Hlavtox raised, rather than removing `priceValidUntil`, I think we could improve the current code for when the product is discounted. It seems we already have the information available to make the data correct in the JSON-LD, since the value is exposed as `$product.specific_prices.to`.

So this to test and format the data:
```smarty
{assign var=priceValidUntil value=false}

{if !empty($product.specific_prices.to) && $product.specific_prices.to != '0000-00-00 00:00:00'}
  {assign var=priceValidUntil value=$product.specific_prices.to|date_format:'%Y-%m-%d'}
{/if}
```
and this to output it in the structure should be enough:
```smarty
{if $priceValidUntil}"priceValidUntil": {$priceValidUntil|json_encode nofilter},{/if}
```

What do you think, @mihailovs2000, @Hlavtox?

PR #1098, opened 2026-09-05
Respect the country tax label setting on the product page

boo-code:
| Questions         | Answers
| ----------------- | -------------------------------------------------------
| Description?    | The product page prints "No tax" whenever taxes are disabled without consulting the country's "Display tax label" setting, which the very next branch of the same template already honours. `$configuration.display_taxes_label` is `$this->context->country->display_tax_label` (`FrontController::getDisplayTaxesLabel()`), so a merchant who turns that checkbox off still gets a tax statement on every product page. Both branches are now gated on the setting.
| Type?           | bug fix
| BC breaks?      | no
| Deprecations?   | no
| Fixed ticket?   | Fixes https://github.com/PrestaShop/PrestaShop/issues/21484
| Sponsor company |
| How to test?    | Set a country's "Display tax label" to No (International > Locations > Countries), then disable taxes (International > Taxes > Enable taxes: No) and open a product page. Before: "No tax" is displayed. After: nothing is displayed. With the label set to Yes the behaviour is unchanged in both tax states.

### Before

```smarty
{if !$configuration.taxes_enabled}
  {l s='No tax' d='Shop.Theme.Catalog'}
{elseif $configuration.display_taxes_label}
  {$product.labels.tax_long}
{/if}
```

### After

```smarty
{if $configuration.display_taxes_label}
  {if $configuration.taxes_enabled}
    {$product.labels.tax_long}
  {else}
    {l s='No tax' d='Shop.Theme.Catalog'}
  {/if}
{/if}
```

### Empty block

With the label off and no ecotax, `product__tax-infos` rendered empty and still took the flex gap. It is now
rendered only when it has a label, an ecotax line or price hook output: the two `displayProductPriceBlock` hooks
are captured first and the gate tests the capture with `|trim`. In debug mode the `<!-- begin/end -->` comments
around an empty hook count as output, so a dev shop can still show the empty block; a production shop cannot.

### Why this shape

https://github.com/PrestaShop/PrestaShop/pull/25882 fixed the same defect for the cart summary and the
order confirmation table by gating on `display_taxes_label` together with `taxes_enabled`, and was merged
in October 2021. The product page was not part of it, which is why the issue is still open with this exact
template quoted in the thread. Using the same gate keeps the three places consistent rather than inventing
a rule for this one.

### Scope

Only the product page template changes. The cart summary and the order confirmation table already respect
the setting since https://github.com/PrestaShop/PrestaShop/pull/25882, and so does `DeliveryOptionsFinder`.

Companion to the same change in `classic-theme`: the two themes carry the same template and the same defect.

### DIFF
--- a/templates/catalog/_partials/product-prices.tpl
+++ b/templates/catalog/_partials/product-prices.tpl
@@ -57,10 +57,12 @@
 
         <div class="product__tax-infos">
           <span class="product__tax-label">
-            {if !$configuration.taxes_enabled}
-              {l s='No tax' d='Shop.Theme.Catalog'}
-            {elseif $configuration.display_taxes_label}
-              {$product.labels.tax_long}
+            {if $configuration.display_taxes_label}
+              {if $configuration.taxes_enabled}
+                {$product.labels.tax_long}
+              {else}
+                {l s='No tax' d='Shop.Theme.Catalog'}
+              {/if}
             {/if}
             
             {hook h='displayProductPriceBlock' product=$product type="price"}


tblivet (changes requested):
Hi @boo-code, the fix looks right, but I noticed one thing while testing: with the label off and no ecotax, the `product__tax-infos` div renders empty and still takes the `0.75rem` flex gap from `_product.scss`. Not something you introduced, but this adds another way to get there.

Adding a `hidden` attribute won't do it, since the block contains the two `displayProductPriceBlock` hooks, so it needs a capture. I have it working locally, I can push it here or do a follow-up, up to you.

boo-code:
@tblivet Done here: the two `displayProductPriceBlock` hooks are captured and `product__tax-infos` renders only with a label, an ecotax line or hook output. The capture is tested after stripping the `<!-- begin/end -->` comments debug mode wraps around rendered templates, otherwise an empty hook keeps the block on a dev shop. Rendered with empty hooks, comments-only hooks, hook content, label only and ecotax only: no block in the first two, the block in the other three.

templates/catalog/_partials/product-prices.tpl
```diff
@@ -55,32 +55,38 @@
...
-              {$product.labels.tax_long}
-            {/if}
-            
-            {hook h='displayProductPriceBlock' product=$product type="price"}
-            {hook h='displayProductPriceBlock' product=$product type="after_price"}
-          </span>
+        {capture name='product_price_hooks'}{hook h='displayProductPriceBlock' product=$product type="price"}{hook h='displayProductPriceBlock' product=$product type="after_price"}{/capture}
+        {* WHY: the block holds only optional parts, and an empty one still takes the flex gap. Debug mode wraps
+           every rendered template in <!-- begin/end --> comments, so the hooks count only once those are stripped. *}
+        {if $configuration.display_taxes_label || $product.ecotax.amount > 0 || ($smarty.capture.product_price_hooks|regex_replace:'/<!--.*?-->/s':''|trim) != ''}
```
tblivet:
```suggestion
        {if $configuration.display_taxes_label || $product.ecotax.amount > 0 || $smarty.capture.product_price_hooks|trim}
```

tblivet (changes requested):
Hi @boo-code, the gate is good but I think we should avoid the `regex_replace`. We don't use it anywhere else in the theme, and the `<!-- begin/end -->` comments only show up in dev mode so visitors should never see that empty block. We can, however, keep the `|trim`. So can you also drop the {* WHY *} comment.

boo-code:
@tblivet Applied: the gate is your suggestion (`|trim`, no `regex_replace`) and the WHY comment is gone, squashed into the second commit. Checked the condition with Smarty 4.5.7: an empty or whitespace-only capture renders no block; hook content, the label or ecotax render it; debug-mode comments around an empty hook still render it, as you said.

tblivet (approved):
Thanks for changes @boo-code 👍

tblivet:
Thank you @boo-code 👍 

---

✅ **QA: approved.**

Tested on PrestaShop 9.3.0, PHP 8.1.33, hummingbird 2.1.0, once on the commit this branch started from and once on the branch. All sixteen combinations of four switches were read on both: the country "Display tax label", "Enable taxes", the tax included or tax excluded price display for the visitor group, and a product with an ecotax against one without. A 20% French VAT was installed first, since this shop shipped with no tax rules and the two price modes would otherwise have shown the same amount.

Ten of the sixteen were wrong before and all sixteen are right now. Nothing else moved.

| Country label | Taxes | Prices | Before | After |
| --- | --- | --- | --- | --- |
| shown | on | tax incl. | `Tax included`, €22.14 | unchanged |
| shown | on | tax excl. | `Tax excluded`, €19.12 | unchanged |
| shown | off | either | `No tax` | unchanged |
| hidden | on | either | empty area rendered | area gone |
| hidden | off | either | `No tax` | nothing |

With the label hidden and an ecotax present, the old code ran the two together as `No tax Including €5.00 for ecotax`, with no separator, because the separator was already gated on the setting. That now reads `Including €5.00 for ecotax (not impacted by the discount)`. The ecotax line, the separator and the two `displayProductPriceBlock` captures survive in every state where they belong, and the quick view was read in every state too.

Also checked in both runs: the front page, a product page and the cart, at 1280, 768 and 375 wide. No console or network error either side.

Before:
clip-08-the-price-block-with-the-tax-label-off-taxes-off

After:
clip-08-the-price-block-with-the-tax-label-off-taxes-off

PR #1103, opened 2026-09-08
Emit only the srcset candidates whose image type exists

boo-code:
| Questions         | Answers
| Description?      | Every srcset in the theme named its image types unconditionally, so a shop missing one emitted a candidate with a descriptor and no URL - `" 216w"` or `" 2x"`. That is not a valid srcset candidate, the browser resolves the bare descriptor as a relative URL and requests `/216w`, and the product listing loses its images. It also raised an `Undefined array key` warning per candidate - **22 on a single category page** in the reported configuration. Each srcset is now built from the sizes that actually exist, through a shared `catalog/_partials/srcset.tpl`, and the attribute is dropped when none is. Fixes PrestaShop/PrestaShop#41511.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#41511
| Sponsor company   |
| How to test?      | On a shop whose image types do not include everything `config/theme.yml` declares - an upgrade from 9.0.x reproduces it, or rename `default_sm` and `default_lg` in `ps_image_type` - open any page with product miniatures. Before: `srcset=" 216w, .../default_md/....jpg 261w, 336w"`, a 404 on `/216w`, no images in the listing, and `Undefined array key` warnings in debug mode. After: `srcset=".../default_md/....jpg 261w"` - a valid single candidate, images render, no warnings. Restore the image types and the full three-candidate srcset comes back unchanged.

### Why the shop can be missing a type the theme names

An image type declared in `theme.yml` only reaches `bySize` once the shop has a row for it in
`ps_image_type`. A theme's image types are applied when the theme is installed, not again on every
upgrade, so a shop upgraded from 9.0.x keeps the types it had. In the reported case the shop carried
`default_xs, default_s, default_m, default_md, default_xl` while Hummingbird 2.0 asks for `default_sm`,
`default_md` and `default_lg` - `default_md` was the only one of the three present, which is exactly why
the reporter saw listings break while product pages were fine: those use `default_xl` and `product_main`,
which the shop did have.

**The data half of this is not a theme fix.** Making an upgrade re-apply the enabled theme's declared
image types belongs in the core `ThemeManager` or in the update process, per the theme's own boundary
rule in `CONTEXT.md`. This PR is the presentation half: the theme should not emit invalid markup when the
shop does not match what `theme.yml` declares.

### The whole class, not just the reported page

The reported symptom was the product miniature, but every srcset in the theme had the same shape. Two
descriptor forms are in use and both were affected:

- width descriptors (`216w`) in the catalog templates;
- density descriptors (`2x`) in the cart, order-confirmation, order-detail and pack lines.

The partial takes the descriptor verbatim, so one implementation covers both. After the change no
template names an image type inside a srcset - verified by grep across `templates/`.

Two details worth calling out:

- lookups go through `isset()` rather than a `|default` modifier, because the modifier only runs after
  PHP has already evaluated the missing offset and warned;
- the front office runs with `$smarty->escape_html = true`, so the partial's output is already escaped
  and callers print the capture with `nofilter`. Escaping again would turn an `&` in a URL into
  `&amp;amp;`.

### Also fixed, found while doing this

The avif `<source>` in the four order-detail templates closed its srcset with `2x",` - a stray comma that
landed in the attribute list as `<source srcset="..." , type="image/avif">`. Removed.

### Deliberately out of scope

`width`, `height` and `src` attributes still name a fixed image type, so on a shop missing that type they
still emit a warning and an empty value. That is a different defect - missing intrinsic dimensions rather
than an invalid srcset candidate - and the right fix there is to fall back to an available size, not to
omit the attribute, since omitting `width`/`height` reintroduces layout shift. It is left untouched rather
than half-fixed; the srcset class in this PR is complete.

### Verification

- **Reproduced first**, by renaming `default_sm`/`default_lg` in `ps_image_type` on a 9.2 shop:
  `srcset=" 216w, .../default_md/....jpg 261w, 336w"` and **22** `Undefined array key` warnings on one
  category page. Restoring the types restored the correct output, so the trigger is confirmed.
- **After:** with those types missing, `srcset=".../default_md/....jpg 261w"`, 0 malformed candidates,
  0 warnings. With all types present, the three-candidate srcset is byte-identical in content to before.
- Also exercised with `default_xs` missing, which is the density-form equivalent: the cart line degrades
  to a single `... 2x` candidate, still valid, 0 malformed.
- Pages actually rendered: home (miniatures), a product page (cover + thumbnail strip + modal) and the
  cart (detailed product line, density form). All 13 changed templates additionally pass a Smarty
  compile check (`compileTemplateSource()`), since the order-confirmation and order-detail templates were
  not reachable by rendering in this environment.

### Notes for publishing

- The reporter's theme is Themevolty, a third-party theme based on Hummingbird, so their own copy of
  these templates needs the same change; the fix here only covers the bundled theme.
- Mention the core/upgrade half explicitly so it is not assumed this closes the whole problem: a shop
  still ends up without the responsive variants until its image types match `theme.yml`.

### DIFF
--- a/templates/catalog/_partials/miniatures/product-image.tpl
+++ b/templates/catalog/_partials/miniatures/product-image.tpl
@@ -3,27 +3,25 @@
  * LICENSE.md file that was distributed with this source code.
  *}
 {block name='product_miniature_image'}
+  {assign var='miniatureSizes' value=['default_sm' => '216w', 'default_md' => '261w', 'default_lg' => '336w']}
   <div class="{$componentName}__image-container thumbnail-container">
     <a href="{$product.url}" class="{$componentName}__image-link outline outline--rounded">
       {if $product.cover}
+        {capture name='miniatureSrcsetAvif'}{include file='catalog/_partials/srcset.tpl' image=$product.cover sizes=$miniatureSizes srcsetVariant='avif'}{/capture}
+        {capture name='miniatureSrcsetWebp'}{include file='catalog/_partials/srcset.tpl' image=$product.cover sizes=$miniatureSizes srcsetVariant='webp'}{/capture}
+        {capture name='miniatureSrcset'}{include file='catalog/_partials/srcset.tpl' image=$product.cover sizes=$miniatureSizes}{/capture}
         <picture>
-          {if isset($product.cover.bySize.default_md.sources.avif)}
+          {if $smarty.capture.miniatureSrcsetAvif|trim}
             <source
-              srcset="
-                {$product.cover.bySize.default_sm.sources.avif} 216w,
-                {$product.cover.bySize.default_md.sources.avif} 261w,
-                {$product.cover.bySize.default_lg.sources.avif} 336w"
+              srcset="{$smarty.capture.miniatureSrcsetAvif|trim nofilter}"
               sizes="(min-width: 992px) 25vw, (min-width: 360px) 50vw, 100vw"
               type="image/avif"
             >
           {/if}
 
-          {if isset($product.cover.bySize.default_md.sources.webp)}
+          {if $smarty.capture.miniatureSrcsetWebp|trim}
             <source
-              srcset="
-                {$product.cover.bySize.default_sm.sources.webp} 216w,
-                {$product.cover.bySize.default_md.sources.webp} 261w,
-                {$product.cover.bySize.default_lg.sources.webp} 336w"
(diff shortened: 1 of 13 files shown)


ps-jarvis:

- `Shop.Theme.Catalog`
    - [ ] `Slide to product image %number%`
    - [ ] `Products details`
- `Shop.Theme.Customeraccount`
    - [ ] `Products details`

PR #1104, opened 2026-09-09
Add QA testing checklist

tblivet:
| Questions         | Answers
| Description?      | Adds `docs/qa/testing-checklist.md`, a full test pass on the theme, cross-page checks, page by page, the back-office settings that change the front office, the native modules, overrides, accessibility and responsive checks. It is also meant to become the checklist for AI-assisted end-to-end tests.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | --

### DIFF
--- a/docs/qa/testing-checklist.md
+++ b/docs/qa/testing-checklist.md
@@ -106,7 +106,7 @@ Do these once per profile on any page, then spot-check elsewhere.
 ### 2.3 Shared partials
 
 - [ ] Breadcrumb: correct trail, current page not a link
-- [ ] Notifications and toasts: success, warning, error
+- [ ] Notifications, and toasts, which come from errors and from the quantity input rather than from add to cart
 - [ ] Pagination: first, last, middle page, and the SEO `rel` links
 - [ ] Page title section, section titles
 - [ ] Quantity input (`useQuantityInput`): min, max, step, manual typing
@@ -143,7 +143,7 @@ Applies to `category`, `search`, `best-sales`, `new-products`, `prices-drop`, `m
 - [ ] Category with a cover image but no thumbnail, and the reverse, since the two images are separate fields
 - [ ] `displaySubcategories` hook: the list still renders when a module provides content and there is no native subcategory
 - [ ] Product miniatures: image, name, price, flags (new, on sale, pack, out of stock), review stars
-- [ ] Add to cart from a miniature: toast, cart preview and totals updated, quantity input respected
+- [ ] Add to cart from a miniature: the Added to your cart modal opens, cart preview and totals updated, quantity input respected
 - [ ] Miniature quantity input honours the product's minimum quantity, and a lower value is refused
 - [ ] Product with combinations added from a miniature lands in the cart with its default combination
 - [ ] Miniature falls back to a See details link when the product has no `add_to_cart_url`, so not available for order, or out of stock with orders denied
@@ -181,7 +181,7 @@ BO tab and its front-office footprint:
 
 | BO tab | Check in FO |
 | --- | --- |
-| Description | Summary in the buy block, Description under the gallery, rich text and inline images, Categories and Default category (breadcrumb, URL, structured data), Brand, Related products in the You might also like block |
+| Description | Images: the back-office order is the gallery order, the cover is the listing miniature, the caption fills `alt` and `title`, and a product with no image falls back to No image available. Summary in the buy block, Description under the gallery, rich text and images inside the description, Categories and Default category (breadcrumb, URL, structured data), Brand, Related products in the You might also like block |
 | Details | Reference, MPN, UPC barcode, GTIN and ISBN in Product details, Display condition on product page, Features in Data sheet, Attached files in Download, and the Customization fields |
 | Stocks (standard product) | Quantity, Minimum quantity for sale, When out of stock (deny, allow, or the shop default), Label when in stock, Label when out of stock, Availability date, `ps_emailalerts` back-in-stock form |
 | Combinations (combination product) | Selector rendering per attribute type (select, radio, colour swatch), price and quantity update on selection, image swap, combination reference, impossible combination, default combination on load. When out of stock and the two labels are set here, there is no Stocks tab |
@@ -195,14 +195,14 @@ BO tab and its front-office footprint:
 Also on the product page:
 
 - [ ] Image gallery: thumbnails, zoom, fullscreen modal, keyboard arrows
-- [ ] Add to cart: toast, cart preview updated, quantity respected
+- [ ] Add to cart: the Added to your cart modal opens with the product and the cart summary, cart preview updated, quantity respected
 - [ ] Quantity below the minimum is refused
 - [ ] Customisation, from the Customization fields on the Details tab: text and file fields, required ones block add to cart while optional ones do not, errors shown on the field, saved value visible in the cart and on the order
 - [ ] Accessories block, titled You might also like: rendered when Related products are set on the Description tab, absent when none are
 - [ ] Accessory miniatures behave like listing ones: flags, review stars, add to cart, quickview, and the See details fallback
 - [ ] A disabled or non-visible product set as an accessory does not appear
 - [ ] The block is rendered without the `container` wrapper, so check its width against the rest of the page
-- [ ] `ps_sharebuttons`, `productcomments`, `ps_productinfo`, `ps_categoryproducts`, `ps_crossselling`, `ps_viewedproduct`
+- [ ] `ps_sharebuttons`, `productcomments`, `ps_categoryproducts`, `ps_crossselling`, `ps_viewedproduct`
 - [ ] Product JSON-LD present and valid
 
 ### 3.5 Cart
@@ -332,13 +332,13 @@ Not a separate pass. These are the settings that change what section 3 renders, 
 
 Every native module Hummingbird overrides or hooks. Hooks come from `config/theme.yml`, templates from `modules/<name>/`. JS means the module has theme-side TypeScript in `src/js/modules/`.
 
+`ps_advertising`, `ps_productinfo` and `ps_rssfeed` are left out: PrestaShop does not install them and their repositories are archived, so the theme's overrides for them never render.
+
 Module versions are pinned by the core, not by the theme. Record the ones you tested against, from the PrestaShop checkout:
 
 ```
 composer show 'prestashop/*' --direct | grep -E 'blockreassurance|contactform|productcomments|psgdpr|ps_'
 ```
-
-Three overrides target modules the core does not require, so a stock install will not exercise them. They are still published on Packagist and can be installed by hand: `ps_advertising`, `ps_productinfo`, `ps_rssfeed`. Check them against the core's `composer.json` before reporting them as dead overrides.
 
 ### 5.1 Header and navigation
 
@@ -373,8 +373,6 @@ Three overrides target modules the core does not require, so a stock install wil
 | `ps_specials` | Column or home block, `/prices-drop` | Needs an active specific price |
 | `ps_brandlist` | Column block, `/brands` | Disabled by default. Block, form variant, text variant |
 | `ps_supplierlist` | Column block, `/suppliers` | Disabled by default. Same three variants |
-| `ps_advertising` | Column block | Not required by the core, install it first. Image, link, empty state |
-| `ps_rssfeed` | Column block | Not required by the core, install it first. Valid feed, unreachable feed, malformed feed |
 
 ### 5.4 Product page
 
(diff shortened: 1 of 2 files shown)


docs/qa/testing-checklist.md
```diff
@@ -0,0 +1,486 @@
...
+- [ ] A second language (`ps_languageselector`, translations, date and price formats)
+- [ ] A second currency (`ps_currencyselector`)
+- [ ] Multistore, if the change touches header, footer or URLs **(config)**
+
+### 1.4 Cache
+
+In Advanced Parameters > Performance:
+
+- [ ] CCC (Combine, Compress and Cache): all options disabled, so the pass exercises the theme's own assets rather than a merged bundle
+- [ ] Smarty: Force compilation set to Yes, Cache set to No (`PS_DEV_MODE: 1` in the compose files already forces compilation)
```
mattgoud, line 79: `PS_DEV_MODE: 1` is only in `docker/docker-compose-prestashop.yml` (line 13). `docker-compose-flashlight.yml` doesn't set it, so "the compose files" over-promises for anyone on option A with Flashlight. Suggest naming the file: "`PS_DEV_MODE: 1` in `docker-compose-prestashop.yml` already forces compilation; the Flashlight compose file does not set it".

docs/qa/testing-checklist.md
```diff
@@ -0,0 +1,486 @@
...
+- [ ] Password fields: reveal toggle, password policy meter
+- [ ] Modals: focus trap, `Esc`, backdrop click, scroll lock released on close
+
+### 2.4 Regression sweep
+
+- [ ] No JS error in the console on any page visited
+- [ ] No 404 on CSS, JS, fonts or images
+- [ ] No unstyled flash
+- [ ] Images use the image types declared in `config/theme.yml` and are not upscaled
+- [ ] If the change touches `templates/layouts/`, each layout declared in `config/theme.yml` still renders, with no empty column
```
mattgoud, line 122: Anchoring this on `config/theme.yml` quietly drops two layouts that exist in `templates/layouts/`: `layout-content-only.tpl` and `layout-error.tpl`. `available_layouts` only declares the four selectable ones, but `layout-error.tpl` is what `errors/maintenance.tpl` and `errors/restricted-country.tpl` extend, and those two are already tested in 3.10. Worth making the source `templates/layouts/` rather than the yml, so the sweep covers all six.

docs/qa/testing-checklist.md
```diff
@@ -0,0 +1,486 @@
...
+- [ ] Add to cart from a miniature: the Added to your cart modal opens, cart preview and totals updated, quantity input respected
+- [ ] Miniature quantity input honours the product's minimum quantity, and a lower value is refused
+- [ ] Product with combinations added from a miniature lands in the cart with its default combination
+- [ ] Miniature falls back to a See details link when the product has no `add_to_cart_url`, so not available for order, or out of stock with orders denied
+- [ ] Sort orders: each option changes the order
+- [ ] `ps_facetedsearch`: each facet type (checkbox, dropdown, slider, colour), active filters, clear all, shareable URL, browser back and forward
+- [ ] `ps_categorytree` in the left column
+- [ ] Pagination and items per page (Shop Parameters > Product Settings > Pagination)
+- [ ] Empty listing, and listing with a single product
+- [ ] `/brands` and `/suppliers` (enable `ps_brandlist` and `ps_supplierlist` first, both disabled by default)
```
mattgoud, line 155: The `/brands` and `/suppliers` pages aren't gated by the modules, they're gated by the core settings: `SupplierController::init()` checks `PS_DISPLAY_SUPPLIERS` and `ManufacturerController::init()` checks `PS_DISPLAY_MANUFACTURERS`. `ps_brandlist`/`ps_supplierlist` only provide the column blocks (and those are indeed disabled by default via `to_disable`). Section 4.3 already gets this right with "Display brands / Display suppliers → Enables `/brands` / `/suppliers`", so this line just needs to point at the setting too, otherwise a tester enables the modules and still gets a 404.

docs/qa/testing-checklist.md
```diff
@@ -0,0 +1,486 @@
...
+- [ ] Vouchers page, once a cart rule applies to the customer
+- [ ] `ps_emailalerts` notification list
+- [ ] Guest tracking
+
+### 3.9 CMS and static pages
+
+- [ ] CMS page and CMS category
+- [ ] Sitemap, including the nested list
+- [ ] Stores page: map, store list, opening hours (Shop Parameters > Contact > Stores)
+- [ ] Contact page: `contactform` widget, file attachment, order selector, `ps_contactinfo` in both side columns
```
mattgoud, line 268: "in both side columns" isn't reachable in one configuration. `templates/contact.tpl` renders `displayContactLeftColumn` **or** `displayContactRightColumn` in an `if/elseif` on `$layout`, and `theme_settings.layouts` pins `contact: layout-left-column`, so out of the box only the left one ever fires. Getting the right-column variant means switching the contact page layout in Design > Theme & Logo, which makes it a **(config)** item. Suggest: "`ps_contactinfo` in the left column, then switch the contact page to `layout-right-column` and check the right one **(config)**".

docs/qa/testing-checklist.md
```diff
@@ -0,0 +1,486 @@
...
+| `ps_emailsubscription` | `displayFooterBefore` | Subscribe, unsubscribe, invalid email, already subscribed, GDPR consent, column variant |
+| `ps_socialfollow` | `displayFooterBefore`, unhooked from `displayFooter` | Each network, empty configuration |
+| `contactform` | `displayContactContent` | Subject list, order selector for logged-in customers, file attachment, validation errors, success message |
+| `psgdpr` (JS) | `displayGDPRConsent`, `displayCustomerAccount` | Consent checkbox blocks submission on every form that hosts it, personal data page, export and deletion request |
+
+### 5.7 Coverage
+
+- [ ] Each module above visited once per profile
+- [ ] Modules whose absence changes the layout checked disabled, so at least one per hook position: header nav, home block, column block, footer block, product block. No empty wrapper, no collapsed grid **(config)**
+- [ ] Each column module checked in the left and the right column
```
mattgoud, line 411: Same point as 3.9, generalised: `config/theme.yml` only hooks `ps_categorytree` and `ps_facetedsearch` on `displayLeftColumn`, and no page in `theme_settings.layouts` uses `layout-right-column`. So "checked in the left and the right column" always requires changing a page's layout first, and deserves the **(config)** marker the rest of the file uses so consistently.

docs/qa/testing-checklist.md
```diff
@@ -0,0 +1,486 @@
...
+- [ ] `ps_mainmenu`: desktop dropdowns, mobile burger, keyboard traversal
+- [ ] `ps_searchbar`: autocomplete, no-result state, submit
+- [ ] `ps_customersignin`, `ps_shoppingcart`, `ps_languageselector`, `ps_currencyselector` in the top nav
+- [ ] `ps_contactinfo` in `displayNav1`
+- [ ] `blockreassurance` band in `displayNavFullWidth`
+- [ ] Sticky behaviour, no layout shift on scroll
+
+### 2.2 Footer
+
+- [ ] `ps_socialfollow`, `ps_emailsubscription`, `blockreassurance` in `displayFooterBefore`
```
mattgoud, line 101: Nit: `blockreassurance` is also hooked on `displayFooterAfter`, which 5.1 correctly lists but this line doesn't. Two `blockreassurance` instances render in the footer area, so it's worth distinguishing them here.

mattgoud (changes requested):
Really nice piece of work, and unusually well sourced. I checked the derived lists against the repo rather than just reading them, and nearly everything holds up exactly:

- the `$grid-breakpoints` table matches `src/scss/bootstrap/overrides/variables/_variables.scss` value for value
- the hook columns in section 5 match `config/theme.yml`, including the two `modules_to_unhook` entries
- the `(JS)` marks map 1:1 onto the 10 entries in `src/js/modules/` (`blockcart.ts` for `ps_shoppingcart` included)
- "four templates the default pass never renders" in 3.7 is exactly the four `*-multishipment.tpl` files
- the `small_default` vs `category_default` subcategory fallback in 3.2 is exactly what `templates/catalog/_partials/subcategories.tpl` does
- toasts "come from errors and from the quantity input rather than from add to cart": `useToast` is only consumed by `errors.ts` and `useQuantityInput.ts`
- "Save customization posts to the product URL": `product-customization.tpl` does `action="{$product.url}"`
- the accessories block "rendered without the `container` wrapper": `need_container = false` in `product-accessories.tpl`
- section 3 covers every page-level template, and the README anchors, port 8887 and `npm test` all resolve

One real gap plus a few factual corrections below. The gap is the only thing I'd call blocking, and it's a one-row fix.

**`blockwishlist` is missing.** It has a theme override in `modules/blockwishlist` (a `displayCustomerAccount` hook template and five modals: create, rename, delete, share, add-to-wishlist, login) and it's in `global_settings.modules.to_disable` in `config/theme.yml`. Section 5 says it lists "every native module Hummingbird overrides or hooks" and names only `ps_advertising`, `ps_productinfo` and `ps_rssfeed` as exclusions, and section 9 sets the rule "one entry per overridden module". So by the file's own contract `blockwishlist` needs a row, and the exclusion rationale doesn't apply to it: it isn't archived and it's actively maintained. It's also one of the larger overrides in the theme, so the flows behind it (add to wishlist from a miniature, create/rename/delete/share a list, the login modal for a guest, the account tile) are currently untested by this pass. It's disabled by default like `ps_brandlist`/`ps_supplierlist`, so it can carry the same note. 5.6 looks like the right place.

Nothing else is a blocker. Once `blockwishlist` has a row I'm happy to approve.

tblivet:
Thanks @mattgoud for the review!

All corrections applied, except the one related to `blockwishlist`. We talked about it in the past and decided the module isn't stable: the theme disables it at install, and the overrides are due for removal, so testing them would mean testing code we intend to delete. Documented as an explicit exception in section 5, with section 9's rule amended to match. You were right that the file was silent on this, that's the gap, rather than the missing row.

Also dropped the hardcoded counts ("six layouts", "four templates") in favour of the directory and the glob: a count goes stale silently, which is what section 9 is meant to prevent.

mattgoud (dismissed):
Thanks @tblivet, all five corrections landed and I re-checked each one against the repo rather than just reading the diff:

- `PS_DEV_MODE` now names `docker-compose-prestashop.yml` and says the Flashlight file does not set it, which matches
- the second `blockreassurance` on `displayFooterAfter` is called out in 2.2
- the layouts sweep now takes `templates/layouts/` as its source, and the `content_only` claim checks out: `FrontController::getLayout()` sets `$layout = 'layout-content-only'` when the request carries `content_only`
- `/brands` and `/suppliers` now point at the settings, and the location is right: `display_suppliers` and `display_manufacturers` live in `PreferencesType`, under `Form/Admin/Configure/ShopParameters/General/`, so Shop Parameters > General is correct, and it's marked **(config)**
- 3.9 and 5.7 both say the layout has to be switched first, marked **(config)**

The section 9 table keeps up with all of it: the new `templates/layouts/` row, `config/theme.yml` narrowed to the selectable layouts, and the `modules/` rule amended to admit the exception.

Dropping the hardcoded counts is the right call and better than what I reviewed against. I leaned on "exactly four multishipment templates" myself when I checked the file, and a glob is what section 9 is arguing for everywhere else, so it should not have been a number in the first place.

On `blockwishlist`, that works for me. My finding was never really "this module must be tested", it was that the file claimed to list every overridden module, named three exclusions, and then quietly dropped a fourth. You have fixed that: the exception is stated, the reason is given, the two paths it still owns are named (`modules/blockwishlist/` and `src/scss/prestashop/modules/_blockwishlist.scss`, both present), and there is an exit condition saying what happens if the module stabilises. A documented exception with a way out is a better answer than a row for code you intend to delete.

Approving.

tblivet:
QA approved ✅

These are QA tools only (checklist and helper scripts under docs/qa), nothing that ships with the theme. Consider it a beta: it was used for a first full pass on 2.1.0 / PS 9.2.0 and will improve with each campaign.

mattgoud (approved):

docs/qa/config.php
```diff
@@ -0,0 +1,145 @@
...
+        echo $now . PHP_EOL;
+        exit($now === $value ? 0 : 1);
+
+    case '--unset':
+        if (!$arg) { fwrite(STDERR, "usage: --unset KEY\n"); exit(64); }
+        Configuration::deleteByName($arg);
+        Configuration::loadConfiguration();
+        $now = readValue($arg);
+        echo $now . PHP_EOL;
+        exit($now === UNSET_MARKER ? 0 : 1);
```
Quetzacoalt91: Looks like you could rely on the existing command of PrestaShop: https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Command/ConfigCommand.php#L203

```
php bin/console prestashop:config [...]
```

docs/qa/config.php
```diff
@@ -0,0 +1,145 @@
...
+    case '--set-flag':
+        if (!$arg || !str_contains($arg, '=')) { fwrite(STDERR, "usage: --set-flag NAME=0|1\n"); exit(64); }
+        [$flag, $state] = explode('=', $arg, 2);
+        Db::getInstance()->update('feature_flag', ['state' => (int) $state], 'name = "' . pSQL($flag) . '"');
+        $now = Db::getInstance()->getValue(
+            'SELECT state FROM ' . _DB_PREFIX_ . 'feature_flag WHERE name = "' . pSQL($flag) . '"'
+        );
+        $now = ($now === false || $now === null) ? UNSET_MARKER : (string) (int) $now;
+        echo $now . PHP_EOL;
+        exit($now === (string) (int) $state ? 0 : 1);
```
Quetzacoalt91: https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Command/FeatureFlagCommand.php

PR #1107, opened 2026-09-16
Fix product list rating error

tblivet:
| Questions         | Answers
| Description?      |  Fix product ratings from product listings. The theme sent duplicated product IDs to the productcomments grade endpoint (the same product is rendered by several widgets on the home page), which made the request fail and wiped every rating on the page. IDs are now deduplicated. Also fixes the request URL being built with a literal `?`, which dropped the first ID on shops with friendly URLs disabled, and adds a reset so a product whose reviews were all deleted no longer keeps its former grade.<br><br>Note: the crash is a productcomments bug, already fixed upstream (issue https://github.com/PrestaShop/productcomments/issues/226, fix https://github.com/PrestaShop/productcomments/pull/227) but not in any released version yet, so shops on 8.0.1 and 9.0.0 still hit it, and sending each id once is right anyway.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | Install productcomments (v8 or v9 the module bug is identical in both) and add reviews on several products. Then make the home page render duplicated products **followed by product ids that appear nowhere else**: on the demo catalog, Featured products and Latest arrivals already show the same 4 products, so it is enough to discount two other products (e.g. ids 10 and 11) so they show up in the last block, Special deals. That ordering is what triggers the bug, duplicates on their own are harmless. Before: the home page shows no star at all, on any product. After: every rated product shows its stars, on each of its occurrences. In both cases, category pages (no duplicates) keep working and products without reviews keep showing nothing. Version-independent check: calling `/module/productcomments/CommentGrade?id_products[]=1&id_products[]=1&id_products[]=8&id_products[]=2` returns a 500 before the fix and valid JSON after.


mattgoud (commented):
Tested on a 9.2.x dev shop with hummingbird active and productcomments 9.0.0 (moderation on), comparing a build of `2.x` (bbd31dfe) against `2.x` + this branch, swapping `assets/js/theme.js` and clearing the CCC cache between runs.

One note for whoever tests this next: the duplicate has to come **before** a fresh id. On my home page Special deals was the last block, so the duplicates landed at the end of the list and everything worked fine. Moving `ps_specials` above `ps_newproducts` gives the DOM order `1 2 3 4 11 10 2 1 19 18 17 16`, which reproduces it.

| Scenario | `2.x` | This PR |
|---|---|---|
| Home, duplicates followed by fresh ids | no star on any of the 12 blocks, `SyntaxError: Unexpected token '<'` in console | 1 → 5★ (2 reviews), 2 → 4★, 11 → 2★, 10 → 5★ on every occurrence, 3/4/16/17/18/19 hidden, console clean |
| Listing without duplicates (`/prices-drop`) | works | works |
| Friendly URLs disabled | `...&id_lang=1?id_products[]=1&...`, id 1 is lost | every id arrives, grades correct |
| All reviews deleted, then `updatedProduct` | block stays visible: "Rated 5 out of 5 stars based on 2 review(s)" with 0 reviews left | block hidden, counters reset to 0 |

The three fixes all do what they claim.

### The endpoint bug is already fixed upstream, just not released yet

Worth linking from here, because it changes how this PR reads. The 500 comes from `array_unique()` in `controllers/front/CommentGrade.php` preserving keys, which makes `if ($count - 1 > $index)` in `ProductCommentRepository::getAverageGrades()` and `getCommentsNumberForProducts()` drop a comma between two columns that are not the last one. That is PrestaShop/productcomments#226, fixed by PrestaShop/productcomments#227 (`array_values()`), merged into `dev` on 2026-09-09 at 16:00.

v9.0.0 was cut the same day at 10:46, so the fix is in no released version yet: it is queued for 9.0.1, whose release PR is still open.

So this PR stays useful regardless, since every shop currently on 8.0.1 or 9.0.0 hits the crash, and sending each id once is the right thing to do anyway. It would just help reviewers to say in the description that the server side is already handled and that this is the theme doing its part, with the two links above.

### One correction to the description

With friendly URLs disabled, the literal `?` loses exactly one id, the first one, the one glued onto `id_lang=1`. Not "all but the last ID". Measured against the module:

```
.../index.php?fc=module&module=productcomments&controller=CommentGrade&id_lang=1?id_products[]=1&id_products[]=2&id_products[]=10
→ {"products":[{"id_product":2,...},{"id_product":10,...}]}
```

`id_lang` swallows `1?id_products[]=1`, the rest parses normally. Worth fixing since QA reads that field.

### Tests

The repo has jest (7 suites, 39 tests green here). The three behaviours added are pure logic and would be cheap to cover, but they sit in private statics of a class that is not exported, so it needs a small extraction first. Not blocking, just flagging it.

### Nits

- `clearProductListReview()` now injects a `star-content` with `role="img" aria-label="0 out of 5 stars"` into every product that has no review. Harmless today because `.product-list-review` is `display: none`, but that rule is scoped to `.product-miniature`. Rendered outside a miniature, a screen reader would announce "Rated 0 out of 5 stars based on 0 review(s)" on unrated products. Returning early when there is nothing to clear would avoid it.
- `indexOf()` rather than `Set`: correct, `tsconfig.json` targets `es5`. Nothing to change.
- `productListReviews[0].getAttribute('data-url')` in place of the `querySelector`: strictly equivalent, and guarded by the `length === 0` early return above.
- `if (!productListReviews)` → `length === 0` is a real fix on its own, a `NodeList` is always truthy.

Branch, scope and CI all look right: `2.x` for a bug fix, one file, no lock file noise.

mattgoud (approved):
Approving on the strength of the tests reported above, run against bcb0dbbe: the three fixes reproduce and hold, and the non duplicated listings are unaffected.

The remaining points are all non blocking and can be handled here or later: the "all but the last ID" wording in the description, the missing jest coverage, and the `star-content` rendered into unrated products by `clearProductListReview()`.

cnavarro-prestashop:
🟢 Approved

Tested the documented reproduction: a home page rendering the same product in more than one widget (Featured products and Special deals both show product 1), followed by a block containing products that appear nowhere else on the page (Special deals also shows products 10 and 11, only there).

Before the fix: no product on the page shows a star rating. The ratings request carries the duplicated id list and fails with a 500, which wipes every rating on the page, including for products that only appear once.

image


After the fix: every rated product shows its stars, on each of its occurrences.

image


Category pages (no duplicates) and products without reviews were unaffected in both states. Smoke pages and the 375/768 responsive checks pass in both states.


[video.webm](https://github.com/user-attachments/assets/31d00bbb-7cd0-4b79-b9ce-4307480c0284)

